PR #29, opened 2019-11-18
Add DDL for webhook transaction log

jbrockopp:


database/ddl/postgres/hook.go
```diff
@@ -0,0 +1,46 @@
...
+);
+`
+
+	// CreateHookRepoIDBuildIDIndex represents a query to create an
+	// index on the hooks table for the repo_id and build_id columns.
+	CreateHookRepoIDBuildIDIndex = `
+CREATE INDEX
+IF NOT EXISTS
+hooks_repo_id_build_id
+ON hooks (repo_id, build_id);
```
kneal: Should `source_id` also have an index?

jbrockopp: I don't believe so 👍

PR #52, opened 2019-12-04
feat(metrics): add metrics for how many times an image has been used

meinenec:
Chose to .Scan() the image name and count into a slice of a custom struct then range over that into a map instead of reading directly into a map because I found using a custom struct made things much more readable.

You can see an example of reading into a map [in an issue here](https://github.com/jinzhu/gorm/issues/1167).


database/step_test.go
```diff
@@ -202,6 +202,63 @@ func TestDatabase_Client_GetBuildStepCount(t *testing.T) {
...
+	want["baz"] = 2
+	want["bazian:latest"] = 1
+
+	// setup database
+	db, _ := NewTest()
+	defer func() {
+		db.Database.Exec("delete from steps;")
+		db.Database.Close()
+	}()
+	err := db.CreateStep(sOne)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

database/step_test.go
```diff
@@ -202,6 +202,63 @@ func TestDatabase_Client_GetBuildStepCount(t *testing.T) {
...
+	err = db.CreateStep(sTwo)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

database/step_test.go
```diff
@@ -202,6 +202,63 @@ func TestDatabase_Client_GetBuildStepCount(t *testing.T) {
...
+	err = db.CreateStep(sThree)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

database/client.go
```diff
@@ -92,6 +92,7 @@ func NewTest() (*client, error) {
 
 	// create the database client
 	db, err := gorm.Open(name, config)
+	//db.LogMode(true)
```
jbrockopp: I believe this line can be removed 👍

database/dml/postgres/service.go
```diff
@@ -29,6 +29,14 @@ OFFSET $3;
 SELECT count(*) as count
 FROM services
 WHERE build_id = $1
+`
+
+	// SelectServiceImagesCount represents a query to select
+	// the count of an images appearances in the database.
+	SelectServiceImagesCount = `
+SELECT image, COUNT(image)
```
jbrockopp: I wonder if we want to add the `as count` after the `COUNT(image)` part, similar to how we do in other spots?

i.e. https://github.com/go-vela/server/pull/52/files#diff-da7d2b57765443dd3c6706913bd9b043R29

Also, do you think we should lowercase the `COUNT` here similar to how we do in our other count queries?

i.e. https://github.com/go-vela/server/pull/52/files#diff-da7d2b57765443dd3c6706913bd9b043R29

database/dml/postgres/step.go
```diff
@@ -29,6 +29,14 @@ OFFSET $3;
 SELECT count(*) as count
 FROM steps
 WHERE build_id = $1
+`
+
+	// SelectStepImagesCount represents a query to select
+	// the count of an images appearances in the database.
+	SelectStepImagesCount = `
+SELECT image, COUNT(image) as count
```
jbrockopp: Do you think we should lowercase the `COUNT` here similar to how we do in our other count queries?

i.e. https://github.com/go-vela/server/pull/52/files#diff-14ef7dccc93c0e8ae7848e5f5bdc273dR29

database/dml/sqlite/service.go
```diff
@@ -29,6 +29,14 @@ OFFSET ?;
 SELECT count(*) as count
 FROM services
 WHERE build_id = ?
+`
+
+	// SelectServiceImagesCount represents a query to select
+	// the count of an images appearances in the database.
+	SelectServiceImagesCount = `
+SELECT image, COUNT(image)
```
jbrockopp: I wonder if we want to add the as count after the COUNT(image) part, similar to how we do in other spots?

i.e. https://github.com/go-vela/server/pull/52/files#diff-6aa5921b68a1b3adacfb421ea9f08fbaR29

Also, do you think we should lowercase the COUNT here similar to how we do in our other count queries?

i.e. https://github.com/go-vela/server/pull/52/files#diff-6aa5921b68a1b3adacfb421ea9f08fbaR29

database/dml/sqlite/step.go
```diff
@@ -29,6 +29,14 @@ OFFSET ?;
 SELECT count(*) as count
 FROM steps
 WHERE build_id = ?
+`
+
+	// SelectStepImagesCount represents a query to select
+	// the count of an images appearances in the database.
+	SelectStepImagesCount = `
+SELECT image, COUNT(image) as count
```
jbrockopp: Do you think we should lowercase the `COUNT` here similar to how we do in our other count queries?

i.e. https://github.com/go-vela/server/pull/52/files#diff-5ab24d0f4f051b28a495392dbd42fcaaR29

PR #62, opened 2019-12-17
improvement: populating build fields

jbrockopp:


source/github/webhook_test.go
```diff
@@ -66,6 +66,7 @@ func TestGithub_ProcessWebhook_Push(t *testing.T) {
 	wantBuild.SetCommit("9c93babf58917cd6f6f6772b5df2b098f507ff95")
 	wantBuild.SetSender("Codertocat")
 	wantBuild.SetAuthor("Codertocat")
+	wantBuild.SetEmail("21031067+Codertocat@users.noreply.github.com")
```
meinenec: I know this is a noreply email, but would it be better to use a fake/invalid email address?

source/github/webhook_test.go
```diff
@@ -141,6 +142,7 @@ func TestGithub_ProcessWebhook_Push_NoSender(t *testing.T) {
...
```
meinenec: I know this is a noreply email, but would it be better to use a fake/invalid email address?

source/github/webhook_test.go
jbrockopp: I'm not sure if it makes a difference. I'd tend to believe not, but if we do want to change it, we'd also have to change the payload we're using for our tests.

The email itself actually comes from the test payload we're using:

https://github.com/go-vela/server/blob/improvement/build_fields/source/github/testdata/push.json#L47

Which is actually ripped from GitHub's API documentation:

https://developer.github.com/v3/activity/events/types/#pushevent

source/github/webhook_test.go
jbrockopp: I'm not sure if it makes a difference. I'd tend to believe not, but if we do want to change it, we'd also have to change the payload we're using for our tests.

The email itself actually comes from the test payload we're using:

https://github.com/go-vela/server/blob/improvement/build_fields/source/github/testdata/push.json#L47

Which is actually ripped from GitHub's API documentation:

https://developer.github.com/v3/activity/events/types/#pushevent

PR #67, opened 2019-12-20
fix(secrets): default allow command to true

jbrockopp:


api/secret.go
```diff
@@ -51,6 +51,9 @@ func CreateSecret(c *gin.Context) {
 	if len(input.GetEvents()) > 0 {
 		input.SetEvents(unique(input.GetEvents()))
 	}
+	if !input.GetAllowCommand() {
```
golangcibot: if statements should only be cuddled with assignments (from `wsl`)

api/secret.go
```diff
@@ -200,6 +203,10 @@ func UpdateSecret(c *gin.Context) {
 	if len(input.GetEvents()) > 0 {
 		input.SetEvents(unique(input.GetEvents()))
 	}
+	if input.AllowCommand != nil {
```
golangcibot: if statements should only be cuddled with assignments (from `wsl`)

PR #73, opened 2020-01-07
feature(source_repos): implemented GetOrgRepoList to enhance SourceRepos with active status

davidvader:


api/hook.go
```diff
@@ -114,7 +114,7 @@ func GetHooks(c *gin.Context) {
 	// send API call to capture the total number of webhooks for the repo
 	t, err := database.FromContext(c).GetRepoHookCount(r)
 	if err != nil {
-		retErr := fmt.Errorf("unable to get steps count for repo %s: %w", r.GetFullName(), err)
+		retErr := fmt.Errorf("unable to get hooks count for repo %s: %w", r.GetFullName(), err)
```
davidvader: found this simple typo

api/user.go
```diff
@@ -213,17 +179,65 @@ func GetUserSourceRepos(c *gin.Context) {
...
-			Org:  org,
-			Name: name,
+			Org:    org,
+			Name:   name,
+			Active: &active,
 		}
 		output[srepo.GetOrg()] = append(output[srepo.GetOrg()], repo)
 	}
 
+	for org, _ := range output {
```
golangcibot: S1005: should omit value from range; this loop is equivalent to `for org := range ...` (from `gosimple`)

davidvader: OK OK

Issue #87, opened 2020-02-05
Allow vela projects and builds to be viewed by authenticated users


kneal:
## Description

_What is your idea?_

## Value

_Why is this important? Who does it impact? Will this make something better, faster, etc?_

## Definition of Done

_What is the end goal of this story?_

## Effort  (Optional)

_Estimated effort to complete this story? (Best Guess e.g. 1-2 days)_

## Impacted Personas (Optional)

_Which personas will benefit from completing this story? (Users, Administrators, etc)_

jbrockopp:
This story is complete 👍 🎉

PR #116, opened 2020-03-12
feat: enable public access to builds/repos

jbrockopp:
Part of the work effort for:

go-vela/server#87

### DIFF
--- a/api/repo.go
+++ b/api/repo.go
@@ -58,7 +58,7 @@ func CreateRepo(c *gin.Context) {
 	}
 
 	if len(input.GetVisibility()) == 0 {
-		input.SetVisibility("public")
+		input.SetVisibility(constants.VisibilityPublic)
 	}
 
 	if len(input.GetFullName()) == 0 {
--- a/database/repo_test.go
+++ b/database/repo_test.go
@@ -34,6 +34,7 @@ func TestDatabase_Client_GetRepo(t *testing.T) {
 	want.SetOrg("foo")
 	want.SetName("bar")
 	want.SetFullName("foo/bar")
+	want.SetVisibility("public")
 
 	// setup database
 	db, _ := NewTest()
@@ -66,6 +67,7 @@ func TestDatabase_Client_GetRepoList(t *testing.T) {
 	rOne.SetOrg("foo")
 	rOne.SetName("bar")
 	rOne.SetFullName("foo/bar")
+	rOne.SetVisibility("public")
 
 	rTwo := testRepo()
 	rTwo.SetID(2)
@@ -74,6 +76,7 @@ func TestDatabase_Client_GetRepoList(t *testing.T) {
 	rTwo.SetOrg("bar")
 	rTwo.SetName("foo")
 	rTwo.SetFullName("bar/foo")
+	rTwo.SetVisibility("public")
 
 	want := []*library.Repo{rOne, rTwo}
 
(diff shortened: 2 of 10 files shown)


JordanSussman:
Should the language of this be switched from `public` to `internal` like what [GitHub](https://help.github.com/en/github/administering-a-repository/setting-repository-visibility#making-a-repository-internal) is moving towards?

jbrockopp:
@JordanSussman we understand wanting to do our best to keep the language consistent between us and GitHub 👍 

Out of curiosity, do you see there being any problems with misunderstanding what the `internal` visibility setting means?

Asking specifically from the viewpoint that if we did swap to `internal`, we'd have the following possible settings:

* `private` - ideally this one is understandable.
* `internal` - could cause questions?

JordanSussman:
> Out of curiosity, do you see there being any problems with misunderstanding what the internal visibility setting means?

I think this will be somewhat confusing, but as GitHub continues to shift towards this verbiage it will hopefully be less confusing. 

I personally think you should align with GitHub to look like:

- `private` - ideally this one is understandable.
- `internal` - available to anyone that provides OAuth to the UI
- `public` - available to anyone

`public` likely wouldn't be allowed/enabled for Enterprise usage, but suppose I wanted to host `vela` for github.com stuff and wanted users the ability to view without needing to login

jbrockopp:
> `public` likely wouldn't be allowed/enabled for Enterprise usage, but suppose I wanted to host `vela` for github.com stuff and wanted users the ability to view without needing to login

Yeah that's the tricky part.. we never actually intended to provide an anonymous ("available to anyone") functionality. It was our plan to always enforce providing an authentication mechanism in order to login to Vela.

I agree, that we should attempt to align with GitHub to the best of our ability 👍 And honestly, we likely would have adopted GitHub's verbiage exactly if we didn't plan on only allowing two settings.

The reason we went with `public` and `private` visibility setting and didn't go with a `internal` one is because we didn't plan to allow anonymous support. When we were discussing this, we felt that `public` was a little less confusing than `internal` when it comes to describing the behavior. That kinda circles into my point around `internal` causing questions.

I'm struggling with wanting to support a use-case that doesn't really exist yet? And even if it did exist, I'm struggling with allowing a user out their to run their Vela instance in an unsecure way.

JordanSussman:
> I'm struggling with wanting to support a use-case that doesn't really exist yet? And even if it did exist, I'm struggling with allowing a user out there to run their Vela instance in an unsecure way.

I do agree the `internal` type was really prompted for a yet to be seen use-case, but potential migrations/changes down the road could be cumbersome to implement. 

> I'm struggling with wanting to support a use-case that doesn't really exist yet? And even if it did exist, I'm struggling with allowing a user out there to run their Vela instance in an unsecure way.

I'd also continue to argue that anonymous read may be desired functionality instead of security concern **if** the use-case exits.

Maybe continuing with `public` and `private` makes sense with the potential to add `anonymous` read down the road if the use-case arrives.

jbrockopp:
@JordanSussman ok 👍 

I think our plan will be to move forward with this PR. We'd like to limit scope creep and ensure that progress continues forward. To elaborate, at this time there is no logic that can handle allowing a user to access the Vela UI without authenticating which makes the value of adding the third visibility setting decrease.

**All that being said**, we'll plan to implement and account for this functionality in the future. Specifically, we're planning to create a permissions table in the database that creates a mapping of repos to user access. This will enable us to no longer sending an API call for every request to the source control system. At that time, we feel we'd have a better handle on thoughtfully adding the `anonymous` (or switch `public` to `internal`?) visibility setting.

Does that sound like an acceptable compromise?

JordanSussman:
> Does that sound like an acceptable compromise?

Sounds good.

PR #122, opened 2020-04-07
feat: add comment for repo and secrets

KellyMerrick:


source/github/github.go
```diff
@@ -20,6 +20,13 @@ import (
 const (
 	defaultURL = "https://github.com"     // Default GitHub URL
 	defaultAPI = "https://api.github.com" // Default GitHub API URL
+
+	// events for repo webhooks
+	eventPush         = "push"
+	eventPullRequest  = "pull_request"
+	eventTag          = "tag"
```
golangcibot, line 27: `eventTag` is unused (from `deadcode`)

api/webhook.go
```diff
@@ -219,6 +218,22 @@ func PostWebhook(c *gin.Context) {
...
 
+	if b.GetEvent() == constants.EventComment {
+		commit, branch, baseref, err := source.FromContext(c).GetPullRequest(u, r, number)
+		if err != nil {
+			retErr := fmt.Errorf("%s: failed to get pull request info for %s: %v", baseErr, r.GetFullName(), err)
+			util.HandleError(c, http.StatusInternalServerError, retErr)
+
+			h.SetStatus(constants.StatusFailure)
+			h.SetError(retErr.Error())
+			return
```
golangcibot, line 229: return statements should not be cuddled if block has more than two lines (from `wsl`)

api/webhook.go
```diff
@@ -409,3 +429,27 @@ func publishToQueue(queue queue.Service, p *pipeline.Build, b *library.Build, r
 		}
 	}
 }
+
+// helper function to parse the PR number out of a build ref
+func parsePRNumberFromRef(ref string) (int, error) {
```
golangcibot, line 434: block should not start with a whitespace (from `wsl`)

source/github/webhook.go
```diff
@@ -207,5 +216,52 @@ func processPREvent(h *library.Hook, payload *github.PullRequestEvent) (*library
...
+		Comment:  "",
+		PRNumber: 0,
+		Hook:     h,
+		Repo:     r,
+		Build:    b,
+	}, nil
+}
+
+// processIssueCommentEvent is a helper function to process the issue comment event
+func processIssueCommentEvent(h *library.Hook, payload *github.IssueCommentEvent) (*types.Webhook, error) {
```
golangcibot, line 229: block should not start with a whitespace (from `wsl`)

source/github/webhook_test.go
```diff
@@ -579,3 +581,79 @@ func TestGithub_VerifyWebhook_NoSecret(t *testing.T) {
...
+
+	if err != nil {
+		t.Errorf("ProcessWebhook returned err: %v", err)
+	}
+
+	if !reflect.DeepEqual(got, want) {
+		if diff := cmp.Diff(want, got); diff != "" {
+			t.Errorf("MakeGatewayInfo() mismatch (-want +got):\n%s", diff)
+		}
+		t.Errorf("ProcessWebhook webhook is %v, want %v", got, want)
```
golangcibot, line 657: expressions should not be cuddled with blocks (from `wsl`)

source/github/github.go
```diff
@@ -20,6 +20,13 @@ import (
 const (
 	defaultURL = "https://github.com"     // Default GitHub URL
 	defaultAPI = "https://api.github.com" // Default GitHub API URL
+
+	// events for repo webhooks
+	eventPush         = "push"
```
jbrockopp, line 25: Any reason why we are duplicating these events here? These variables are already declared in `go-vela/types/constants` 👍 

Specifically, you can find them here:

https://github.com/go-vela/types/blob/80de769c682b6b87d09b6a76935cacea8ca8d78f/constants/event.go#L8-L23

source/github/repo.go
```diff
@@ -117,9 +117,10 @@ func (c *client) Enable(u *library.User, org, name, secret string) (string, erro
 	// create the hook object to make the API call
 	hook := &github.Hook{
 		Events: []string{
-			constants.EventPush,
-			constants.EventPull,
-			constants.EventDeploy,
+			eventPush,
```
jbrockopp, line 120: Similar to above, we should use the variables declared in `go-vela/types/constants` 👍 

https://github.com/go-vela/types/blob/80de769c682b6b87d09b6a76935cacea8ca8d78f/constants/event.go#L8-L23

source/github/github.go
KellyMerrick: yes, actually. "comment" is the new event type we are using for repo and secret that we are rolling the multiple comment-related event types into (issue_comment, pull_request_review_comment, pull_request_review). we set it up this way to be easier to add the other pr related events to the "comment" event when we get to those stories (public project 0.4).

source/github/repo.go
kneal: The goal was to separate out the Vela specific syntax from the source provider. While `push` and `pull_request` match, the comment event does not match the approved proposal and other event types (labels, reactions, etc) we may choose to have a different syntax in Vela than what he source provider provides. 

Also, should help us when we look to add other providers like GitLab. When I was looking through their [documentation on events](https://docs.gitlab.com/ee/api/events.html) it appeared they also their events will be slightly different from GitHub when we do constants. So, we thought of moving towards a design where the provider is responsible for its provider's specific syntax would be valuable.

source/github/github.go
KellyMerrick: reminder: each place in a pr to add a comment will trigger a different event. from the user's perspective, they likely don't know/care that they are different events. this also helps minimize the number of events a repo and secret will need to have enabled for the same functionality.

source/github/repo.go
KellyMerrick: see my other comments above

source/github/github.go
jbrockopp: 👍 I see the difference now!

source/github/repo.go
jbrockopp: 👍 I see the difference now!

Issue #148, opened 2020-05-04
Github OAuth authorization failure


annamalai87:
## Description

During OAuth login using Github.com as a source (instead of Github Enterprise), code fails to construct correct URL which results in no user found while authorizing. This is happening because the source url provided in docker-compose.yml is has additional "/" in the end when compared to default url GitHub.com defined in the code.

https://github.com/go-vela/server/blob/e20520b752adef0f96bfeffb4bab148468ef3137/source/github/github.go#L21

https://github.com/go-vela/server/blob/e20520b752adef0f96bfeffb4bab148468ef3137/docker-compose.yml#L28

As docker-compose yml has additional "/" in the end,  server is authorizing the token using the incorrect url  https://github.com/api/v3/user instead of using https://api.github.com/user

## Value

Fixes Github login.

## Definition of Done

User should be able login using Github.

## Effort  (Optional)

1 days

## Impacted Personas (Optional)

Users

PR #149, opened 2020-05-05
fix(source): OAuth url fix

annamalai87:
This fixes https://github.com/go-vela/server/issues/148


source/github/github.go
```diff
@@ -18,8 +18,8 @@ import (
 )
 
 const (
-	defaultURL = "https://github.com"     // Default GitHub URL
-	defaultAPI = "https://api.github.com" // Default GitHub API URL
+	defaultURL = "https://github.com/"     // Default GitHub URL
```
annamalai87, line 21: When defaultURL and source url (from docker-compose.yml) are not same, we construct auth api url using source url which results in 404 error.
https://github.com/go-vela/server/blob/e20520b752adef0f96bfeffb4bab148468ef3137/source/github/github.go#L59

**defaultAPI should be used to authorize the Github user.**

PR #161, opened 2020-06-12
Error loop

jenmartTarget:


source/github/repo.go
```diff
@@ -27,15 +28,24 @@ func (c *client) Config(u *library.User, org, name, ref string) ([]byte, error)
 	opts := &github.RepositoryContentGetOptions{
 		Ref: ref,
 	}
-
-	// send API call to capture the .vela.yml pipeline configuration
 	data, _, resp, err := client.Repositories.GetContents(ctx, org, name, ".vela.yml", opts)
```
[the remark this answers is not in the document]
jenmartTarget: GetContents can't be called with declaring four variables.

source/github/repo.go
```diff
@@ -27,15 +28,24 @@ func (c *client) Config(u *library.User, org, name, ref string) ([]byte, error)
...
 	if err != nil {
-		if resp.StatusCode != http.StatusNotFound {
-			return nil, err
+		for i := 0; i < 5; i++ {
```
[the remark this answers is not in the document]
jenmartTarget: This sounds great! Could we have this as a future story?

source/github/repo.go
```diff
@@ -16,6 +17,17 @@ import (
 	"github.com/sirupsen/logrus"
 )
```
kneal, line 19: This function is missing its comment

source/source.go
```diff
@@ -74,6 +74,9 @@ type Service interface {
 	// Config defines a function that captures
 	// the pipeline configuration from a repo.
 	Config(user *library.User, org, name, ref string) ([]byte, error)
+	// ConfigBackoff defines a function wrapper that
+	// checks if yml/yaml file is valid.
```
kneal, line 78: We might want to think about adding wording noting the frequency or type (Exponential, Constant, Linear, etc) of backoff we do.

jenmartTarget: ConfigBackOff is a "truncated constant" backoff and is called five times every five seconds.

source/github/repo.go
jenmartTarget: Moved text fromMoved "ConfigBackoff defines a function wrapper that checks if yml/yaml file is valid" comment from source.go to here instead.

source/github/repo.go
```diff
@@ -16,6 +17,20 @@ import (
...
 
+// ConfigBackoff is a wrapper for config that checks if yaml/yml file is valid.
+func (c *client) ConfigBackoff(u *library.User, org, name, ref string) (data []byte, err error) {
```
jbrockopp, line 21: I'm not entirely sure the comment here is accurate so I think we'll have to update this a bit 👍 

At least, I have some concerns that this description could be seen as a bit misleading.

I say this because this `ConfigBackoff` function is currently a wrapper around the `Config` function.

The `Config` function doesn't actually verify the `.yaml`/`.yml` file (a.k.a. pipeline configuration) is valid.

The purpose of these functions is to get the pipeline configuration from the source repo stored on GitHub 🙂 

FWIW:

Validating the pipeline configuration is the responsibility of our compiler and get's called later on in the code base 👍

source/source.go
```diff
@@ -74,6 +74,10 @@ type Service interface {
 	// Config defines a function that captures
 	// the pipeline configuration from a repo.
 	Config(user *library.User, org, name, ref string) ([]byte, error)
+	// ConfigBackoff is a truncated constant backoff wrapper.
+	// Calls Config to check if yml/yaml is valid.
+	// Retries every five seconds and returns an error after five tries.
+	ConfigBackoff(user *library.User, org, name, ref string) ([]byte, error)
```
jbrockopp, line 80: Similar to my previous comment, I think we'll want to update the description to be a bit more accurate 👍

source/github/repo.go
```diff
@@ -16,6 +17,20 @@ import (
...
 
+// ConfigBackoff is a wrapper for Config that will retry five times if the function fails to retrieve the yaml/yml file.
+func (c *client) ConfigBackoff(u *library.User, org, name, ref string) (data []byte, err error) {
+	for i := 0; i < 5; i++ {
+		data, err = c.Config(u, org, name, ref)
+		if err != nil {
+			return
+		}
+
+		time.Sleep(5 * time.Second)
```
jbrockopp, line 28: I wonder if waiting `5s` between each API call is what we want here.

I think we could improve this by making it an exponential backoff period between each call 👍 

Something like:

```suggestion
		time.Sleep((i + 1) * time.Second)
```

source/github/repo.go
```diff
@@ -16,6 +17,20 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// ConfigBackoff is a wrapper for Config that will retry five times if the function fails to retrieve the yaml/yml file.
+func (c *client) ConfigBackoff(u *library.User, org, name, ref string) (data []byte, err error) {
+	for i := 0; i < 5; i++ {
```
davidvader, line 22: It's not necessary, but having a variable for retry limit and sleep time would make it cleaner and easier to update if we want to tweak those values

source/github/repo.go
wass3r: agreed. i think most of the time, a minor delay will probably do the job, so a jump to 5s immediately seems a bit steep.

source/github/repo.go
```diff
@@ -16,6 +17,20 @@ import (
...
 
+// ConfigBackoff is a wrapper for Config that will retry five times if the function fails to retrieve the yaml/yml file.
+func (c *client) ConfigBackoff(u *library.User, org, name, ref string) (data []byte, err error) {
+	retryLimit := 5
+	for i := 0; i < retryLimit; i++ {
+		data, err = c.Config(u, org, name, ref)
+		if err != nil {
+			return
+		}
+		time.Sleep(time.Duration(i+1) * time.Second)
```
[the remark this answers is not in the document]
wass3r: unless i'm mistaken this is still an issue?

davidvader:
I was looking back at this and I realized that it might need a small tweak.
If the data is valid on the first pass over, isn't it still going to retry 5 times?

davidvader:
proposed some changes https://github.com/go-vela/server/pull/163/files

PR #165, opened 2020-07-10
provides functionality of login using github token

justsimplify:
It only replaces that rather than getting the `authorization token` from the credentials provided, if we just pass the `token`, then we can eliminate the process of getting the token first and getting info. Just use the token provided and get the info from that token.

### DIFF
--- a/api/authenticate.go
+++ b/api/authenticate.go
@@ -184,7 +184,7 @@ func AuthenticateCLI(c *gin.Context) {
 	}
 
 	// register user with OAuth application
-	newUser, err := source.FromContext(c).LoginCLI(input.GetUsername(), input.GetPassword(), input.GetOTP())
+	newUser, err := source.FromContext(c).LoginCLI(input.GetUsername(), input.GetPassword(), input.GetOTP(), input.GetToken())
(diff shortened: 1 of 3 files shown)


wass3r:
@justsimplify sorry, for the _very_ delayed response. thank you for taking the time to submit this PR. we do have some changes coming soon that will make the edits proposed here *very* short-lived since some of the code you are touching will be gone. we have opened a new issue with an alternative implementation request: https://github.com/go-vela/community/issues/126 . if you're willing to rework your PR to accommodate that issue, we'd appreciate it. otherwise, we will probably close this request by the end of the week and look to complete the work for the mentioned issue afterwards.

wass3r (changes requested):
see last comment to request implementation of https://github.com/go-vela/community/issues/126

justsimplify:
@wass3r Sure will get that moving.

justsimplify:
@wass3r https://github.com/go-vela/community/issues/126#issuecomment-721607154
Would like you to take a look at this and have your thoughts as well.

justsimplify:
Not needed.

PR #166, opened 2020-07-15
fix(webhook): fix issue for push events without changeset

wass3r:
- fixes assumption made in code that push events without changeset (ie. no files changed) are pull requests
- populate pull request number on webhook struct being returned when processing the webhook (where applicable)
- accounts for comment events that *do not* happen on pull requests
- add test case for non-pull-request comments
- adjust existing tests to account for `PRNumber` being populated now


api/webhook.go
```diff
@@ -335,31 +324,20 @@ func PostWebhook(c *gin.Context) {
...
-				return
-			}
-
-			// send API call to capture list of files changed for the pull request
-			files, err = source.FromContext(c).ChangesetPR(u, r, number)
-			if err != nil {
-				retErr := fmt.Errorf("%s: failed to get changeset for %s: %v", baseErr, r.GetFullName(), err)
-				util.HandleError(c, http.StatusInternalServerError, retErr)
+	// check if the build event is a pull_request
+	if strings.EqualFold(b.GetEvent(), constants.EventPull) && webhook.PRNumber != 0 {
```
wass3r, line 330: This was the main culprit. Previously, at this point, if we were dealing with a push event that had no changeset (which was processed above) we would end up here with `len(files)` being zero. We would then continue processing the request as if it was a pull request causing `500` errors while trying to extract pull request info that doesn't exist.

api/webhook.go
```diff
@@ -257,19 +256,9 @@ func PostWebhook(c *gin.Context) {
...
-
-			h.SetStatus(constants.StatusFailure)
-			h.SetError(retErr.Error())
-
-			return
-		}
-
-		commit, branch, baseref, err := source.FromContext(c).GetPullRequest(u, r, number)
+	// if this is a comment on a pull_request event
+	if strings.EqualFold(b.GetEvent(), constants.EventComment) && webhook.PRNumber != 0 {
```
wass3r, line 260: We are now passing through pull request number on the `webhook` struct and are using that instead of calling the utility function to strip it from the ref.

source/github/webhook.go
```diff
@@ -218,10 +218,11 @@ func processPREvent(h *library.Hook, payload *github.PullRequestEvent) (*types.W
 	}
 
 	return &types.Webhook{
-		Comment: "",
-		Hook:    h,
-		Repo:    r,
-		Build:   b,
+		Comment:  "",
+		PRNumber: payload.GetNumber(),
```
wass3r, line 222: Now passing through the pull request number with the struct we return to avoid using utility function later to extract that info.

source/github/webhook.go
```diff
@@ -319,12 +320,23 @@ func processIssueCommentEvent(h *library.Hook, payload *github.IssueCommentEvent
...
+		pr = payload.GetIssue().GetNumber()
+	}
 
 	return &types.Webhook{
-		Comment: payload.GetComment().GetBody(),
-		Hook:    h,
-		Repo:    r,
-		Build:   b,
+		Comment:  payload.GetComment().GetBody(),
+		PRNumber: pr,
```
wass3r, line 337: Passing through 0 (zero) for normal issue comments, and the actual pull request number for pull request comments. Will help avoid using utility function to extract pull request number later in path.

source/github/webhook.go
```diff
@@ -319,12 +320,23 @@ func processIssueCommentEvent(h *library.Hook, payload *github.IssueCommentEvent
 	b.SetSender(payload.GetSender().GetLogin())
 	b.SetAuthor(payload.GetIssue().GetUser().GetLogin())
 	b.SetEmail(payload.GetIssue().GetUser().GetEmail())
-	b.SetRef(fmt.Sprintf("refs/pull/%d/head", payload.GetIssue().GetNumber()))
+	// treat as non-pull-request comment by default and
+	// set ref to default branch for the repo
+	b.SetRef(fmt.Sprintf("refs/heads/%s", r.GetBranch()))
```
wass3r, line 325: Open to other options here. Pull request ref doesn't make sense for normal issue comments, though.

api/webhook.go
```diff
@@ -466,26 +444,3 @@ func publishToQueue(queue queue.Service, p *pipeline.Build, b *library.Build, r
 		}
 	}
 }
-
-// helper function to parse the PR number out of a build ref
-func parsePRNumberFromRef(ref string) (int, error) {
```
wass3r, line 471: Removing utility function as the information is now passed in.

api/webhook.go
```diff
@@ -335,31 +324,20 @@ func PostWebhook(c *gin.Context) {
 				return
 			}
 		}
+	}
 
-		// files is empty if the build event is pull_request
-		if len(files) == 0 {
```
wass3r, line 340: (commenting to make old code visible in conversation thread)

source/github/webhook.go
jbrockopp: I think what you have makes sense 👍 

Ideally, if they are commenting on a normal issue, we should inherit the "default" branch from the repo.

The only thing I can think of is maybe in a future iteration we allow them to provide a ref/branch/commit etc via some specially formatted comment? 🤷‍♂️ 

But that's way beyond the scope of what you have here!

source/github/webhook.go
jbrockopp: Good idea 👍

source/github/webhook.go
jbrockopp: Good idea 👍

PR #170, opened 2020-07-27
feat(token): change to use expiring tokens

wass3r:
This PR implements https://github.com/go-vela/community/issues/29, notably:
- add refreshtoken to user in database to allow revoking and for looking up a user
- adds ability to log in via web ui, cli, or headless-ly using the same oauth app
- modified login to allow for different auth paths
- add token refresh endpoint
- ability to set access and refresh token expirations

note: golangci-lint errors are not pertaining to my code changes but old code


cmd/vela-server/main.go
```diff
@@ -60,6 +60,21 @@ func main() {
 			Value:   false,
 		},
 
+		// Security Flags
+
+		&cli.DurationFlag{
+			EnvVars: []string{"VELA_ACCESS_TOKEN_DURATION", "ACCESS_TOKEN_DURATION"},
+			Name:    "access-token-duration",
+			Usage:   "sets the duration of the access token",
+			Value:   15 * time.Minute,
```
davidvader, line 69: just wondering, is there any significance to choosing 15 minutes?

wass3r: none, it's just a default. easy to override. i think the point is just to make the token useful for a small time window in case it gets compromised.

api/authenticate.go
```diff
@@ -194,110 +189,46 @@ func Authenticate(c *gin.Context) {
...
+func AuthenticateType(c *gin.Context) {
+	// load the metadata
+	m := c.MustGet("metadata").(*types.Metadata)
 
-		// compose JWT token for user
-		t, err := token.Compose(u)
-		if err != nil {
-			retErr := fmt.Errorf("unable to compose token for user %s: %w", u.GetName(), err)
+	t := c.Param("type")
+	p := c.Param("port")
```
davidvader, line 215: Does port have a default value?

api/logout.go
```diff
@@ -0,0 +1,53 @@
...
+		// silently fail
+		// TODO: reconsider?
+		return
+	}
+
+	// remove the refresh token from the cookies, Max-Age value -1 will do it
+	c.SetCookie(constants.RefreshTokenName, "", -1, "/", addr.Hostname(), false, true)
+
+	// return 200 for successful logout
+	c.JSON(http.StatusOK, "ok")
```
davidvader, line 52: Is there a reason we don't clear the user's refresh token in the database? Unnecessary?

wass3r: i didn't see this comment initially, but yes it should. i think it's handled in the newer pushes, though this comment doesn't show as outdated.

jbrockopp (approved):
LGTM

kneal (approved):
LGTM 🐬

PR #178, opened 2020-08-26
fix(secrets): properly set default value for allow_command

JordanSussman:
The current implementation will incorrectly update the `AllowCommand` to be `true` even if the user supplies `false`. For example: 

```
vela add secret --secret.engine vault --secret.type repo --org JordanSussman --repo vela-local-test --name test --value bar --event comment --event deployment --event pull_request --event push --event tag --commands=false
{
	AllowCommand: true,
	Events: [comment deployment pull_request push tag],
	ID: 0,
	Images: [],
	Name: test,
	Org: JordanSussman,
	Repo: vela-local-test,
	Team: ,
	Type: repo,
	Value: [secure],
}
```

### DIFF
--- a/api/secret.go
+++ b/api/secret.go
@@ -120,7 +120,7 @@ func CreateSecret(c *gin.Context) {
 		input.SetEvents(defaultEvents)
 	}
 
-	if !input.GetAllowCommand() {
+	if input.AllowCommand == nil {
 		input.SetAllowCommand(true)
 	}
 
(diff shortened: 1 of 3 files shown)


api/secret.go
```diff
@@ -120,7 +120,7 @@ func CreateSecret(c *gin.Context) {
 		input.SetEvents(defaultEvents)
 	}
 
-	if !input.GetAllowCommand() {
+	if input.AllowCommand == nil {
```
JordanSussman, line 123: Set to true only when user has not defined any value.

PR #179, opened 2020-09-01
fix(api): setting active flag for repo

jbrockopp:
See https://github.com/go-vela/cli/pull/163 for more information.

This PR updates the `active` flag to have a default value of `true` **only** if it hasn't been provided in the request body 👍

### DIFF
--- a/api/repo.go
+++ b/api/repo.go
@@ -96,7 +96,7 @@ func CreateRepo(c *gin.Context) {
 	// update fields in repo object
 	input.SetUserID(u.GetID())
 
-	if input.Active != nil {
+	if input.Active == nil {
 		input.SetActive(true)
 	}
 
--- a/api/repo.go
+++ b/api/repo.go
@@ -96,7 +96,7 @@ func CreateRepo(c *gin.Context) {
 	// update fields in repo object
 	input.SetUserID(u.GetID())
 
-	if !input.GetActive() {
+	if input.Active != nil {
 		input.SetActive(true)
 	}
 


JordanSussman:
Should https://sourcegraph.com/github.com/go-vela/server/-/blob/api/repo.go#L119-121 also be updated?

jbrockopp:
@JordanSussman I don't believe so 👍 

Those lines of code are setup to ensure that a repo can't end up with no events enabled.

There is a bit more to it than that because if a repo is provided with no events enabled, it will set the "default" (`push` and `pull_request`) events to be enabled.

But maybe seeing a snippet of code would help see if there is a better way to handle this?

wass3r:
@JordanSussman did you mean to link to: https://sourcegraph.com/github.com/go-vela/server/-/blob/api/repo.go#L99-101 ?

i think those are the lines i pointed out in the other issue.

JordanSussman:
> @JordanSussman I don't believe so 👍
> 
> Those lines of code are setup to ensure that a repo can't end up with no events enabled.
> 
> There is a bit more to it than that because if a repo is provided with no events enabled, it will set the "default" (`push` and `pull_request`) events to be enabled.
> 
> But maybe seeing a snippet of code would help see if there is a better way to handle this?

Was initially thinking that could could allow the value of `nil` thus allowing the check to be invalidated, but I validated that the `Get<thing>` function returns false if the underlying value is `nil` so it should be fine as is.

JordanSussman (approved):

meinenec (approved):

PR #180, opened 2020-09-09
feat(metrics): support collecting service statuses

jbrockopp:
This replicates the functionality for `services` that already exists for other resources on the `/metrics` endpoint 👍 

Currently on the `/metrics` endpoint, we expose a breakdown of all the different `status` fields recorded for an object in the database as well as how many objects have that `status`.

However, we're only doing this for `builds` and `steps` resources. An example from my local stack:

```
vela_totals{field="status",resource="build",value="error"} 0
vela_totals{field="status",resource="build",value="failed"} 2
vela_totals{field="status",resource="build",value="killed"} 0
vela_totals{field="status",resource="build",value="pending"} 0
vela_totals{field="status",resource="build",value="running"} 0
vela_totals{field="status",resource="build",value="success"} 4
vela_totals{field="status",resource="steps",value="failure"} 2
vela_totals{field="status",resource="steps",value="killed"} 10
vela_totals{field="status",resource="steps",value="success"} 12
```

This PR adds that same functionality for `services` 🎉


meinenec (approved):

api/metrics.go
```diff
@@ -119,19 +119,24 @@ func recordGauges(c *gin.Context) {
...
 	if err != nil {
-		logrus.Errorf("Error while reading all error builds: %v", err)
+		logrus.Errorf("Error while reading all step images: %v", err)
 	}
 
-	stepImageMap, err := database.FromContext(c).GetStepImageCount()
+	stepStatusMap, err := database.FromContext(c).GetStepStatusCount()
 	if err != nil {
-		logrus.Errorf("Error while reading all images: %v", err)
+		logrus.Errorf("Error while reading all step statuses: %v", err)
```
kneal, line 129: I'm guessing this is similar error handling to the others, but does it make sense to fix them now?

I believe we do not capitalize the first word and went are from stating error in an error because it's redundant.

jbrockopp: Yeah I thought about fixing them but wasn't sure if it made sense to do it in this PR due to scope 🤷 

I'll just update them to be safe 👍

kneal: Thanks!

jbrockopp: @kneal fixed! 🎉

wass3r (approved):

kneal (approved):
LGTM 🐬

PR #183, opened 2020-09-22
feat: add worker table

meinenec:
test by running locally (`make up`) and inserting into the database with:
```INSERT INTO workers VALUES (1, 'worker_0', 'localhost:8080', 'foo,bar,baz', true, 12345);```

dependent on go-vela/types/pull/109

adding api endpoints next in a separate PR.

### CODE REVIEW

File: database/ddl/postgres/worker.go [hunk 1]
```diff
@@ -0,0 +1,41 @@
...
+IF NOT EXISTS
+workers (
+	id					SERIAL PRIMARY KEY,
+	hostname			VARCHAR(250),
+	address				VARCHAR(250),
+	routes				VARCHAR(1000),
+	active     			BOOLEAN,
+	last_checked_in		INTEGER,
+	UNIQUE(hostname, address)
+);
```
jbrockopp, line 21: Could we fix the spacing here?

```suggestion
CREATE TABLE
IF NOT EXISTS
workers (
	id              SERIAL PRIMARY KEY,
	hostname        VARCHAR(250),
	address         VARCHAR(250),
	routes          VARCHAR(1000),
	active          BOOLEAN,
	last_checked_in INTEGER,
	UNIQUE(hostname, address)
);
```

database/ddl/sqlite/worker.go
```diff
@@ -0,0 +1,40 @@
...
+CREATE TABLE
+IF NOT EXISTS
+Workers (
+	id            INTEGER PRIMARY KEY AUTOINCREMENT,
+	hostname		TEXT,
+	address			TEXT,
+	routes          TEXT,
+	active          TEXT,
+	last_checked_in	INTEGER
+);
```
jbrockopp, line 20: Could we fix the spacing here?

Also the name of the table should be `workers` instead of `Workers` 👍 

```suggestion
CREATE TABLE
IF NOT EXISTS
workers (
	id              INTEGER PRIMARY KEY AUTOINCREMENT,
	hostname        TEXT,
	address         TEXT,
	routes          TEXT,
	active          TEXT,
	last_checked_in INTEGER
);
```

jbrockopp (changes requested):

database/ddl/sqlite/worker.go
```diff
@@ -0,0 +1,40 @@
...
+
+package sqlite
+
+const (
+	// CreateWorkerTable represents a query to
+	// create the workers table for Vela.
+	CreateWorkerTable = `
+CREATE TABLE
+IF NOT EXISTS
+Workers (
```
jbrockopp, line 13: This should be lowercase `workers` 👍 

```suggestion
workers (
```

jbrockopp (changes requested):

jbrockopp (approved):
LGTM

database/worker.go
```diff
@@ -0,0 +1,137 @@
...
+	// variable to store query results
+	var w []int64
+
+	// send query to the database and store result in variable
+	err := c.Database.
+		Table(constants.TableWorker).
+		Raw(c.DML.WorkerService.Select["count"]).
+		Pluck("count", &w).Error
+
+	return w[0], err
```
davidvader, line 87: (i know it exists elsewhere in the codebase so feel free to resolve this)
But if the Database.Table.Raw.Pluck fails will (L87) cause a segfault? Should we check if len(w) == 0 ?

meinenec: I think we should check that len is 0 or that err != nil. I'll make that change

kneal (approved):
LGTM 🐬

database/ddl/sqlite/worker.go
```diff
@@ -0,0 +1,40 @@
...
+	last_checked_in	INTEGER
+);
+`
+
+	// CreateWorkersHostnameAddressIndex represents a query to create an
+	// index on the repos table for the hostname and address columns.
+	CreateWorkersHostnameAddressIndex = `
+CREATE INDEX
+IF NOT EXISTS
+secrets_type_org_repo
```
wass3r, line 28: probably want to rename this.. maybe `workers_hostname_address`?

database/ddl/sqlite/worker.go
```diff
@@ -0,0 +1,40 @@
...
+	hostname        TEXT,
+	address         TEXT,
+	routes          TEXT,
+	active          TEXT,
+	last_checked_in	INTEGER
+);
+`
+
+	// CreateWorkersHostnameAddressIndex represents a query to create an
+	// index on the repos table for the hostname and address columns.
```
wass3r, line 24:
```suggestion
	// index on the workers table for the hostname and address columns.
```

database/ddl/sqlite/worker.go
```diff
@@ -0,0 +1,40 @@
...
+	CreateWorkersHostnameAddressIndex = `
+CREATE INDEX
+IF NOT EXISTS
+secrets_type_org_repo
+ON workers (type, org, repo);
```
wass3r, line 29:
```suggestion
ON workers (hostname, address);
```

database/ddl/sqlite/worker.go
```diff
@@ -0,0 +1,40 @@
...
+`
+)
+
+// createSecretService is a helper function to return
```
wass3r, line 33:
```suggestion
// createWorkerService is a helper function to return
```

database/ddl/sqlite/worker.go
```diff
@@ -0,0 +1,40 @@
...
+// a service for interacting with the secrets table.
```
wass3r, line 34:
```suggestion
// a service for interacting with the workers table.
```

database/ddl/postgres/worker.go
```diff
@@ -0,0 +1,41 @@
...
+	address          VARCHAR(250),
+	routes           VARCHAR(1000),
+	active           BOOLEAN,
+	last_checked_in  INTEGER,
+	UNIQUE(hostname, address)
+);
+`
+
+	// CreateWorkersHostnameAddressIndex represents a query to create an
+	// index on the repos table for the hostname and address columns.
```
wass3r, line 25:
```suggestion
	// index on the workers table for the hostname and address columns.
```

wass3r (changes requested):

wass3r (dismissed):

meinenec:
latest commit changed `UNIQUE(hostname, address)` to `UNIQUE(hostname)`

`UNIQUE(hostname, address)` made it so the combined fields of hostname and address were unique which would allow a worker to be registered if the host was the same, but the address was different. My initial understanding was that both would need to be unique. After thinking on it, I believe we only care that the hostname is unique.

wass3r (approved):

jbrockopp (approved):
LGTM

PR #185, opened 2020-09-23
fix(build): restart on build without changeset

wass3r:
fixes https://github.com/go-vela/community/issues/116

we have applied a similar fix in https://github.com/go-vela/server/pull/166 and missed a couple of instances. applied it in `RestartBuild` and then also noticed it in `CreateBuild`.


jbrockopp (approved):
LGTM

davidvader (approved):
LGTM!

PR #190, opened 2020-09-29
feat(vault): add logic to obtain/refresh token with aws auth method

JordanSussman:
This code should be fairly easy to modify down the road incase someone would like to add another [auth method](https://www.vaultproject.io/docs/auth) to obtain the token.

closes https://github.com/go-vela/community/issues/106


cmd/vela-server/secret.go
```diff
@@ -51,5 +51,5 @@ func setupNative(c *cli.Context, d database.Service) (secret.Service, error) {
 // helper function to setup the Vault secret engine from the CLI arguments.
 func setupVault(c *cli.Context) (secret.Service, error) {
 	logrus.Tracef("Creating %s secret client from CLI configuration", constants.DriverVault)
-	return vault.New(c.String("vault-addr"), c.String("vault-token"), c.String("vault-version"), c.String("vault-prefix"))
+	return vault.New(c.String("vault-addr"), c.String("vault-token"), c.String("vault-version"), c.String("vault-prefix"), c.String("vault-auth-method"), c.String("vault-aws-role"), c.Duration("vault-renewal"))
```
kneal, line 54: If we can think of a way to pass a struct vs values I think that would be more ideal. 

Values are good, just from a code readability standpoint, it feels better to pass a struct when you get into instances where functions have more than 4-5 parameters.

kneal (approved):

cmd/vela-server/main.go
```diff
@@ -157,6 +157,22 @@ func main() {
...
+			Name:    "vault-auth-method",
+			Usage:   "auth method to utilize to obtain token",
+		},
+		&cli.StringFlag{
+			EnvVars: []string{"VELA_SECRET_VAULT_AWS_ROLE", "SECRET_VAULT_AWS_ROLE"},
+			Name:    "vault-aws-role",
+			Usage:   "vault role to connect to the auth/aws/login endpoint with",
+		},
+		&cli.DurationFlag{
+			EnvVars: []string{"VELA_SECRET_VAULT_RENEWAL", "VELA_SECRET_VAULT_RENEWAL"},
```
jbrockopp, line 171: Should these be the same?

Based off your pattern above, I'm thinking it should look like:

```suggestion
			EnvVars: []string{"VELA_SECRET_VAULT_RENEWAL", "SECRET_VAULT_RENEWAL"},
```

cmd/vela-server/secret.go
jbrockopp: It'd be great to come up with a standard for this in the Vela codebase 👍 

Something like:

> If a function takes more than 3 parameters of the same type, we should make it accept a struct instead.

And maybe we don't even care about the types of the parameters, and focus on the sheer number of parameters?

jbrockopp (changes requested):

cmd/vela-server/secret.go
JordanSussman: Update to utilize a `vault.Config` struct

cmd/vela-server/main.go
JordanSussman: updated

jbrockopp (approved):
LGTM

wass3r (approved):

wass3r:
not sure what's going with the `checks`, but the last commit showed all green.. merging.

PR #192, opened 2020-10-09
fix(vault): skip auth method if token is provided

jbrockopp:
Currently, when spinning up the Vela server, an error message is thrown:

```sh
{"level":"fatal","msg":"vault auth method of '' is unsupported","time":"2020-10-09T16:07:06Z"}
```

After some digging, it appears we're always expecting the `vault-auth-method` flag to be provided, even if the user specifies a `vault-token` flag.

I believe this PR addresses the issue by only checking the `vault-auth-method` flag if no `vault-token` flag is provided.


meinenec (approved):

JordanSussman (approved):

kneal (approved):
LGTM 🐬

PR #193, opened 2020-10-14
feat: add api endpoints for managing the worker table

meinenec:


JordanSussman:
Do you intend to add something to periodically prune the list of workers if they have not checked in for x interval? I fear the data will grow unbound in autoscaling scenarios unless something is clearing up stale data.

router/worker.go
```diff
@@ -0,0 +1,38 @@
...
+// GET    /api/v1/workers
+// GET    /api/v1/workers/:worker
+// PUT    /api/v1/workers/:worker
+// DELETE /api/v1/workers/:worker
+func WorkerHandlers(base *gin.RouterGroup) {
+	// Workers endpoints
+	workers := base.Group("/workers")
+	{
+		workers.POST("", perm.MustPlatformAdmin(), middleware.Payload(), api.CreateWorker)
+		workers.GET("", api.GetWorkers)
```
jbrockopp, line 28: I noticed we aren't using the `perm.MustPlatformAdmin()` function here.

This means that any user within Vela will be able to capture this information 👍 

At this point, I can't come up with any good reasons why we wouldn't want to allow this so my comment here is more of a food for thought looking forward to the future.

If, at any point, we decide to embed more sensitive information into the `workers` table, we might have to restrict this to only being accessible by platform admins.

router/worker.go
```diff
@@ -0,0 +1,38 @@
...
+
+		// Worker endpoints
+		w := workers.Group("/:worker")
+		{
+			w.GET("", worker.Establish(), api.GetWorker)
```
jbrockopp, line 33: Similar thought here 👍 

Can't come up with any reason to block this so I think we're good for now.

jbrockopp (approved):
LGTM

router/worker.go
meinenec: That was my thought process as well. I'm open to changing it if we have concerns with other features we may want to add.

meinenec:
@JordanSussman That does need to be implemented but left that out for the initial implementation. I realize an issue hasn't been made to create any sort of clean-up so I'll make that now.

wass3r:
sorry, :disappear: .. fix conflict one more time? :(

wass3r (approved):

router/worker.go
kneal: My only concern with it not being plat admin would be some type of security attack vector. The only type I can really think of is some type of social engineering attack. None of the fields within that type stand out as something I'm super concerned with but I think depending on how that grows. Security would be my continued concern with not being a plat admin

kneal (approved):
LGTM 🐬

jbrockopp (approved):
LGTM

PR #196, opened 2020-10-16
feat: use new version pkg

jbrockopp:
Dependent on https://github.com/go-vela/types/pull/112

Updating to the new `go-vela/types/version` package to improve the information provided surrounding the application.

If I compile the binary specifically for the OS on my workstation, I'm met with this from the `--version` flag:

```sh
$ release/vela-server --version
vela-server version v0.6.1
```

When I run `make up` locally, here is the output from http://localhost:8080/version:

```json
{
  "canonical": "v0.6.1",
  "major": 0,
  "minor": 6,
  "patch": 1,
  "metadata": {
    "architecture": "amd64",
    "build_date": "2020-10-16T09:15:43Z",
    "compiler": "gc",
    "git_commit": "44c10cfa6b25f02721609cccd2fbcedea0109f7b",
    "go_version": "go1.15.2",
    "operating_system": "linux"
  }
}
```

**NOTE: This also updates the GitHub Actions pipelines to use our `make` commands for building the binary.**


wass3r (approved):

router/middleware/header.go
```diff
@@ -65,26 +65,24 @@ func Cors(c *gin.Context) {
...
+	v := version.New()
 
 	if gin.Mode() == "debug" {
-		c.Request.Header.Set("X-Vela-Version", apiVersion.String())
+		c.Request.Header.Set("X-Vela-Version", v.Semantic())
 	} else { // in prod we don't want the build number metadata
-		apiVersion.Metadata = ""
-		c.Request.Header.Set("X-Vela-Version", apiVersion.String())
+		c.Request.Header.Set("X-Vela-Version", v.Semantic())
 	}
```
wass3r, line 74: do we need this if statement if we're returning the same thing?

wass3r (changes requested):

router/middleware/header.go
jbrockopp: No we don't need the if statement 👍 

If you'd like I can comment out the `if` but I left it in favor of being explicit on how we could customize it in the future.

In short, my thought process was by leaving it there, if (or when) we want to look at customizing how that header is populated from the API when running in `debug` mode then the shell of the code is there.

wass3r (approved):

router/middleware/header.go
wass3r: k, not a big deal to leave it in 👍

kneal (approved):
LGTM 🐬

PR #199, opened 2020-10-19
chore(docker): publish alpine images

JordanSussman:
Adding another `Dockerfile` based on alpine since I unfortunately need access to some image with a shell to allow me to modify the environment variables before the binary is started. My specific use case involves needing to generate the `VELA_DATABASE_CONFIG` and `VELA_QUEUE_CONFIG` environment variables utilizing others that are injected at run time.

## changes

- Add `Dockerfile-alpine` file to utilize alpine as base image instead of scratch
- Update `.github/workflows/publish.yml` to publish `latest-alpine` image
- Update `.github/workflows/prerelease.yml` to publish `${GITHUB_TAG}-latest` image

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -37,3 +37,13 @@ jobs:
         tag_names: true
         username: ${{ secrets.DOCKER_USERNAME }}
         password: ${{ secrets.DOCKER_PASSWORD }}
+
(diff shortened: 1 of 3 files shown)


kneal (approved):
LGTM 🐬

jbrockopp (approved):
LGTM

PR #200, opened 2020-10-20
enhancement: pull repo info during repo creation

wass3r:
This change will query the source provider for repo information during the repo activation process. It will short circuit the process if the request fails. The change allows us to utilize the information returned instead of manually setting it and/or making assumptions such as what the default branch is for a given repo.

This change entails:
1. Adds new source interface function `GetRepo()` for retrieving repo information
1. Uses `GetRepo()` in `CreateRepo()` code path
1. Bug fix for source `Enable()` function
1. Linter and consistency fixes

Considerations:
1. The change introduces another API call to the source provider that will go against any rate limits that are set on the source provider.
1. I opted not to override what is returned for (default)branch from the source provider even if the user provided _branch_ as input for repo activation. We could check to see if `GetBranch` is empty after we make the call to the source provider and override it in that case - seems unlikely?


api/repo.go
```diff
@@ -90,87 +91,90 @@ func CreateRepo(c *gin.Context) {
 		return
 	}
 
+	// get repo information from the source
+	r, err := source.FromContext(c).GetRepo(u, input)
```
wass3r, line 95: This is the main change. After we parse the given input, we get the repo information from the source. We will use the returned repo struct as a building block for what will eventually be activated.

api/repo.go
```diff
@@ -90,87 +91,90 @@ func CreateRepo(c *gin.Context) {
...
 	if len(input.GetVisibility()) == 0 {
-		input.SetVisibility(constants.VisibilityPublic)
-	}
-
-	if len(input.GetFullName()) == 0 {
-		input.SetFullName(fmt.Sprintf("%s/%s", input.GetOrg(), input.GetName()))
-	}
-
-	if len(input.GetBranch()) == 0 {
-		input.SetBranch("master")
```
wass3r, line 113: These properties are are now returned via `GetRepo` as part of the successful response from the source.

api/repo.go
```diff
@@ -90,87 +91,90 @@ func CreateRepo(c *gin.Context) {
...
 	if err != nil {
 		retErr := fmt.Errorf("unable to create webhook for %s: %w", r.GetFullName(), err)
 
 		switch err.Error() {
-		case "Repo already enabled":
+		case "repo already enabled":
 			util.HandleError(c, http.StatusConflict, retErr)
 			return
-		case "Repo not found":
+		case "repo not found":
```
wass3r, line 177: This is actually a bug fix. `.Enable()` returns the error all lower case (see https://github.com/go-vela/server/blob/v0.6.1/source/github/repo.go#L175-L180). Currently, if there's an error happening at this point the server will return a 5xx error instead of a more accurate 4xx in cases such as enabling a non-existing repo.

api/repo.go
```diff
@@ -180,43 +184,36 @@ func CreateRepo(c *gin.Context) {
...
 	}
 
-	// TODO: build these from the source client
-	if len(input.GetLink()) == 0 {
-		input.SetLink(url)
-	}
-
-	if len(input.GetClone()) == 0 {
-		input.SetClone(fmt.Sprintf("%s.git", url))
-	}
```
wass3r, line 190: These properties are are now returned via `GetRepo` as part of the successful response from the source.

source/github/repo.go
```diff
@@ -19,13 +19,13 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// ConfigBackoff is a wrapper for Config that will retry five times if the function fails to retrieve the yaml/yml file.
+// ConfigBackoff is a wrapper for Config that will retry five times if the function
+// fails to retrieve the yaml/yml file.
```
wass3r, line 23: Fixes linter (long line length) issue.

source/github/repo.go
```diff
@@ -173,9 +173,9 @@ func (c *client) Enable(u *library.User, org, name, secret string) (string, erro
 	_, resp, err := client.Repositories.CreateHook(ctx, org, name, hook)
 
 	switch resp.StatusCode {
-	case 422:
+	case http.StatusUnprocessableEntity:
 		return "", fmt.Errorf("repo already enabled")
-	case 404:
+	case http.StatusNotFound:
```
wass3r, line 178: Fixes linter (magic number) issue.

source/github/repo.go
```diff
@@ -275,6 +275,22 @@ func (c *client) Status(u *library.User, b *library.Build, org, name string) err
 	return err
 }
 
+// GetRepo gets repo information from Github.
+func (c *client) GetRepo(u *library.User, r *library.Repo) (*library.Repo, error) {
```
wass3r, line 279: Added a new function to the GitHub source package to retrieve repo information.

source/github/repo_test.go
```diff
@@ -747,6 +747,94 @@ func TestGithub_Status_Error(t *testing.T) {
 	}
 }
 
+func TestGithub_GetRepo(t *testing.T) {
```
wass3r, line 750: Add happy path test.

source/github/repo_test.go
```diff
@@ -747,6 +747,94 @@ func TestGithub_Status_Error(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("GetRepo returned err: %v", err)
+	}
+
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("GetRepo is %v, want %v", got, want)
+	}
+}
+
+func TestGithub_GetRepo_Fail(t *testing.T) {
```
wass3r, line 803: Add unhappy path test :)

source/source.go
```diff
@@ -73,11 +73,11 @@ type Service interface {
 
 	// Config defines a function that captures
 	// the pipeline configuration from a repo.
-	Config(user *library.User, org, name, ref string) ([]byte, error)
+	Config(*library.User, string, string, string) ([]byte, error)
```
wass3r, line 76: Changed to stay consistent.

source/source.go
```diff
@@ -73,11 +73,11 @@ type Service interface {
...
 	// ConfigBackoff is a truncated constant backoff wrapper for Config.
 	// Retry again in five seconds if Config fails to retrieve yaml/yml file.
 	// Will return an error after five failed attempts.
-	ConfigBackoff(user *library.User, org, name, ref string) ([]byte, error)
+	ConfigBackoff(*library.User, string, string, string) ([]byte, error)
```
wass3r, line 80: Changed for consistency.

source/source.go
```diff
@@ -93,6 +93,9 @@ type Service interface {
 	// GetPullRequest defines a function that retrieves
 	// a pull request for a repo.
 	GetPullRequest(*library.User, *library.Repo, int) (string, string, string, string, error)
+	// GetRepo defines a function that retrieves
+	// details for a repo.
+	GetRepo(*library.User, *library.Repo) (*library.Repo, error)
```
wass3r, line 98: Adds new `GetRepo` function to source interface.

wass3r:
CI status is being misreported on main PR page. Check Actions tab for proper results - https://github.com/go-vela/server/actions?query=branch%3Aenhance%2Fcreate-repo :(

api/repo.go
wass3r: This is reporting as "Outdated" because the location moved due to changes happening via another merged change. The comment is still valid, however.

api/repo.go
wass3r: This is reporting as "Outdated" because the location moved due to changes happening via another merged change. The comment is still valid, however.

jbrockopp (approved):
LGTM

davidvader (approved):
LGTM. nice work

PR #203, opened 2020-10-22
feat: add pipeline endpoints

KellyMerrick:
Part of the [Cargill/Target Hackathon](https://gophers.slack.com/archives/CNRRKE8KY/p1602194963000100) for Vela

This adds a collection of new `/api/v1/pipelines` endpoints:

```
GET   /api/v1/pipelines/:org/:repo
POST  /api/v1/pipelines/:org/:repo/compile
POST  /api/v1/pipelines/:org/:repo/expand
GET   /api/v1/pipelines/:org/:repo/templates
POST  /api/v1/pipelines/:org/:repo/validate
```


davidvader:
- Endpoints will return 404 when .vela.yml does not exist for the repo/ref
- Endpoints will return 400 when .vela.yml/templates produce an invalid pipeline

Although further enhancements could be made, such as
- GET to /templates should return a lit of templates that have set a new field "link" to the actual path in the source provider to get to the template file

davidvader:
TODO: verify swagger comments, generate/overwrite swagger

davidvader:
endpoints cleaned up, swagger spec added, RFR

router/router.go
```diff
@@ -109,6 +109,16 @@ func Load(options ...gin.HandlerFunc) *gin.Engine {
 
 		// Worker endpoints
 		WorkerHandlers(baseAPI)
+
+		// Pipeline endpoints
+		pipelines := baseAPI.Group("pipelines/:org/:repo", repo.Establish())
```
wass3r, line 114: should we wrap this in a function as other handlers are here? i think we have some other pretty minimal ones wrapped.

wass3r (changes requested):

router/router.go
davidvader: yes absolutely! 👍  done!

api/pipeline.go
```diff
@@ -0,0 +1,585 @@
...
+// responses:
+//   '200':
+//     description: Successfully retrieved the pipeline
+//     type: json
+//     schema:
+//       "$ref": "#/definitions/Pipeline"
+//   '400':
+//     description: Unable to retrieve the pipeline configuration templates
+//     schema:
+//       type: string
```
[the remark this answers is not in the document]
davidvader: good point..... that could use an update across the project

api/pipeline.go
```diff
@@ -0,0 +1,585 @@
+package api
```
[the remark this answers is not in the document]
davidvader: oh good catch, fixed 👍

wass3r (approved):

davidvader:
- changed `/compile` to return yaml.Build over pipeline.Build
- removed EnvironmentSteps and EnvironmentServices due to the majority of this information being populated depends on build/runtime information, and it also exposes some environment variables that are sensitive to the system.
- `/compile` now runs Parse, ExpandSteps, SubstituteSteps, then returns the yaml.Build

api/pipeline.go
```diff
@@ -0,0 +1,647 @@
...
+)
+
+// swagger:operation GET /api/v1/pipelines/{org}/{repo} pipelines GetPipeline
+//
+// Get a pipeline configuration from the source provider
+//
+// ---
+// x-success_http_code: '200'
+// produces:
+// - application/json
```
jbrockopp, line 37: By default, this endpoint produces YAML rather than JSON.

I think we'll want to make sure this is properly reflected in the API spec 👍 

```suggestion
// produces:
// - application/x-yaml
// - application/json
```

api/pipeline.go
```diff
@@ -0,0 +1,647 @@
...
+// - in: path
+//   name: repo
+//   description: Name of the repo
+//   required: true
+//   type: string
+// - in: path
+//   name: org
+//   description: Name of the org
+//   required: true
+//   type: string
```
jbrockopp, line 48: I noticed we don't mention the query parameters here.. should we?

Since we support two query parameters (`output`, `ref`), I'm thinking it might look like:

```suggestion
// parameters:
// - in: path
//   name: repo
//   description: Name of the repo
//   required: true
//   type: string
// - in: path
//   name: org
//   description: Name of the org
//   required: true
//   type: string
// - in: query
//   name: output
//   description: format of the output returned ( json | yaml )
//   required: false
//   type: string
// - in: query
//   name: ref
//   description: source reference to fetch pipeline from
//   required: false
//   type: string
```

api/pipeline.go
```diff
@@ -0,0 +1,647 @@
...
+
+// GetPipeline represents the API handler to capture a
+// pipeline configuration for a repo from the the source provider.
+func GetPipeline(c *gin.Context) {
+	// capture middleware values
+	m := c.MustGet("metadata").(*types.Metadata)
+	r := repo.Retrieve(c)
+
+	// capture query parameters
+	output := c.DefaultQuery("output", "yaml")
```
jbrockopp, line 78: Reference the `const` we created above 👍 

```suggestion
	output := c.DefaultQuery("output", outputYAML)
```

api/pipeline.go
```diff
@@ -0,0 +1,647 @@
...
+// - in: path
+//   name: repo
+//   description: Name of the repo
+//   required: true
+//   type: string
+// - in: path
+//   name: org
+//   description: Name of the org
+//   required: true
+//   type: string
```
jbrockopp, line 146: Same thing here.. should we reference the `output` and `ref` query parameters?

api/pipeline.go
```diff
@@ -0,0 +1,647 @@
...
```
jbrockopp, line 247: Query parameters?

api/pipeline.go
```diff
@@ -0,0 +1,647 @@
...
```
jbrockopp, line 371: Query parameter?

api/pipeline.go
```diff
@@ -0,0 +1,647 @@
...
```
jbrockopp, line 500: Query parameters?

api/pipeline.go
```diff
@@ -0,0 +1,647 @@
...
+// helper function that creates a map of templates from a yaml configuration.
+func mapFromTemplates(templates []*yaml.Template) map[string]*yaml.Template {
+	m := make(map[string]*yaml.Template)
+
+	for _, tmpl := range templates {
+		m[tmpl.Name] = tmpl
+	}
+
+	return m
+}
```
jbrockopp, line 647: Do you think it's worth considering to create this map object in `types` for templates?

We have an identical function in the [go-vela/compiler](https://github.com/go-vela/compiler) repo 👍 

https://github.com/go-vela/compiler/blob/master/compiler/native/expand.go#L92-L101

jbrockopp (changes requested):

api/pipeline.go
davidvader: Yeah I think we discussed with that Id also like to add a type for Template that has a Link field for use with the UI.

davidvader: It was going to be roped into an enhancement PR for /templates but I could just do it in this PR

jbrockopp: If you plan to do it in a different PR then I think that is totally fine 👍

davidvader: I'll just do it now, rather not merge in the duplicate code and risk forgetting about it 😅

davidvader: done 👍

api/pipeline.go
davidvader: done 👍

davidvader:
need to add documentation for these endpoints

jbrockopp (approved):
LGTM

wass3r (approved):

PR #202, opened 2020-10-21
enhancement: add flag for disabling webhook validation

antoniobeyah:
# Overview
When developing locally you currently have to comment out code in order to get the server to allow you to enable/disable repos.  This pr adds a new flag for disabling webhook validation.

`VELA_DISABLE_WEBHOOK_VALIDATION` - default: false

Changes were tested locally for both cases (with and without validation enabled)

PR #204, opened 2020-10-23
docs: update info for local development

jbrockopp:
Related to https://github.com/go-vela/server/pull/202

It's been a long time since the local development docs have been updated so thought I'd give it a try.

For convenience, here are links to the old/new local development docs:

* [Old](https://github.com/go-vela/server/blob/master/.github/DOCS.md)
* [New](https://github.com/go-vela/server/blob/docs/local/DOCS.md)

Found within this PR are the following changes:

* moved the `DOCS.md` to the root of the repository so it's easier to find (previously was nested under `.github`)
* updated the `DOCS.md` content to be more focused on starting the Vela stack using `make`
* [Per feedback] updated the `DOCS.md` with information on the different compose services
* [Per feedback] added to the `DOCS.md` instructions for running builds locally when the Vela stack is started
* updated the `CONTRIBUTING.md` to reference the `DOCS.md` to reduce duplication of docs
* updated the `docker-compose.yml` to use `'` for strings (previously, we were using `'` and `"` interchangeably)
* added a `secrets.env.example` file so users can reference it as a guide to creating their own

While doing this, I was battling with myself on trying to keep the docs "lightweight" while providing valuable content.

It left me with questions that I'll provide here:

* What are people's thoughts on the new content?
* Do you think I should explain each of the services created in the compose file?
* Should we describe what the `make` commands are doing in depth?

I think once we get some of this stuff figured out, I'll plan to make the same/similar changes to [go-vela/ui](https://github.com/go-vela/ui) and [go-vela/worker](https://github.com/go-vela/worker).


JordanSussman:
> Do you think I should explain each of the services created in the compose file?

I think completing https://github.com/go-vela/community/issues/4 and/or adding some explanation of the services being utilized would be very helpful for newcomers. Some brief network/flow diagram would also likely help illustrate how everything works.

DOCS.md
```diff
@@ -0,0 +1,66 @@
...
+**NOTE: This will work for GitHub or GitHub Enterprise.**
+
+* Add OAuth client secrets to a local `secrets.env` file:
+
+```bash
+# add Github Client ID to local secrets file for `docker-compose`
+echo "VELA_SOURCE_CLIENT=<Github OAuth Client ID>" >> secrets.env
+
+# add Github Client Secret to local secrets file for `docker-compose`
+echo "VELA_SOURCE_SECRET=<Github OAuth Client Secret>" >> secrets.env
```
JordanSussman, line 45: Should we add another note about the `VELA_SOURCE_URL` variable for GHE users?

DOCS.md
```diff
@@ -0,0 +1,66 @@
...
+cd $HOME/go-vela/server
+```
+
+* Run the repository code:
+
+```bash
+# execute the `up` target with `make`
+make up
+```
+
```
JordanSussman, line 65: Any thoughts on adding notes about repo activation & faking out webhooks sent by GitHub?

DOCS.md
jbrockopp: I can add some extra information to the document on updating the `VELA_SOURCE_URL` in the compose file 👍 

Unfortunately, I don't think we want to set that up specifically in the `secrets.env` file as the `environment` block in a Docker compose file overwrites any variable set via the `.env` file.

At least, that was the behavior we experienced when testing this 🤷

DOCS.md
jbrockopp: In regards to repo activation, would providing a link to our external docs on the topic suffice?

https://go-vela.github.io/docs/usage/getting-started/enable_repo/

As far as faking out webhooks go, what content would you like to see?

I think it's going to be slightly more challenging to accurately document that process since it requires a more in-depth knowledge of how webhooks work as well as the Vela system in general.

I'd almost rather focus efforts on writing a script that creates the webhook payload for the user 🤷

JordanSussman: I agree that documenting the process is somewhat challenging due to the expected knowledge. However, I think we could at least briefly mention the following:

- Process to update postgres repos table with the proper `link` and `clone` columns
    
   ```sh
  psql -U vela -h localhost
  UPDATE repos SET link = 'https://git.company.com/<org>/<repo>', clone = 'https://git.company.com/<org>/<repo>.git' WHERE id = <id>;
   ```
- Talk through the process of obtaining a real payload from `https://git.company.com/<org>/<repo>settings/hooks/<id> ` and sending that to the with curl, insomnia, etc. to generate a build
  - Providing a script to create some of that data would also work

jbrockopp: I'm happy to say that the first bullet point **shouldn't** be necessary very soon! 🎉 

Shoutout to @wass3r for https://github.com/go-vela/server/pull/200

The important part of that PR is Vela will now be querying GitHub for the repo information when creating a repo so the `link` and `clone` fields will be populated from that call rather than the way we were doing it previously.

In regards to the second bullet point, would you also expect to see docs instructing the user that the real payload needs to be from a commit with a valid Vela pipeline (`.vela.yml`/`.vela.yaml`)?

If so, then would you also expect to see documentation on how to create/setup a valid Vela pipeline?

jbrockopp:
@JordanSussman I believe I have addressed your feedback 👍 

Please let me know your thoughts when you have a chance!

wass3r (dismissed):

JordanSussman (dismissed):

wass3r (approved):

KellyMerrick (approved):
lgtm

PR #207, opened 2020-10-29
fix: bug, set the default repo visibility to public

davidvader:
fixes a bug when adding a new repo using minimal options

bug to fix:
```bash
$ git checkout master
$ git pull
$ make up
$ vela update config --api.addr http://localhost:8080
$ vela add repo --org DavidVader --repo heyvela
FATA[0000] unable to create new repo DavidVader/heyvela: empty repo visibility provided 
```

(in fact, args to add repo are somewhat ignored. separate bug)


wass3r (commented):
Are you sure? I just tried with `0.6.0` of the CLI after adding a new repository to Github and it worked fine just using `vela add repo --org myorg --repo myrepo`

davidvader:
> Are you sure? I just tried with `0.6.0` of the CLI after adding a new repository to Github and it worked fine just using `vela add repo --org myorg --repo myrepo`

Were you using the unreleased server changes on `master`? This only started occurring after merging with the recent webhookvalidation changes

davidvader:
@wass3r so i was running something like:
```bash
$ git checkout master
$ git pull
$ make up
$ vela update config --api.addr http://localhost:8080
$ vela add repo --org DavidVader --repo heyvela
```

wass3r:
Just wanted to make sure 👍 I did consider the latest commit to check against. Thanks for the confirmation.

wass3r:
I agree, this must be a bigger issue. `GetVisibility()` should be `"public"` by default if not provided in the CLI command (see https://github.com/go-vela/cli/blob/master/action/repo_add.go#L65). I know I've recently been in this part of the code, so it's probably something I changed, though I didn't really touch the part that binds the incoming data to the struct 🤔

api/repo.go
```diff
@@ -112,9 +112,7 @@ func CreateRepo(c *gin.Context) {
 		r.SetTimeout(constants.BuildTimeoutMin)
 	}
 
-	if len(input.GetVisibility()) == 0 {
-		r.SetVisibility(constants.VisibilityPublic)
-	}
+	r.SetVisibility(constants.VisibilityPublic)
```
jbrockopp, line 115: I think we're close with this change but I think I'd like to see it modified slightly 👍 

The behavior you're experiencing is due to how the code is written.

Previously, we told the API set the `repo.Visibility` field to `public` **if one was not** provided from the `input`.

The problem is we don't do anything with the `repo.Visibility` field **if one was** provided from the `input`.

Instead, I believe the code should look like:

```suggestion
	if len(input.GetVisibility()) == 0 {
		r.SetVisibility(constants.VisibilityPublic)
	} else {
		r.SetVisibility(input.GetVisibility())
	}
```

jbrockopp (changes requested):

api/repo.go
wass3r: 🤦 looks like i goofed that in a recent change. my bad. good fine @jbrockopp 👍

davidvader: yeah youre right, good call, I mentioned that in the PR description and saw it was in updaterepo

davidvader:
changes applied 👍

jbrockopp (approved):
LGTM

wass3r (approved):

PR #209, opened 2020-11-04
feat: add API endpoint to login using PAT

justsimplify:
Ref: https://github.com/go-vela/community/issues/126
Refactored the logic to fetch the token and updating the database to a separate function `performUserOperation`.

Will add tests in next commit.


wass3r:
thanks @justsimplify ! I think this will work. We'll just hold off a bit as we're working out some pending changes. there might be some tweaks to this after those changes go in. we'll reach out again when that happens! thanks for the patience.

justsimplify:
Because of the conflicts, raising a different [PR](https://github.com/go-vela/server/pull/246). 
This way, less confusion.

PR #210, opened 2020-11-04
fix(webhook): set build created

davidvader:
fixes https://github.com/go-vela/community/issues/144

### DIFF
--- a/api/webhook.go
+++ b/api/webhook.go
@@ -388,6 +388,9 @@ func PostWebhook(c *gin.Context) {
 			)
 		}
 
+		// set build created timestamp to now
+		b.SetCreated(time.Now().UTC().Unix())
+
 		// parse and compile the pipeline configuration file
 		p, err = compiler.FromContext(c).
 			WithBuild(b).


davidvader:
Created is not set here
https://github.com/go-vela/server/blob/master/api/webhook.go#L125
and is never updated

restarting build works because of this line
https://github.com/go-vela/server/blob/master/api/build.go#L578

api/webhook.go
```diff
@@ -388,6 +388,9 @@ func PostWebhook(c *gin.Context) {
 			)
 		}
 
+		// set build created timestamp to now
+		b.SetCreated(time.Now().UTC().Unix())
```
jbrockopp, line 392: I believe we already set this in the [`planBuild()` function](https://github.com/go-vela/server/blob/master/api/build.go#L915-L953).

The direct lines that achieve this functionality are [here](https://github.com/go-vela/server/blob/master/api/build.go#L919-L920):

```go
// planBuild is a helper function to plan the build for
// execution. This creates all resources, like steps
// and services, for the build in the configured backend.
func planBuild(database database.Service, p *pipeline.Build, b *library.Build, r *library.Repo) error {
	// update fields in build object
	b.SetCreated(time.Now().UTC().Unix())

	// send API call to create the build
	err := database.CreateBuild(b)
	if err != nil {
		// clean up the objects from the pipeline in the database
		cleanBuild(database, b, nil, nil)

		return fmt.Errorf("unable to create new build for %s: %v", r.GetFullName(), err)
	}

	// send API call to capture the created build
	b, _ = database.GetBuild(int(b.GetNumber()), r)

	// plan all services for the build
	services, err := planServices(database, p, b)
	if err != nil {
		// clean up the objects from the pipeline in the database
		cleanBuild(database, b, services, nil)

		return err
	}

	// plan all steps for the build
	steps, err := planSteps(database, p, b)
	if err != nil {
		// clean up the objects from the pipeline in the database
		cleanBuild(database, b, services, steps)

		return err
	}

	return nil
}
```

jbrockopp (changes requested):

davidvader:
Converted to draft, will implement via compiler/executor 🦖

PR #213, opened 2020-11-10
fix(open-api): spec parameter order correction

slushpupie:
Fix swagger spec parameters
    
This fixes a few things with swagger spec parameters:
 * Consistently put the path arguments in the same order found in the path
 * Put body arguments after path arguments
 * Correct the description of the Service arguments to indicate its an ID, not a name
 * Fix the Workers API to use the "security" parameter, rather than header parameters
    
This is important to code-generation, as many of the created functions will have positional arguments, and we need them to match the API usage.

### DIFF
--- a/api-spec/vela-server.json
+++ b/api-spec/vela-server.json
@@ -666,15 +666,15 @@
         "parameters": [
           {
             "type": "string",
-            "description": "Name of the repo",
-            "name": "repo",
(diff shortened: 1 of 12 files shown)


jbrockopp (approved):
LGTM

wass3r (approved):

PR #216, opened 2020-11-18
fix: add encoding component

kneal:
This fix is to improve the experience when an admin needs to modify secrets. Sending encryted data over HTTP when it's not encoding is unsafe because it does not follow the UTF-8 standard. 

When the data does get sent some contents within the value field to not get marshaled properly which means the secret becomes invalid. This fix adds a simple base64 encode/decode process to ensure the character set is compatible with all interactions of the secrets. 

fixes: https://github.com/go-vela/community/issues/100

### DIFF
--- a/secret/native/crypto.go
+++ b/secret/native/crypto.go
@@ -8,6 +8,7 @@ import (
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/rand"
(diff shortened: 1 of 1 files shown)


jbrockopp (approved):
LGTM

wass3r (approved):

PR #217, opened 2020-12-02
enhance(deployment): add payload metadata from github

JordanSussman:
This utilizes the `payload` field passed by GitHub to inject arbitrary metadata into builds as `DEPLOYMENT_PARAMETER_<key>` environment variables. 

Notes:

- All existing postgres and sqlite instances will need to be manually updated to add the `deploy_payload` column within the build table until we implement something to alter the table automagically.

xref: https://github.com/go-vela/community/issues/89


go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/go-vela/server
 
 go 1.13
 
+replace github.com/go-vela/types => github.com/JordanSussman/types v0.1.2-0.20201202154222-3231298c44c8
```
JordanSussman, line 5: Will be removed once https://github.com/go-vela/types/pull/117 has been merged.

source/github/webhook.go
```diff
@@ -259,6 +261,22 @@ func processDeploymentEvent(h *library.Hook, payload *github.DeploymentEvent) (*
 	b.SetBranch(payload.GetDeployment().GetRef())
 	b.SetRef(payload.GetDeployment().GetRef())
 
+	// check if payload is provided within request
+	if payload.GetDeployment().Payload != nil {
+		deployPayload := raw.StringSliceMap{}
+		// unmarshal the payload into the expected map[string]string format
+		err := json.Unmarshal(payload.GetDeployment().Payload, &deployPayload)
+		if err != nil {
+			return nil, err
```
JordanSussman, line 270: Thoughts on if we should skip the payload data if it doesn't match `map[string]string` instead of returning some error?

database/ddl/postgres/build.go
```diff
@@ -11,37 +11,44 @@ const (
...
+	parent         INTEGER,
+	event          VARCHAR(250),
+	status         VARCHAR(250),
+	error          VARCHAR(500),
+	enqueued       INTEGER,
+	created        INTEGER,
+	started        INTEGER,
+	finished       INTEGER,
+	deploy         VARCHAR(500),
+	deploy_payload VARCHAR(2000),
```
JordanSussman, line 26: Should we declare a higher limit for this column?

source/github/webhook.go
davidvader: Can a deployment contain an empty payload?
Is it possible the deploy payload value could just be a string or a nil value that isnt {}? In the event of an error, etc?
If so, I think skipping when it doesnt match on map[string]string is fine, but if deployment is expected to _always_ have an error I like checking for unmarshal errors.

database/ddl/postgres/build.go
davidvader: Do you have any sample data to compare this value to? 
Or for example, what you might expect to add to go-vela/mock for the build response?
https://github.com/go-vela/mock/blob/master/server/build.go

database/ddl/ddl_test.go
```diff
@@ -18,37 +18,37 @@ func TestDDL_NewMap_Postgres(t *testing.T) {
 	// setup types
 	want := &Map{
 		BuildService: &Service{
-			Create: postgres.CreateBuildTable,
+			Create: []string{postgres.CreateBuildTable, postgres.CreatePayloadColumn},
```
davidvader, line 21: This is the first time we're introducing Alter sets.
Have we considered separating Create queries and Alter queries? In that way we dont have to reuse Create for both creation and changeset, which might potentially create confusion?

(this does invoke a conversation on how we should do database changesets, which are not currently implemented anywhere....)

davidvader: Or... renaming Create to.... Setup? Not sure yet. Perhaps meant for a separate refactor...

davidvader:
I motion that we utilize the single Create query to add the DeployPayload column, but have discussion for refactoring how we do database updates/versioning, which might be more appropriate in a separate PR?

database/ddl/ddl_test.go
JordanSussman: I debated on creating a separate section within the `Service` to handle alter sets but thought the ability to specify order within the single slice over weighed separating into multiple slices.

source/github/webhook.go
JordanSussman: I inquired with GitHub support about the payload and they responded with: “If the payload is passed in the input when creating a deployment, this endpoint expects it to either be a JSON object or a string. Otherwise, a validation error will be thrown.“. I'll play around with creating some deployments directly with the GitHub API and add relevant tests to handle those scenarios.

database/ddl/postgres/build.go
JordanSussman: This can technically contain any valid JSON or string if the deployment was created directly via the GitHub API. One example of expected input can be found [here](https://github.com/go-vela/server/pull/217/files#diff-e152bd7186c7835e4de082e909c8df1bb0550503a5bbdcf58e1444034ed69c91R10-R13).

JordanSussman:
> I motion that we utilize the single Create query to add the DeployPayload column, but have discussion for refactoring how we do database updates/versioning, which might be more appropriate in a separate PR?

Mixed opinions with introducing changes that break existing database installs unless the admin runs manual commands against the postgres db. Do you propose we create some immediate follow-up PR to handle this?

davidvader:
> Mixed opinions with introducing changes that break existing database installs unless the admin runs manual commands against the postgres db. Do you propose we create some immediate follow-up PR to handle this?

Oh I was a bit unclear. I meant this PR is doing 2 things, updating our change-strategy _and_ adding a column to a table. My suggestion was to split them up. 
So have a separate effort/refactor that implements database versioning using alter/changesets with a rollback strategy, etc. That way we don't have to try to fit it into our current implementation that isn't really set up for it.

But sadly during release we will need to manually execute the changeset if we split them and merge this in without change-strategy updates.... which is annoying I agree.

JordanSussman:
> Oh I was a bit unclear. I meant this PR is doing 2 things, updating our change-strategy and adding a column to a table. My suggestion was to split them up.
So have a separate effort/refactor that implements database versioning using alter/changesets with a rollback strategy, etc. That way we don't have to try to fit it into our current implementation that isn't really set up for it.

I'm fine with splitting the logic into a separate PR (assuming it is merged before 0.7.0-rc1). My upcoming changes within https://github.com/go-vela/community/issues/164 are going to introduce the same problem of needing to alter the already existing table, so coming up with a strategy to handle this scenario is likely worthwhile. How about we start a new thread in [slack](https://gophers.slack.com/archives/CNRRKE8KY) to gather other input about the alter strategy?

JordanSussman:
@davidvader - ready for another review.

davidvader (approved):
excellent, thanks for making those updates, LGTM 👍

wass3r (approved):

PR #218, opened 2020-12-07
feat(repo settings): add build timeout default 30 minutes

KellyMerrick:
* create repo default timeout to use new constant
* dependent on types pr: https://github.com/go-vela/types/pull/120
* supports https://git.target.com/vela/feature-requests/issues/86


matt-fevold (dismissed):
LGTM - pending merge of that dependent PR

api/repo.go
```diff
@@ -115,7 +115,7 @@ func CreateRepo(c *gin.Context) {
 	// set the timeout field based off the input provided
 	if input.GetTimeout() == 0 {
 		// default build timeout to 30m
-		r.SetTimeout(constants.BuildTimeoutMin)
+		r.SetTimeout(constants.BuildTimeoutDefault)
```
JordanSussman, line 118: Any objections to updating this to also allow the server to override the default value? Example: purpose the Vela administrator wants to set the default build timeout to 90 minutes for all builds. Adding another `cli.DurationFlag` within the [main.go](https://github.com/go-vela/server/blob/f86014e062baf2301cffba9917d20bb4ba89742a/cmd/vela-server/main.go) to allow this to be overwritten would be my preference.

davidvader: I agree with Jordan, I think adding this as a flag, similar to the CLI's default for setting repo add/update timeout default, would create the most flexibility.

Like this
https://github.com/go-vela/cli/blob/master/action/repo_add.go#L67-L73

But Jordan is right, `cli.DurationFlag` is probably more appropriate than `cli.Int64Flag` which is used in the CLI.

So, a flag that sets a default, that if not provided, uses the new types default value that youve implemented 👍

KellyMerrick: Created https://github.com/go-vela/community/issues/179 story for this. We'll merge this PR in to make our next release candidate.

wass3r (approved):

jbrockopp (approved):
LGTM

PR #219, opened 2020-12-11
enhance(workers): track build_limit

JordanSussman:
Updates the `/metrics` endpoint to return three new metrics:

```
vela_totals{field="sum",resource="worker",value="build_limit"}
vela_totals{field="count",resource="worker",value="active"}
vela_totals{field="count",resource="worker",value="inactive"}
```

Notes:

- The values of those metrics only contain workers that have checked in within the last 60 seconds.
- We could opt to make the worker metrics optional by adding some URL query parameter. e.g. `/metrics?workers=true`.
- This functionality will further help enable creating some external service to auto scale workers based on pending builds vs build limit.
- Existing postgres and sqlite instances will need to be manually updated to include the `build_limit` column on the `workers` table. e.g. `ALTER TABLE workers ADD COLUMN build_limit INTEGER;`.

xref: go-vela/community#164


api/metrics.go
```diff
@@ -163,6 +171,23 @@ func recordGauges(c *gin.Context) {
 	totals.WithLabelValues("build", "status", "success").Set(float64(bSucc))
 	totals.WithLabelValues("build", "status", "error").Set(float64(bErr))
 
+	// Add worker metrics
+	var buildLimit int64
+	var workerCount int64
+	// get the unix time from 1 minute ago
+	before := time.Now().UTC().Add(-1 * time.Minute).Unix()
```
matt-fevold, line 178: Still new to the project, how often do workers check in?

JordanSussman: Good question - the current default check in interval is [15 minutes](https://github.com/go-vela/worker/blob/7e41a9d30647af09eab67e97c6ada259e4f0b3f2/cmd/vela-worker/flags.go#L33). Few thoughts to handle the difference:

- Update this code to utilize a configurable variable within the [main.go](https://github.com/go-vela/server/blob/f86014e062baf2301cffba9917d20bb4ba89742a/cmd/vela-server/main.go) to allow administrators to refine this value.
- Update the default worker check in interval to be shorter.
- Do nothing: assume the Vela administrator overrides the `VELA_CHECK_IN` variable to be sub 1 minute if they want values.

jbrockopp: @matt-fevold @JordanSussman 

I think it's worth exploring the idea of decreasing the worker check in interval from `15` minutes.

Not entirely sure what's a good threshold, but I could potentially see it be `5` - `10` minutes 👍 

I also think that we should consider increasing the threshold for the last time a worker checked into a server.

1 minute seems a bit steep but I do see value in having near real-time information on this.

Maybe making it configurable like you said would be a good middle ground?

I almost wonder if we should explore creating separate queries for counting the active vs inactive workers too?

JordanSussman: > Not entirely sure what's a good threshold, but I could potentially see it be 5 - 10 minutes 

I'll likely override the default to around 1 minute since I want near real-time information for triggering autoscaling events. Regardless, I'll update the `main.go` to have a 5 minute default interval.

> I almost wonder if we should explore creating separate DB queries for counting the active vs inactive workers too?

I'm not sure this would provide much value unless we allow the administrator to control which metric(s) are returned by the `/metrics` endpoint. Are you thinking of other scenarios where you want to query those values independently of each other?

jbrockopp: > I'll likely override the default to around 1 minute since I want near real-time information for triggering autoscaling events. Regardless, I'll update the main.go to have a 5 minute default interval.

Yeah I think that's fair 👍 

And maybe as we continue to grow our own use-cases for some of this information we might want to have a `1` minute interval like you 🤷 

> I'm not sure this would provide much value unless we allow the administrator to control which metric(s) are returned by the `/metrics` endpoint. Are you thinking of other scenarios where you want to query those values independently of each other?

I agree that adding these separate queries for the solely the `/metrics` endpoint is a bit overkill.

So I'd say feel free to leave it out and we can plan to implement them later 👍 

But to answer your question, yes I was thinking in the future the idea/ability for platform admins to "purge"/"clean" (w/e you want to call the endpoint) the workers from that table.

To elaborate, it would be cool for us to offer [a new endpoint to the workers API](https://github.com/go-vela/server/blob/master/router/worker.go) that might look like:

```
// POST   /api/v1/workers/purge
```

I think we'd restrict this endpoint to only platform admins being able to use it. Then I think the endpoint would execute [a new database query](https://github.com/go-vela/server/blob/master/database/dml/postgres/worker.go) that would look for only inactive workers:

```sql
SELECT *
FROM workers
WHERE active = f;
```

We could then iterate through the list of returned workers and delete them one at a time 👍 

But again, think that's likely out of scope for this change so think we can ignore adding any such queries at this time.

JordanSussman: My eventual plans of adding some `/api/v1/pause` endpoint to the worker to allow the platform admins to remove workers from plucking jobs from the queue will likely complicate some of this logic. However, I agree that this current implementation shouldn't attempt to handle some of the future optimizations/changes.

JordanSussman: @jbrockopp - the latest commits should address the feedback you provided.

jbrockopp (approved):
LGTM

kneal (approved):
LGTM 🐬

PR #235, opened 2021-01-09
 feat: add endpoint for canceling a running build

meinenec:
Adds a new endpoint `DELETE /api/v1/repos/:org/:repo/builds/:build/cancel` which will cancel a running build.

This endpoint does not currently add the ability to remove abandoned builds or remove a build from the queue. I think both of those should come, but want to get the ability to cancel a build enabled sooner than later.


router/middleware/executors/executor_test.go
```diff
@@ -0,0 +1,34 @@
...
+	context, _ := gin.CreateTestContext(nil)
+	ToContext(context, want)
+
+	// run test
+	got := Retrieve(context)
+
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("Retrieve is %v, want %v", got, want)
+	}
+}
```
meinenec, line 34: I did not include a test for `TestExecutors_Establish` since that requires calling a worker to get the executors. I believe this could be mocked. Does anyone have any thoughts or suggestions?

router/middleware/executors/context.go
```diff
@@ -0,0 +1,42 @@
...
+	if value == nil {
+		return nil
+	}
+
+	e, ok := value.([]library.Executor)
+	if !ok {
+		return nil
+	}
+
+	logrus.Infof("FromContext: %v", e)
```
jbrockopp, line 33: Is this intended to be here?

router/middleware/executors/context.go
```diff
@@ -0,0 +1,42 @@
...
+	return e
+}
+
+// ToContext adds the executor to this context if it supports
+// the Setter interface.
+func ToContext(c Setter, e []library.Executor) {
+	logrus.Infof("ToContext: %v", e)
```
jbrockopp, line 40: Is this intended to be here?

jbrockopp (approved):
LGTM

wass3r (approved):

router/build.go
```diff
@@ -54,6 +56,7 @@ func BuildHandlers(base *gin.RouterGroup) {
 			build.GET("", perm.MustRead(), api.GetBuild)
 			build.PUT("", perm.MustWrite(), middleware.Payload(), api.UpdateBuild)
 			build.DELETE("", perm.MustPlatformAdmin(), api.DeleteBuild)
+			build.DELETE("/cancel", executors.Establish(), perm.MustAdmin(), api.CancelBuild)
```
JordanSussman, line 59: Thoughts on updating this to `perm.MustWrite(` instead of `perm.MustAdmin()`?

meinenec: To start I figured admins was enough. Do you believe it should be MustWrite instead?

I'll merge this for now, but that's something we could come back to.

PR #246, opened 2021-01-18
feat: add login using PAT

justsimplify:
Ref: https://github.com/go-vela/community/issues/126


wass3r (approved):
looks like our linter doesn't handle PR originating from forks well. i think here are couple of things to address, but we can tweak after merge. thanks for the contribution!

jbrockopp (approved):

PR #236, opened 2021-01-11
chore: update server address flag to behave like worker

meinenec:
Updates the server flags to use net/url to store the address similar to the [worker](https://github.com/go-vela/worker/pull/130)

PR #247, opened 2021-01-19
fix(ingress): revert #236

JordanSussman:
Reverting the changes within #236 since the changes cause invalid redirects if the server is setup with a load balancer handling TLS termination.  More context:

Purpose the Vela Administrator would like the server to listen on port 8080 (HTTP) but redirect clients to port 443 (HTTPS) to be served by the load balancer. Utilizing the `VELA_ADDR` value of `https://vela.company.com:8080` will cause the server to listen on 8080 but redirect clients to `https://vela.company.com:8080` instead of `https://vela.company.com`. 

I don't see any simple code changes to accomplish this behavior besides reverting the change altogether.

### DIFF
--- a/cmd/vela-server/main.go
+++ b/cmd/vela-server/main.go
@@ -21,6 +21,12 @@ func main() {
 
 	app.Flags = []cli.Flag{
 		&cli.StringFlag{
+			EnvVars: []string{"VELA_PORT"},
+			Name:    "server-port",
+			Usage:   "API port to listen on",
+			Value:   ":8080",
+		},
+		&cli.StringFlag{
 			EnvVars: []string{"VELA_LOG_LEVEL", "LOG_LEVEL"},
 			Name:    "log-level",
 			Usage:   "set log level - options: (trace|debug|info|warn|error|fatal|panic)",
--- a/cmd/vela-server/server.go
+++ b/cmd/vela-server/server.go
@@ -6,9 +6,7 @@ package main
 
 import (
 	"context"
-	"fmt"
 	"net/http"
-	"net/url"
 	"time"
 
 	"github.com/go-vela/server/router"
@@ -99,25 +97,11 @@ func server(c *cli.Context) error {
 		middleware.Worker(c.Duration("worker-active-interval")),
 	)
 
-	addr, err := url.Parse(c.String("server-addr"))
-	if err != nil {
-		return err
-	}
-
 	var tomb tomb.Tomb
 	// start http server
 	tomb.Go(func() error {
-		port := addr.Port()
+		srv := &http.Server{Addr: c.String("server-port"), Handler: router}
 
-		// check if a port is part of the address
-		if len(port) == 0 {
-			port = "8080"
-		}
-
-		// gin expects the address to be ":<port>" ie ":8080"
-		srv := &http.Server{Addr: fmt.Sprintf(":%s", port), Handler: router}
-
-		logrus.Infof("running server on %s", addr.Host)
 		go func() {
 			logrus.Info("Starting HTTP server...")
 			err := srv.ListenAndServe()
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,6 +30,7 @@ services:
       VELA_QUEUE_DRIVER: redis
       VELA_QUEUE_CONFIG: 'redis://redis:6379'
       VELA_QUEUE_WORKER_ROUTES: 'docker,local,docker:local'
+      VELA_PORT: ':8080'
       VELA_SECRET: 'zB7mrKDTZqNeNTD8z47yG4DHywspAh'
       VELA_SOURCE_DRIVER: github
       VELA_SECRET_NATIVE_KEY: 'C639A572E14D5075C526FDDD43E4ECF6'


jbrockopp (changes requested):
It's pretty early in the morning so apologies if I'm not following what you are saying correctly.

AFAICT, the concern expressed is that if you update the `VELA_ADDR` to a value of `https://vela.company.com:8080` then you get invalid redirects.

Have you tried dropping the `:8080` prefix from your `VELA_ADDR` when using [v0.7.0-rc2](https://github.com/go-vela/server/releases/tag/v0.7.0-rc2)?

i.e. `VELA_ADDR=https://vela.company.com`

I ask, because I believe providing `:8080` in the `VELA_ADDR` is redundant.

The server is already setup to listen on port `8080` by default, without any extra configuration:

https://github.com/go-vela/server/blob/60a7ed0d1001eeccb756f585d752b7b0ec90d9cf/cmd/vela-server/server.go#L112-L115

https://github.com/go-vela/server/blob/60a7ed0d1001eeccb756f585d752b7b0ec90d9cf/Dockerfile#L13

JordanSussman:
You appear to be correct that utilizing the `VELA_ADDR` of `https://vela.company.com` when utilizing the changes introduced within https://github.com/go-vela/server/pull/243 allows this to successfully work. However, I'd argue that this change in behavior is a poor administrator experience for those that may wish to utilize HTTPS with the server listening on some port besides 8080.

jbrockopp:
> this change in behavior is a poor administrator experience for those that may wish to utilize HTTPS with the server listening on some port besides 8080.

I was thinking about that and was trying to determine what the reason would be for changing it beyond `8080`?

After all, we're talking about the ability to configure what port inside a Docker container the server will respond to.

Even our Dockerfile is hardcoded for exposing port 8080 so I think the scenarios for this use-case are slim to none.

The only one I could come up with is if you are running the container with host networking and trying to prevent a port conflict.

All that being said, I don't disagree that the experience here could certainly be improved 👍 

However, I'm unsure if a full revert of the change is the best path forward based off the ability to drop the `:8080` prefix?

I'm sure there are more options, but as an example of one that won't require a full revert, we could make this line configurable:

https://github.com/go-vela/server/blob/0e44936ba59a7a55b7607f907f9f06bb6f166f9b/cmd/vela-server/server.go#L114

JordanSussman:
> Even our Dockerfile is hardcoded for exposing port 8080 so I think the scenarios for this use-case are slim to none.

The `EXPOSE` instructions means little beyond documentation:

> The EXPOSE instruction informs Docker that the container listens on the specified network ports at runtime. You can specify whether the port listens on TCP or UDP, and the default is TCP if the protocol is not specified.

> The EXPOSE instruction does not actually publish the port. It functions as a type of documentation between the person who builds the image and the person who runs the container, about which ports are intended to be published. To actually publish the port when running the container, use the -p flag on docker run to publish and map one or more ports, or the -P flag to publish all exposed ports and map them to high-order ports.

source: https://docs.docker.com/engine/reference/builder/#expose

Suppose the administrator could utilize the syntax of `docker run -p 8200:8080 ...` if they wanted to make the server available on the non default port. 

> I'm sure there are more options, but as an example of one that won't require a full revert, we could make this line configurable:

IMO that outcome is basically the same as reverting the change besides the little bit of extra validation that [url.Parse](https://github.com/go-vela/server/pull/236/files#diff-f52da6a4762e595f71b4675b66693f12eeaa61df500f8899ed35ca338aef4800R102) offers.

jbrockopp:
> > Even our Dockerfile is hardcoded for exposing port 8080 so I think the scenarios for this use-case are slim to none.
>
> The EXPOSE instructions means little beyond documentation:

I know what `EXPOSE` does but I disagree with your statement.. not with the functionality of `EXPOSE` but with it meaning little.

Referring to those docs, using `EXPOSE` is our way of saying we expect that you should be running the server on port 8080.

I think we both can attest to how important following documentation is when using any technical product 😅 

But this is off topic and I think the more important part of my original statement was missed.

Regardless of what `EXPOSE` is, can we come up with a use-case that would require not running the server on 8080?

> Suppose the administrator could utilize the syntax of `docker run -p 8200:8080 ...` if they wanted to make the server available on the non default port.

Exactly 👍 

This adds to my point on being unable to come up with a valid use case for not running on the server on 8080.

> IMO that outcome is basically the same as reverting the change besides the little bit of extra validation that [url.Parse](https://github.com/go-vela/server/pull/236/files#diff-f52da6a4762e595f71b4675b66693f12eeaa61df500f8899ed35ca338aef4800R102) offers.

While the final outcome may be the same, the approach to those outcomes are different.

Personally, I'd prefer using `url.Parse` vs not using it 🤷 

FWIW:

This also gives us the benefit of being more consistent with the worker and arguably better looking into the future.

jbrockopp:
@JordanSussman I've opened https://github.com/go-vela/server/pull/258 in an attempt to address the concerns in this PR.

Does this satisfy what you were thinking in regards to being able to customize the API port for the server?

NOTE: This means we'll still use `url.Parse()` under the hood.

JordanSussman:
> Does this satisfy what you were thinking in regards to being able to customize the API port for the server?

The changes within #258 do indeed address my feedback 👍.

PR #256, opened 2021-01-21
fix(webhook): duplicate compiler before processing

wass3r:
should fix https://github.com/go-vela/community/issues/181

This fix will create a clone of the `Engine` with basic settings before compiling the incoming request. This is to avoid situation where the shared compiler attached at application start up is "corrupted" with other build data.

Tested by emitting multiple concurrent requests and logging on the worker where the situation was observed prior to the fix and not after.

Testing surfaced another issue where we attach an error if the first attempt to create a build fails. When the subsequent attempt succeeds the error is not cleared.

Note: an arguably nicer solution would be to handle this via mutex. i did spend some time on that, but i think it might require some additional changes.

### DIFF
--- a/api/webhook.go
+++ b/api/webhook.go
@@ -390,6 +390,7 @@ func PostWebhook(c *gin.Context) {
 
 		// parse and compile the pipeline configuration file
 		p, err = compiler.FromContext(c).
+			Duplicate().
 			WithBuild(b).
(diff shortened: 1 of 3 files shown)


matt-fevold (approved):

jbrockopp (approved):
LGTM

PR #277, opened 2021-02-04
fix(cancel): return build struct from the worker

meinenec:
return the struct that the worker responds with
https://github.com/go-vela/worker/pull/152

### DIFF
--- a/api/build.go
+++ b/api/build.go
@@ -5,6 +5,7 @@
 package api
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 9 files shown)


api/build.go
```diff
@@ -1150,7 +1151,10 @@ func CancelBuild(c *gin.Context) {
 				util.HandleError(c, http.StatusBadRequest, retErr)
 				return
 			}
-			c.JSON(resp.StatusCode, strings.Trim(string(respBody), "\""))
+
+			json.Unmarshal(respBody, b)
```
davidvader, line 1155: can we check the err from Unmarshal here?
```suggestion
			err = json.Unmarshal(respBody, b)
			if err != nil {
			
		            util.HandleError(c, http.StatusInternalServerError, err)

		            return
	                }
			
```

meinenec: TYVM 😄

davidvader (approved):

jbrockopp (approved):
LGTM

PR #289, opened 2021-02-16
docs: fix/update api spec

wass3r:
depends on: https://github.com/go-vela/types/pull/152

changes include:
- add paging information to relevant endpoints
- add other missing parameters
- utilize error response defintion
- correct errors (response code, spelling, words, etc)
- bump version
- remove base api path, we also define them per route, so in some apps like insomnia they double up, plus base path applies to all endpoints and there didn't seem to be a way to exclude those that are not at base path


matt-fevold (dismissed):

jbrockopp (dismissed):
LGTM

jbrockopp (approved):
LGTM

matt-fevold (approved):

PR #291, opened 2021-02-18
feat(database): compress logs to reduce storage

jbrockopp:
Dependent on https://github.com/go-vela/types/pull/154

Before this change, we ran some builds locally and checked the size of the `logs` table in the database:

```
vela=# SELECT pg_size_pretty (pg_relation_size('logs'));
 pg_size_pretty
----------------
 24 kB
(1 row)
```

After this change, when we run those same builds and check the size of the `logs` table in the database:

```
vela=# SELECT pg_size_pretty (pg_relation_size('logs'));
 pg_size_pretty
----------------
 8192 bytes
(1 row)
```

Doing a rough calculation from the above data, we were able to shrink the size of the `logs` table to `1/3` it's original size.

i.e. `8` kB  /  `24` kB = `1/3`

> NOTE: This change **is backward compatible**. If you try fetching an uncompressed log, everything will work as expected.
>
> We'll plan to offer some sort of utility that will assist in compressing all existing logs in the database.

### DIFF
--- a/database/log.go
+++ b/database/log.go
@@ -5,10 +5,7 @@
 package database
 
 import (
-	"bytes"
-	"compress/zlib"
 	"fmt"
-	"io/ioutil"
 
 	"github.com/go-vela/types/constants"
 	"github.com/go-vela/types/database"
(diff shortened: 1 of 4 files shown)


KellyMerrick (approved):
lgtm

wass3r (approved):

PR #297, opened 2021-02-24
enhance(database): support setting different compression level

jbrockopp:
Dependent on https://github.com/go-vela/types/pull/156

This enables us to set the compression level for the `go-vela/types/database.Log.Compress()` function.

This can be set via the flag `database.compression.level` or environment variables:

* `VELA_DATABASE_COMPRESSION_LEVEL`
* `DATABASE_COMPRESSION_LEVEL`

The default compression level is set to `3` which is faster than some levels but has a cost of larger data in the database.

This also has an added benefit of being able to "turn off" log compression by passing a compression level of `0`.

Also added validation logic to ensure the compression level set is between the supported range of `-1` - `9`

### DIFF
--- a/cmd/vela-server/main.go
+++ b/cmd/vela-server/main.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/go-vela/server/version"
+	"github.com/go-vela/types/constants"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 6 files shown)


KellyMerrick (approved):
lgtm

matt-fevold (approved):

wass3r (approved):

PR #305, opened 2021-03-03
fix: set user ID when repairing a repo

davidvader:
### description
set userid when repairing a repo, keep git webhook owner and vela repo owner in sync

(discussion) is there a scenario where we would want this to stay out of sync?

### problem
fixes 403 permissions errors that are hard to track down because webhook owner and repo owner can become out of sync.

### scenario
USER_A is the owner of a Vela webhook, USER_A leaves the git organization. Then, USER_B receives `403 account was suspended` runs Repair Repo, owner of the webhook in git is now USER_B but the owner in Vela is still USER_A, where they continue to receive 403 errors.

### DIFF
--- a/api/repo.go
+++ b/api/repo.go
@@ -706,16 +706,19 @@ func RepairRepo(c *gin.Context) {
 	// if the repo was previously inactive, mark it as active
 	if !r.GetActive() {
 		r.SetActive(true)
+	}
 
-		// send API call to update the repo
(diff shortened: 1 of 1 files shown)


matt-fevold:
we discussed this at a code review session, we talked about combining `chown repo` and `repair repo` into one action. @davidvader to follow up with more of a proposal. From a user's perspective there's not much difference and require admin access for the repo.

PR #320, opened 2021-03-18
feat(database): encrypt user fields for enhanced security

jbrockopp:
Dependent on https://github.com/go-vela/types/pull/164

This change will introduce encryption for user fields that are stored in the Vela database.

We've decided to use the Advanced Encryption Standard (`AES`) with a `256` bit key (a.k.a. `AES-256`):

https://en.wikipedia.org/wiki/Advanced_Encryption_Standard

> NOTE: This change is backward compatible. If you try fetching an unencrypted user, everything will work as expected.
>
> We'll plan to offer some sort of utility that will assist in encrypting all existing users in the database.


KellyMerrick (approved):
lgtm, expecting my review may be removed when the branch is updated since currently out of date

matt-fevold (approved):

PR #321, opened 2021-03-18
feat(database): encrypt repo fields for enhanced security

jbrockopp:
Dependent on https://github.com/go-vela/types/pull/163

This change will introduce encryption for repo fields that are stored in the Vela database.

We've decided to use the Advanced Encryption Standard (`AES`) with a `256` bit key (a.k.a. `AES-256`):

https://en.wikipedia.org/wiki/Advanced_Encryption_Standard

> NOTE: This change is backward compatible. If you try fetching an unencrypted repo, everything will work as expected.
>
> We'll plan to offer some sort of utility that will assist in encrypting all existing repos in the database.


database/repo.go
```diff
@@ -161,6 +223,14 @@ func (c *client) CreateRepo(r *library.Repo) error {
 		return err
 	}
 
+	// decrypt the fields for the repo
```
davidvader, line 226:
```suggestion
	// encrypt the fields for the repo
```

database/repo.go
```diff
@@ -180,6 +250,14 @@ func (c *client) UpdateRepo(r *library.Repo) error {
...
```
davidvader, line 253:
```suggestion
	// encrypt the fields for the repo
```

davidvader (changes requested):
small typos

database/repo.go
jbrockopp: fixed 👍

database/repo.go
jbrockopp: fixed 👍

matt-fevold (approved):

davidvader (approved):

PR #358, opened 2021-05-04
feat(postgres): add vela resource dml

jbrockopp:
Part of the effort for https://github.com/go-vela/community/issues/248

Related to https://github.com/go-vela/server/pull/341

This adds the Postgres [data manipulation language (DML)](https://en.wikipedia.org/wiki/Data_manipulation_language) for Vela.

All of the DML is contained in the `github.com/go-vela/server/database/postgres/dml` package.

The DML was copied from [the existing `../../server/database/dml/postgres`](https://github.com/go-vela/server/tree/master/database/dml/postgres) old package structure.

This is an ongoing effort to make the `../../server/database` package conform to standards set in other packages.

PR #392, opened 2021-05-19
fix(postgres): DML query placeholders for parameters

jbrockopp:
Part of the effort for https://github.com/go-vela/community/issues/248

Related to https://github.com/go-vela/server/pull/341 and https://github.com/go-vela/server/pull/358

Dependent on https://github.com/go-vela/server/pull/378 and https://github.com/go-vela/types/pull/174

**This is a workaround to a problem found in an upstream library.**

This updates the `go-vela/server/database/postgres/dml` to use `?` instead of `$` for parameter placeholders in queries.

The official Postgres syntax for query placeholders references using `$` for placeholders:

https://www.postgresql.org/docs/current/xfunc-sql.html#XFUNC-SQL-FUNCTION-ARGUMENTS

However, it appears there is a bug in the `gorm.io/gorm` codebase that prevent this from working as intended:

https://github.com/go-gorm/gorm/issues/4307

More information on testing with this bug can be found in this comment on that bug issue:

https://github.com/go-gorm/gorm/issues/4307#issuecomment-844129524

> EDIT: Apparently, this isn't a bug? It's been declared as intentional functionality.
>
> https://github.com/go-gorm/gorm/issues/4307#issuecomment-845603150

Unfortunately this change is big because we have to modify all the tests to account for `?` in place of `$` for the queries.

To accomplish this, I followed one of GORM's documented methods of using the dry run query functionality:

https://gorm.io/docs/sql_builder.html#DryRun-Mode


KellyMerrick (approved):
lgtm

PR #398, opened 2021-05-21
refactor(database)!: switch to use postgres.New()

jbrockopp:
Part of the effort for https://github.com/go-vela/community/issues/248

Related to https://github.com/go-vela/server/pull/341

This updates the codebase to now utilize the `github.com/go-vela/server/database/postgres.New()` function:

https://github.com/go-vela/server/blob/82a4db624dd03940accdfd6d795be036271ca90e/cmd/vela-server/database.go#L33-L48

Along with this change, comes a small update to the order of values provided for two existing database functions:

https://github.com/go-vela/server/blob/82a4db624dd03940accdfd6d795be036271ca90e/database/database.go#L40-L45

In the old database client, these previously were declared like so:

https://github.com/go-vela/server/blob/7c7205ca14d4d99d976f8d35f5f97fd0aa3d7b8b/database/database.go#L40-L45

The two `int` values provided to those functions are `page` and `perPage` for API pagination.

The reason for this change has to do with updating these functions to be consistent with all other functions.

For all other functions, when providing the `page` and `perPage` arguments, we always declare those at the end:

https://github.com/go-vela/server/blob/breaking/refactor/database/postgres/database/database.go

### DIFF
--- a/api/build.go
+++ b/api/build.go
@@ -362,7 +362,7 @@ func GetBuilds(c *gin.Context) {
 
 	// send API call to capture the list of builds for the repo (and event type if passed in)
 	if len(event) > 0 {
-		b, t, err = database.FromContext(c).GetRepoBuildListByEvent(r, page, perPage, event)
+		b, t, err = database.FromContext(c).GetRepoBuildListByEvent(r, event, page, perPage)
 	} else {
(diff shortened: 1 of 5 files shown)


KellyMerrick (approved):
lgtm

meinenec (approved):

PR #401, opened 2021-05-23
refactor(spec): build spec as part of release

wass3r:
This PR adds a Github Actions workflow to release the api-spec as an asset on a `release` event, mimicking what we are doing in `go-vela/types` for the pipeline JSON schema.

Includes the following changes:
- changes invocation for generating spec from `make swagger` to `make spec`
- creates three new `make` targets:
    - `spec-install` to install the dependencies for generating a spec (only meant for CI since it assumes debian-based environment provided by `golang:<version>` docker images)
    - `spec-gen` to execute go-swagger to create the spec file
    - `spec-version-update` to grab the current git tag and update the generated spec JSON with the correct version
- adds a new GitHub Actions workflow (`spec.yml`) to trigger on a `release` event to generate and upload the asset
- changes the hard-coded metadata for `version` to be `0.0.0-dev` (open to suggestions) - a placeholder for when someone is generating locally and doesn't have the required tools to update the version accurately
- removes current spec file from git


jbrockopp (approved):
LGTM

kneal (approved):
LGTM 🐬

meinenec (approved):

PR #409, opened 2021-05-27
fix(postgres): add workaround for gorm record not found

jbrockopp:
Leftover work for https://github.com/go-vela/community/issues/248

Related to https://github.com/go-gorm/gorm/issues/4416

**This is a workaround to a problem found in an upstream library.**

This updates the `go-vela/server/database/postgres` client to specifically check for the rows returned from queries:

https://github.com/go-vela/server/blob/95e6969b501087ab414fca37f8a60ff1faed9736/database/postgres/build.go#L35-L38

You'll note that we are explicitly checking the `RowsAffected` field is equal to `0` from the result returned by GORM.

This is required due to what initially appears to be a bug in GORM (See above [go-gorm/gorm](https://github.com/go-gorm/gorm) issue for more details).

GORM no longer returns `record not found` when a query is ran against the database and no rows are returned.

### DIFF
--- a/database/postgres/build.go
+++ b/database/postgres/build.go
@@ -18,6 +18,8 @@ import (
 )
 
 // GetBuild gets a build by number and repo ID from the database.
+//
+// nolint: dupl // ignore false positive of duplicate code
 func (c *client) GetBuild(number int, r *library.Repo) (*library.Build, error) {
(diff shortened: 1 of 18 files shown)


KellyMerrick (approved):
lgtm

matt-fevold (approved):

PR #393, opened 2021-05-20
feat(repo): use repo counter as source for build number

kneal:
PR includes the following:

* Updates to database pkg for new repo `counter` field
* Updates to repo `PUT` endpoint for setting the counter field
* Updates to webhook and build restart to make build number sync with repo counter

Related to: https://github.com/go-vela/community/issues/137

Depends on: https://github.com/go-vela/types/pull/176 _should fail without this change in the mod_

PR #425, opened 2021-06-08
fix(api): setting repo counter and build numbers

jbrockopp:
Related to https://github.com/go-vela/server/pull/393

This change will ensure that the `repo.Counter` and `build.Number` fields are set properly.

This is required because Vela attempts to utilize a "retry" logic when processing webhooks.

In order to ensure we set the proper `repo.Counter` and `build.Number` fields, we need to move the logic responsible for incrementing these values inside the retry logic.

### DIFF
--- a/api/webhook.go
+++ b/api/webhook.go
@@ -370,8 +370,14 @@ func PostWebhook(c *gin.Context) {
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


wass3r (approved):

KellyMerrick (approved):
lgtm

PR #429, opened 2021-06-15
feat: output version info to stdout

jbrockopp:
This updates the `github.com/go-vela/server` app to output the version information to stdout on startup:

```sh
$ docker logs server
{
  "canonical": "v0.8.2",
  "major": 0,
  "minor": 8,
  "patch": 2,
  "metadata": {
    "architecture": "amd64",
    "build_date": "2021-06-15T08:57:00Z",
    "compiler": "gc",
    "git_commit": "7ed708324825cc05e596306093a2855da47330c7",
    "go_version": "go1.16.5",
    "operating_system": "linux"
  }
}
<more logs>
```

### DIFF
--- a/cmd/vela-server/main.go
+++ b/cmd/vela-server/main.go
@@ -183,7 +183,8 @@ func main() {
 	// set logrus to log in JSON format
 	logrus.SetFormatter(&logrus.JSONFormatter{})
 
-	if err := app.Run(os.Args); err != nil {
+	err = app.Run(os.Args)
+	if err != nil {
 		logrus.Fatal(err)
 	}
 }
--- a/cmd/vela-server/main.go
+++ b/cmd/vela-server/main.go
@@ -5,6 +5,8 @@
 package main
 
 import (
+	"encoding/json"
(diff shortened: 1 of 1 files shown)


matt-fevold (approved):
great idea - LGTM

matt-fevold (changes requested):
When reviewing the worker PR i see you updated one `err :=` ref later on to `err =` 

Is that not needed here too?

https://github.com/go-vela/server/pull/429/files#diff-dc189c3a5470d24230c46570e8bbb88c1faae799eac6e75f45c01ad204bdcab4L172

jbrockopp:
@matt-fevold the reason that change isn't applied here is because the assignment to `err` happens inside of the `if`.

However, I'll update the code here to look more like the worker that way we are consistent! 👍

jbrockopp:
@matt-fevold this should be complete 😃

matt-fevold (approved):
Ahhh makes sense :) Noice!

wass3r (approved):

PR #438, opened 2021-06-28
fix(api): assume more results exist if list not empty

JordanSussman:
This resolves the issue where the `/api/v1/user/source/repos` endpoint was incorrectly returning `active: false` for some repositories when the given user had >= 100 repos returned by the aforementioned endpoint.

### DIFF
--- a/api/user.go
+++ b/api/user.go
@@ -423,7 +423,7 @@ func GetUserSourceRepos(c *gin.Context) {
 			// add repos to list of database org repos
 			dbRepos = append(dbRepos, dbReposPart...)
 
-			// assume another page exists if under 100 results are returned
+			// assume no more pages exist if under 100 results are returned
 			//
 			// nolint: gomnd // ignore magic number
 			if len(dbReposPart) < 100 {
--- a/api/user.go
+++ b/api/user.go
@@ -423,13 +423,13 @@ func GetUserSourceRepos(c *gin.Context) {
 			// add repos to list of database org repos
 			dbRepos = append(dbRepos, dbReposPart...)
 
-			// assume another page exists if any results are returned
+			// assume another page exists if under 100 results are returned
 			//
 			// nolint: gomnd // ignore magic number
-			if len(dbReposPart) >= 1 {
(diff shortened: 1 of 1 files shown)


api/user.go
```diff
@@ -423,10 +423,10 @@ func GetUserSourceRepos(c *gin.Context) {
 			// add repos to list of database org repos
 			dbRepos = append(dbRepos, dbReposPart...)
 
-			// making an assumption that 50 means there is another page
+			// assume another page exists if any results are returned
 			//
 			// nolint: gomnd // ignore magic number
-			if len(dbReposPart) == 50 {
+			if len(dbReposPart) >= 1 {
```
wass3r, line 429: I think the original intent was to keep paging if the per_page limit (`100`) was hit from the query here: https://github.com/go-vela/server/blob/09d7009b1bbfae2081178be372b9f38109be9da7/api/user.go#L414 It's unlikely there's another page if the current result set is fewer than `100`. So, I think instead of the original `50`, it should probably match that `per_page` value of `100`. Does that sound right? Or are you trying to solve another issue?

JordanSussman: @wass3r I've updated the code to utilize the same `100` parameter for comparison to know if another page exists. The end result is the same as my prior logic but this change should reduce the number of redundant database queries.

api/user.go
```diff
@@ -423,13 +423,13 @@ func GetUserSourceRepos(c *gin.Context) {
 			// add repos to list of database org repos
 			dbRepos = append(dbRepos, dbReposPart...)
 
-			// making an assumption that 50 means there is another page
+			// assume another page exists if under 100 results are returned
```
wass3r:
```suggestion
			// assume this is the last page if under 100 results are returned
```
I think we want this comment changed to something like this ^ ?

api/user.go
wass3r: Thanks!

wass3r (approved):

jbrockopp (approved):
LGTM

PR #444, opened 2021-07-08
feat(pipeline): enable usage of template languages within base 

JordanSussman:
Changes included in PR:

- Enable the user to consume template languages (golang & starlark) within the `.vela.yml` without `templates`
- Update pipeline endpoints to expand base configuration when starlark and golang is utilized
- Resolve some of the outstanding `.golangci` issues
- Update `.golangci` configuration to remove deprecated linters

Co-authored-by: @kneal 

xref: https://github.com/go-vela/community/issues/326


go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/go-vela/server
 
+replace github.com/go-vela/compiler => github.com/JordanSussman/compiler v0.1.3-0.20210709181443-362fa898a72c
```
JordanSussman, line 3: This will be removed once https://github.com/go-vela/compiler/pull/199 has been merged.

api/pipeline.go
```diff
@@ -723,3 +723,37 @@ func setTemplateLinks(c *gin.Context, u *library.User, r *library.Repo, template
 
 	return m, nil
 }
+
+// parseConfig returns the parsed yaml.Build from the input config.
+func parseConfig(comp compiler.Engine, config []byte, r *library.Repo) (*yaml.Build, error) {
```
jbrockopp, line 728: Thoughts on this logic getting pushed to the compiler?

To be clear, I'm not saying we need to make that change now 👍 

More so, I'm looking to understand what problems exist that would prevent this from being in the compiler?

I believe the compiler `Parse()` function should be able to handle this.

jbrockopp (changes requested):
Requesting changes for visibility purposes.

No action is actually required at this time.

api/pipeline.go
kneal: So, this function is largely written to replicate this behavior in the compiler:
https://github.com/go-vela/compiler/blob/master/compiler/native/compile.go#L50-L79

The main issue with the `Parse` function in the compiler is the series of func calls regardless of passing the interface as bytes, reader, etc ends up at this part of the code:
https://github.com/go-vela/compiler/blob/master/compiler/native/parse.go#L70-L74

A `.star` or YAML with Go template syntax can not be parsed as YAML and crashes So, we`ParseRaw` which was a new function added in the change ([compiler#199](https://github.com/go-vela/compiler/pull/199/files#diff-8f4638cf10135b35bdfa69f72231e0f42805738a1fd4bcc08a57e51b0605a956R34-R35)) to get parsing functional with none YAML file types:
https://github.com/go-vela/compiler/blob/2170e6d2aacaf38ad6757e996f551796b1998749/compiler/native/parse.go#L19-L40

Honestly, the whole thing could exist in the compiler if we moved some of the parsing logic into the parse func itself. So, this switch statement:
https://github.com/go-vela/compiler/blob/2170e6d2aacaf38ad6757e996f551796b1998749/compiler/native/compile.go#L55-L73

kneal: I think @JordanSussman had that additional part written before I started helping with the feature. So he might be able to provide a more detailed response if what I said doesn't make sense.

JordanSussman: @kneal did a good job summarizing the current scenario. Thoughts on making a breaking change within the `Parse` function to handle all of this behavior?

jbrockopp: @kneal thanks for the information and insight! That helps 🎉 

@JordanSussman I think I could go either way 🤷 

On the one hand, I think it is an arguably better long-term solution to move this logic to the compiler. I believe it gives us less duplicate code across the codebases. I also feel there is a value proposition for pushing the responsibility from the server to the compiler to handle the different types of pipelines.

On the other hand, I think that it comes with added complexity in the compiler and more time/effort in order to deliver this feature.

Does anyone else feel strongly one way or the other?

If not, we could accept this as is, and the plan to make a future PR that moves it to the compiler?

JordanSussman: Updated the `Parse` function within https://github.com/go-vela/compiler/pull/207 to address this feedback.

jbrockopp (approved):
LGTM

wass3r (approved):

PR #446, opened 2021-07-12
enhance(builds): skip executing if only clone and/or init steps/stages are present

JordanSussman:
Couple of questions that should be considered:

- Should we put this functionality behind some sort of feature flag within the `metadata` section of the `.vela.yml`? Example:

  ```yaml
  version: "1"
  
  metdata:
    skip_empty: false|true
  
  steps:
    - name: foo
      image: alpine
      commands:
        - echo hello
      ruleset:
        event: [pull_request]
  ```
  
- Should we update the other build endpoints to have the same skip logic?

closes https://github.com/go-vela/community/issues/273


jbrockopp (commented):
Thanks for starting the efforts on this 👍 

> Should we put this functionality behind some sort of feature flag within the `metadata` section of the `.vela.yml`?

I'm hesitant to say **yes** to this and am leaning towards not having it be a feature flag.

Before I provide my reasoning, I want to acknowledge that the original issue opened (https://github.com/go-vela/community/issues/273) for this feature request did ask for it to be an on/off switch (configurable) 👀 

I'm unsure about putting this behind a feature flag because of my lack of creativity when trying to come up with a valid reason for why someone would need this turned off.

To reword that in a more digestible context:

Can we come up with a valid use-case for why a user would want a Vela build to run with only the `init` and `clone` steps? 🤔 

And to be clear, I'm not saying we can't have this behind a feature flag. More so, I'm struggling with justifying why we need this to have a feature flag.

> Should we update the other build endpoints to have the same skip logic?

Yes, I think that makes sense to do so to keep the behavior consistent 👍

JordanSussman:
> Can we come up with a valid use-case for why a user would want a Vela build to run with only the init and clone steps?

I'm thinking one may potentially want to disable the skip behavior for testing purposes. Regardless, I would likely have the `skip_empty` default to `true` to force the user to opt into the build running when only `clone` and `init` steps are present.. I'll plan on not implementing the feature flag unless someone has more concrete reasons why it would be needed.

> Yes, I think that makes sense to do so to keep the behavior consistent 

Sounds good.

kneal:
I'm honestly not sure why there would be a valid use case for someone just wanting to run a `init` and `clone` pipeline since they are injected only steps. 

I suppose there could be some arguments around testing but I think in general would be good to throw these away by default. If we wanted to make it a feature flag, I think the user should have to turn it on intentionally.

wass3r:
Do we need to account for cases where people choose to skip injecting the default clone step (https://go-vela.github.io/docs/reference/yaml/metadata/#the-clone-tag) ?

JordanSussman:
> Do we need to account for cases where people choose to skip injecting the default clone step (https://go-vela.github.io/docs/reference/yaml/metadata/#the-clone-tag) ?

Just pushed a slight refactor that accounts for the scenario where only the `init` step is present.

wass3r (approved):

jbrockopp (approved):
LGTM

PR #443, opened 2021-07-08
enhance(api): support querying builds by branch/status

jbrockopp:
Part of the effort for https://github.com/go-vela/community/discussions/321#discussioncomment-974885 and https://github.com/go-vela/community/issues/329

This updates the `GET` endpoint for the path `/api/v1/repos/:org/:repo/builds` to support additional query parameters.

Currently, only the `event` query parameter is supported which allows you to filter the list of builds returned for a repo.

i.e. `curl https://vela.company.com/api/v1/repos/myOrg/myRepo/builds?event=push`

This change will augment the endpoint to now support both `branch` and `status` as query parameters.

i.e.

* `curl https://vela.company.com/api/v1/repos/myOrg/myRepo/builds?branch=main`
* `curl https://vela.company.com/api/v1/repos/myOrg/myRepo/builds?status=success`

It's also important to note that we made sure that you can combine the query parameters together to filter even further:

i.e. `curl https://vela.company.com/api/v1/repos/myOrg/myRepo/builds?branch=main&event=push&status=success`

Also worth mentioning is the extra validation checks added to the `event` and `status` query parameters.

Both parameters have a set of defined values that are valid so we now check for that:

* https://github.com/go-vela/types/blob/v0.8.2/constants/event.go#L7-L23
* https://github.com/go-vela/types/blob/v0.8.2/constants/status.go#L7-L29

For example, if you were to submit the following request:

```
curl https://vela.company.com/api/v1/repos/myOrg/myRepo/builds?event=foobar
```

You would receive an error message like:

```
unable to process event foobar: invalid event type provided
```

Co-authored-by: @Cappo
Co-authored-by: @jschulz79 
Co-authored-by: @matt-fevold

PR #453, opened 2021-07-26
feat: add route for repos at org level

JayCeeJr:
Related Issue: https://github.com/go-vela/community/issues/332

Add route to get a list of repos enabled for an org: /api/v1/repos/::orgName

Move route to get a list a builds at an org level (this route appears to be unused by the UI, I plan to use it for the UI portion outlined in the issue above):
From api/v1/repos/::orgName
To: api/v1/repos/::orgName/builds


api/repo.go
```diff
@@ -356,6 +356,125 @@ func GetRepos(c *gin.Context) {
...
+//         "$ref": "#/definitions/Repo"
+//     headers:
+//       X-Total-Count:
+//         description: Total number of results
+//         type: integer
+//       Link:
+//         description: see https://tools.ietf.org/html/rfc5988
+//         type: string
+//   '400':
+//     description: Unable to retrieve the repo
```
JordanSussman, line 400:
```suggestion
//     description: Unable to retrieve the org
```

api/repo.go
```diff
@@ -356,6 +356,125 @@ func GetRepos(c *gin.Context) {
...
+//     schema:
+//       "$ref": "#/definitions/Error"
+//   '500':
+//     description: Unable to retrieve the repo
```
JordanSussman, line 404:
```suggestion
//     description: Unable to retrieve the org
```

api/repo.go
```diff
@@ -356,6 +356,125 @@ func GetRepos(c *gin.Context) {
...
+
+// GetOrgRepos represents the API handler to capture a list
+// of repos for a org from the configured backend.
+//
+// nolint: dupl // ignore false positive
+func GetOrgRepos(c *gin.Context) {
+	// capture middleware values
+	u := user.Retrieve(c)
+	org := c.Param("org")
+	logrus.Infof("Reading Org repos for org %s", org)
```
JordanSussman, line 416:
```suggestion
	logrus.Infof("Reading repos for org %s", org)
```

api/repo.go
```diff
@@ -356,6 +356,125 @@ func GetRepos(c *gin.Context) {
...
+	// create pagination object
+	pagination := Pagination{
+		Page:    page,
+		PerPage: perPage,
+		Total:   t,
+	}
+	// set pagination headers
+	pagination.SetHeaderLink(c)
+
+	// TODO: Show all public repos for the orgs Hide private repos unless user has access
```
JordanSussman, line 473: Seems like this should be implemented.

database/postgres/repo_count.go
```diff
@@ -28,6 +28,22 @@ func (c *client) GetRepoCount() (int64, error) {
 	return r, err
 }
 
+// GetOrgRepoCount gets a count of all repos for a specific org from the database.
+func (c *client) GetOrgRepoCount(org string) (int64, error) {
+	logrus.Tracef("getting count of repos for user %s in the database", org)
```
JordanSussman, line 33:
```suggestion
	logrus.Tracef("getting count of repos for org %s in the database", org)
```

database/sqlite/dml/repo.go
```diff
@@ -42,6 +42,13 @@ FROM repos
 WHERE org = ?
 AND name = ?
 LIMIT 1;
+`
+	// SelectOrgReposCount represents a query to select
+	// the count of repos for a user_id in the database.
```
JordanSussman, line 47:
```suggestion
	// the count of repos for a org in the database.
```

JordanSussman (changes requested):
initial feedback

api/repo.go
```diff
@@ -356,6 +356,125 @@ func GetRepos(c *gin.Context) {
...
+	t, err := database.FromContext(c).GetOrgRepoCount(org)
+	if err != nil {
+		retErr := fmt.Errorf("unable to get repo count for org %s: %w", org, err)
+
+		util.HandleError(c, http.StatusInternalServerError, retErr)
+
+		return
+	}
+
+	// send API call to capture the list of repos for the user
```
JordanSussman, line 454:
```suggestion
	// send API call to capture the list of repos for the org
```

api/repo.go
```diff
@@ -356,6 +356,125 @@ func GetRepos(c *gin.Context) {
...
+	r, err := database.FromContext(c).GetUserRepoList(u, page, perPage)
+	if err != nil {
+		retErr := fmt.Errorf("unable to get repos for user %s: %w", u.GetName(), err)
```
JordanSussman, line 457:
```suggestion
		retErr := fmt.Errorf("unable to get repos for org %s: %w", u.GetName(), err)
```

JordanSussman (changes requested):
more feedback

api/repo.go
```diff
@@ -356,6 +356,125 @@ func GetRepos(c *gin.Context) {
...
+	if err != nil {
+		retErr := fmt.Errorf("unable to get repo count for org %s: %w", org, err)
+
+		util.HandleError(c, http.StatusInternalServerError, retErr)
+
+		return
+	}
+
+	// send API call to capture the list of repos for the user
+	r, err := database.FromContext(c).GetUserRepoList(u, page, perPage)
```
JordanSussman, line 455:
```suggestion
	r, err := database.FromContext(c).GetOrgRepoList(u, page, perPage)
```

api/repo.go
JayCeeJr: Added filtering for both repos and builds based on private repo flag

api/build.go
```diff
@@ -514,11 +520,35 @@ func GetOrgBuilds(c *gin.Context) {
 	// nolint: gomnd // ignore magic number
 	perPage = util.MaxInt(1, util.MinInt(100, perPage))
 
+	privateRepos, err := database.FromContext(c).GetOrgPrivateRepoList(o)
+	if err != nil {
+		logrus.Errorf("unable to get private repos for org %s : %s", o, err)
```
JordanSussman, line 525: Should this return the error instead of moving on?

database/postgres/repo_list.go
```diff
@@ -54,8 +55,50 @@ func (c *client) GetRepoList() ([]*library.Repo, error) {
 	return repos, nil
 }
 
+// GetOrgPrivateRepoList gets a list of all private repos by org from the database.
+func (c *client) GetOrgPrivateRepoList(org string) ([]*library.Repo, error) {
+	logrus.Tracef("getting repos for org %s from the database", org)
```
JordanSussman, line 60:
```suggestion
	logrus.Tracef("getting private repos for org %s from the database", org)
```

api/build.go
```diff
@@ -514,11 +520,35 @@ func GetOrgBuilds(c *gin.Context) {
...
+	}
+
+	var excludeList []int64
+
+	for _, rr := range privateRepos {
+		perm, err := source.FromContext(c).RepoAccess(u, rr.GetOrg(), rr.GetName())
+		if err != nil {
+			logrus.Errorf("unable to get user %s access level for repo %s", u.GetName(), rr.GetFullName())
```
JordanSussman, line 533: Should this return the error instead of moving on?

JordanSussman (commented):
more input

api/build.go
JayCeeJr: I'll have it fail here, since the DB would most likely be the issue and it will fail on the next call as well.

api/build.go
JayCeeJr: If the github api is down or rate limited this will fail safe and hide the private repos. I don't think it would be ideal to fail the whole process because of a github issue.

wass3r:
looks like the merge of https://github.com/go-vela/server/pull/444 created some conflicts - sorry :(

api/repo.go
```diff
@@ -372,6 +372,146 @@ func GetRepos(c *gin.Context) {
...
+		switch perm {
+		case "admin", "write", "read":
+			continue
+		default:
+			excludeList = append(excludeList, rr.GetName())
+		}
+	}
+	if len(excludeList) == 0 {
+		excludeList = append(excludeList, "")
+	}
```
wass3r, line 481: do we actually need an blank entry? could we instead of `var excludeList []string` do `excludeList := []string{}` and avoid this?

api/repo.go
```diff
@@ -372,6 +372,146 @@ func GetRepos(c *gin.Context) {
...
+
+		return
+	}
+
+	// ensure per_page isn't above or below allowed values
+	//
+	// nolint: gomnd // ignore magic number
+	perPage = util.MaxInt(1, util.MinInt(100, perPage))
+
+	allRepos, err := database.FromContext(c).GetOrgPrivateRepoList(org)
```
wass3r, line 459: From here to the below loop and such, can we add some more comments about why we're doing this?

database/postgres/repo_count_test.go
```diff
@@ -162,3 +162,74 @@ func TestPostgres_Client_GetUserRepoCount(t *testing.T) {
 		}
 	}
 }
+
+func TestPostgres_Client_GetOrgRepoCount(t *testing.T) {
```
wass3r, line 166: I may have missed it but I think we want a test for the exclude functionality as well, right?

database/postgres/repo_list_test.go
```diff
@@ -148,7 +148,81 @@ func TestPostgres_Client_GetOrgRepoList(t *testing.T) {
 
 	// run tests
 	for _, test := range tests {
-		got, err := _database.GetOrgRepoList("foo", 1, 10)
+		got, err := _database.GetOrgRepoList("foo", []string{""}, 1, 10)
```
wass3r, line 151: similar to previous comment, we want a test where the exclude is populated as well, right?

database/postgres/build_list_test.go
```diff
@@ -143,7 +143,7 @@ func TestPostgres_Client_GetOrgBuildList(t *testing.T) {
 
 	// run tests
 	for _, test := range tests {
-		got, _, err := _database.GetOrgBuildList("foo", 1, 10)
+		got, _, err := _database.GetOrgBuildList("foo", []int64{0}, 1, 10)
```
wass3r, line 146: include a test with populated exclusion list?

database/postgres/build_count_test.go
```diff
@@ -190,7 +190,7 @@ func TestPostgres_Client_GetOrgBuildCount(t *testing.T) {
 
 	// run tests
 	for _, test := range tests {
-		got, err := _database.GetOrgBuildCount("foo")
+		got, err := _database.GetOrgBuildCount("foo", []int64{0})
```
wass3r, line 193: include a test with populated exclusion list?

wass3r (changes requested):
thanks for the work/PR.. have few requested changes. let me know if you have questions.

api/repo.go
JayCeeJr: I did that initially, and the query didn't return any results. The empty string will just add an empty string in the where clause. 
```sql
SELECT ...
WHERE name NOT IN ("")
```
And there will never be a repo with an empty string as a name.

api/repo.go
```diff
@@ -372,6 +372,146 @@ func GetRepos(c *gin.Context) {
...
+	if err != nil {
+		logrus.Errorf("unable to get private repos for org %s : %s", org, err)
+	}
+
+	var excludeList []string
+
+	for _, rr := range allRepos {
+		perm, err := source.FromContext(c).RepoAccess(u, rr.GetOrg(), rr.GetName())
```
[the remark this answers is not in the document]
JayCeeJr: I think we could do this, it will add another api call to github, for non-admins. however reduce the number of calls to one for admins.

api/repo.go
wass3r: OK thanks, that's what I feared. Can we add a comment above the code block that adds the empty entry to call that out more?

api/repo.go
```diff
@@ -372,6 +372,146 @@ func GetRepos(c *gin.Context) {
...
+		if err != nil {
+			logrus.Errorf("unable to get user %s access level for repo %s", u.GetName(), rr.GetFullName())
+		}
+
+		switch perm {
+		case "admin", "write", "read":
```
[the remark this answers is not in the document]
JayCeeJr: True, Not sure how we would address that. I believe the perm call is a middleware function, so we do have some control over what is returned.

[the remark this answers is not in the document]
wass3r: well, the middleware calls `source.FromContext(c).` methods. I guess I was thinking we might have to add to the service interface (see `source/service.go` and implement it for `github`.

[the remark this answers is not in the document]
wass3r: @jbrockopp thoughts?

database/postgres/repo_list_test.go
JayCeeJr: I'll add a test, however we are mocking the result-set, so not sure it accomplishes anything.

Any plans on setting up a Postgres instance for tests?

wass3r: Not sure I follow. We create a test database and seed it with data and then mock the response we expect to get given the test.

api/build.go
```diff
@@ -553,11 +556,45 @@ func GetOrgBuilds(c *gin.Context) {
...
 
+	// See if the user is an org admin to bypass individual permission checks
+	perm, err := source.FromContext(c).OrgAccess(u, o)
+	if err != nil {
+		logrus.Errorf("unable to get user %s access level for org %s", u.GetName(), o)
+	}
+	var excludeList []int64
+	if perm != "admin" {
+		// Get a list of private repos to filter out if the user does not have permission to them
+		privateRepos, err := database.FromContext(c).GetOrgPrivateRepoList(o)
```
jbrockopp, line 567: I think we're missing a couple parameters to provide to this function specifically around pagination.

```suggestion
		privateRepos, err := database.FromContext(c).GetOrgPrivateRepoList(o, page, perPage)
```

api/repo.go
```diff
@@ -372,6 +372,155 @@ func GetRepos(c *gin.Context) {
...
+
+	// See if the user is an org admin to bypass individual permission checks
+	perm, err := source.FromContext(c).OrgAccess(u, org)
+	if err != nil {
+		logrus.Errorf("unable to get user %s access level for org %s", u.GetName(), org)
+	}
+	var excludeList []string
+	if perm != "admin" {
+		// Get a list of private repos to filter out if the user does not have permission to them
+		privateRepos, err := database.FromContext(c).GetOrgPrivateRepoList(org)
```
jbrockopp, line 466: I think we're missing a couple parameters to provide to this function specifically around pagination.

```suggestion
		privateRepos, err := database.FromContext(c).GetOrgPrivateRepoList(org, page, perPage)
```

database/postgres/build_list.go
```diff
@@ -170,3 +171,46 @@ func (c *client) GetRepoBuildList(r *library.Repo, filters map[string]string, pa
 
 	return builds, count, err
 }
+
+// GetRepoBuildListByEvent gets a list of all builds by repo and event type from the database.
+//
+// nolint: lll // ignore long line length due to variable names
+func (c *client) GetRepoBuildListByEvent(r *library.Repo, event string, page, perPage int) ([]*library.Build, int64, error) {
```
jbrockopp, line 178: I don't think we want to add this function as it was removed with:

https://github.com/go-vela/server/pull/443

database/postgres/dml/build.go
```diff
@@ -100,6 +103,14 @@ SELECT count(*) as count
 FROM builds
 WHERE repo_id = ?
 AND event = ?;
+`
+	// SelectRepoBuildsByEvent represents a query to select
+	// the count of builds for by repo and event type in the database.
+	SelectRepoBuildsByEvent = `
```
jbrockopp, line 109: I don't think we want to add this query as it was removed with:

https://github.com/go-vela/server/pull/443

database/postgres/dml/repo.go
```diff
@@ -21,6 +21,16 @@ WHERE user_id = ?
...
+`
+
+	// ListPrivateOrgRepos represents a query to list
+	// all repos for an org in the database.
+	ListPrivateOrgRepos = `
+SELECT *
+FROM repos
+WHERE org = ?
+AND visibility = 'private'
+ORDER BY id DESC;
```
jbrockopp, line 33: I think we're missing `LIMIT` and `OFFSET` from this query.

AFAICT, this would capture every repo in the given org with a visibility of `private`.

This can cause a problem if the number of repos returned by the query is abnoramlly large.

You can use the below for reference:

https://github.com/go-vela/server/blob/fefd9eec0fbad52a17717104b23375a0bcf075e9/database/postgres/dml/repo.go#L15-L23

database/postgres/repo_list.go
```diff
@@ -55,8 +55,50 @@ func (c *client) GetRepoList() ([]*library.Repo, error) {
 	return repos, nil
 }
 
+// GetOrgPrivateRepoList gets a list of all private repos by org from the database.
+func (c *client) GetOrgPrivateRepoList(org string) ([]*library.Repo, error) {
```
jbrockopp, line 59: Think we're missing a couple parameters here for pagination (`page` and `perPage`):

```suggestion
func (c *client) GetOrgPrivateRepoList(org string, page, perPage int) ([]*library.Repo, error) {
```

database/sqlite/build_list.go
```diff
@@ -171,3 +172,46 @@ func (c *client) GetRepoBuildList(r *library.Repo, filters map[string]string, pa
 
 	return builds, count, err
 }
+
+// GetRepoBuildListByEvent gets a list of all builds by org name and event type from the database.
+//
+// nolint: lll // ignore long line length due to variable names
+func (c *client) GetRepoBuildListByEvent(r *library.Repo, event string, page, perPage int) ([]*library.Build, int64, error) {
```
jbrockopp, line 179: I don't think we want to add this function as it was removed with:

https://github.com/go-vela/server/pull/443

database/sqlite/dml/build.go
```diff
@@ -114,6 +118,20 @@ SELECT count(*) as count
 FROM builds
 WHERE repo_id = ?
 AND event = ?;
+`
+	// ListRepoBuildsByEvent represents a joined query
+	// between the builds & repos table to select
+	// a build for an org with a specific event type
+	// in the database.
+	ListRepoBuildsByEvent = `
```
jbrockopp, line 126: I don't think we want to add this query as it was removed with:

https://github.com/go-vela/server/pull/443

database/sqlite/dml/repo.go
```diff
@@ -21,6 +21,16 @@ WHERE user_id = ?
...
 OFFSET ?;
+`
+	// ListOrgPrivateRepos represents a query to list
+	// all repos for an org in the database.
+	ListOrgPrivateRepos = `
+SELECT *
+FROM repos
+WHERE org = ?
+AND visibility = 'private'
+ORDER BY id DESC
```
jbrockopp, line 32: I think we're missing `LIMIT` and `OFFSET` from this query.

AFAICT, this would capture every repo in the given org with a visibility of `private`.

This can cause a problem if the number of repos returned by the query is abnoramlly large.

You can use the below for reference:

https://github.com/go-vela/server/blob/fefd9eec0fbad52a17717104b23375a0bcf075e9/database/sqlite/dml/repo.go#L15-L23

jbrockopp (changes requested):

api/build.go
JayCeeJr: @jbrockopp @wass3r 

After thinking about this for a while, maybe we just want to exclude builds and repos that are private from the Org pages. 
We took this approach to avoid data leakage, however I think the overhead of checking permissions by hitting the github api will not scale well.

jbrockopp: @JayCeeJr I'm not opposed to that approach and I think that's a fair way to handle it 👍 

We can always explore attempting to add this functionality in at a later date if it's found valuable by our users.

And perhaps in the future we'll be storing user permissions for orgs and repos in our database?

Thinking that maybe it would help resolve some concerns with taxing the GitHub API.

database/postgres/repo_list_test.go
wass3r: that's fair. i was thinking about sqlite, i think :)

api/build.go
JayCeeJr: @jbrockopp 
I took a middle-ground approach to this. I check if the user is an admin, if not i only show public repos for the org. Same logic applies to the builds.

wass3r (approved):
thanks a ton for working through the changes 👍🏼

jbrockopp (approved):
LGTM

jbrockopp:
@JordanSussman we'll plan to merge this PR once it has your approval since you had requested feedback 👍

JordanSussman (approved):

JordanSussman:
> @JordanSussman we'll plan to merge this PR once it has your approval since you had requested feedback 👍

Approved!

api/build.go
JordanSussman: May need to utilize the [list organization repos](https://docs.github.com/en/rest/reference/repos#list-organization-repositories) endpoint or [graphql](https://github.com/shurcooL/githubv4) in the future.

PR #454, opened 2021-07-26
feat: enable canceling abandoned builds

meinenec:
Fixes go-vela/community/issues/121

This PR allows user to set the status of a build to "canceled" for any builds that have a status of "running" but are not actually running. A scenarios where this may occur is when a worker dies or is turned off while running a build. The resulting build does not fail gracefully so it never gets a chance to update its status.

All running and pending steps will also be set to a canceled state, similarly to how it works for a running build that is canceled.


api/build.go
```diff
@@ -1262,6 +1262,44 @@ func CancelBuild(c *gin.Context) {
...
+	for page > 0 {
+		// retrieve build steps (per page) from the database
+		stepsPart, err := database.FromContext(c).GetBuildStepList(b, page, 100)
+		if err != nil {
+			retErr := fmt.Errorf("unable to retrieve steps for build %d: %w", b.Number, err)
+			util.HandleError(c, http.StatusNotFound, retErr)
+			return
+		}
+
+		// add repos to list of database org repos
```
JordanSussman, line 1281: This comment seems like it was copy/pasted without being updated.

api/build.go
```diff
@@ -1262,6 +1262,44 @@ func CancelBuild(c *gin.Context) {
...
+		//
+		// nolint: gomnd // ignore magic number
+		if len(stepsPart) < 100 {
+			page = 0
+		} else {
+			page++
+		}
+	}
+
+	// iterate over all each step for the build
```
wass3r, line 1294:
```suggestion
	// iterate over each step for the build
```
super minor

api/build.go
```diff
@@ -1262,6 +1262,54 @@ func CancelBuild(c *gin.Context) {
...
+	b.SetStatus(constants.StatusCanceled)
+	err = database.FromContext(c).UpdateBuild(b)
+	if err != nil {
+		retErr := fmt.Errorf("unable to update status for build %d: %w", b.Number, err)
+		util.HandleError(c, http.StatusInternalServerError, retErr)
+		return
+	}
+
+	// retrieve the steps for the build from the step table
+	steps := []*library.Step{}
```
jbrockopp, line 1275: I see that we're adding logic to handle canceling the `steps` attached to the build.

Thoughts on replicating that logic so we also handle `services` attached to the build?

jbrockopp (changes requested):

api/build.go
meinenec: Can't believe that didn't cross my mind. Let me add that quick!

jbrockopp: Sounds good 👍

meinenec: This function is quite the lengthy one though. Not sure how to refactor it to make this better, but we could consider adding some helper functions for getting all the steps/services for a build. Good with adding the logic here for now

jbrockopp (approved):
LGTM

wass3r (approved):

PR #455, opened 2021-07-30
feat(database): support skipping creation of tables and indexes

jbrockopp:
This adds a new `database.skip_creation` boolean flag to the `github.com/go-vela/server/database` package.

The intention of this flag is to enable skipping the creation of DDL objects (tables and indexes) in the database.

This flag can also be set via environment variables with:

* `VELA_DATABASE_SKIP_CREATION`
* `DATABASE_SKIP_CREATION`

This flag can also be set via a value in a file with the path `/vela/database/skip_creation`.

The use-case for this functionality is to enable Vela to support databases without admin access to that database.

This setup can happen when running Vela with a database that has extra security controls preventing DDL changes.

> NOTE: Most of the LOC found in this change are related to the tests added for this new flag.


matt-fevold (approved):
LGTM

wass3r (approved):

PR #458, opened 2021-08-04
feat: shared secrets enhancements

JayCeeJr:
Add ability to list all shared secrets under an Org.

Related issue: https://github.com/go-vela/community/issues/90


database/sqlite/secret_list.go
```diff
@@ -81,10 +81,21 @@ func (c *client) GetTypeSecretList(t, o, n string, page, perPage int) ([]*librar
...
+				Limit(perPage).
+				Offset(offset).
+				Scan(s).Error
+		} else {
+			err = c.Sqlite.
+				Table(constants.TableSecret).
+				Raw(dml.ListSharedSecrets, o, n, perPage, offset).
+				Scan(s).Error
+		}
+
```
jbrockopp, line 98: Could we remove this new line based off the linter feedback?

https://github.com/go-vela/server/actions/runs/1099137662

router/middleware/perm/perm.go
```diff
@@ -87,21 +87,38 @@ func MustSecretAdmin() gin.HandlerFunc {
...
+				perm, err := source.FromContext(c).OrgAccess(u, o)
+				if err != nil {
+					logrus.Errorf("unable to get user %s access level for org %s: %v", u.GetName(), o, err)
+				}
 
-			if !strings.EqualFold(perm, "admin") {
-				// nolint: lll // ignore long line length due to error message
-				retErr := fmt.Errorf("user %s does not have 'admin' permissions for the team %s/%s", u.GetName(), o, n)
+				if !strings.EqualFold(perm, "admin") {
+					retErr := fmt.Errorf("user %s does not have 'admin' permissions for the org %s", u.GetName(), o)
```
jbrockopp, line 99: Could we add a `nolint` comment based off the linter feedback?

https://github.com/go-vela/server/actions/runs/1099137662

```suggestion
					// nolint: lll // ignore long line length due to error message
					retErr := fmt.Errorf("user %s does not have 'admin' permissions for the org %s", u.GetName(), o)
```

jbrockopp (changes requested):

router/middleware/perm/perm.go
```diff
@@ -87,21 +88,39 @@ func MustSecretAdmin() gin.HandlerFunc {
...
+				if !strings.EqualFold(perm, "admin") {
+					// nolint: lll // ignore long line length due to error message
+					retErr := fmt.Errorf("user %s does not have 'admin' permissions for the org %s", u.GetName(), o)
 
-				util.HandleError(c, http.StatusUnauthorized, retErr)
+					util.HandleError(c, http.StatusUnauthorized, retErr)
 
-				return
+					return
+				}
```
jbrockopp, line 106: I might be misunderstanding what the above code is doing.

But if I'm reading this correctly, I think this would break the perms contract Vela has for `shared` secrets?

Looking at this change, it appears it would allow an org `admin` to get a list of `shared` secrets.

The problem with that is Vela always required a user to be a team `admin` to access `shared` secrets.

The idea or thought process is `shared` secrets are meant to be scoped to a specific team in the org.

jbrockopp (changes requested):

router/middleware/perm/perm.go
JayCeeJr: Wouldn't an `org` admin inherit admin for the teams? I think in GitHub `org` admins have full ability to manage teams under their `org`.

On the other hand if you are not an `org` admin you won't see any `shared` secrets in the `org` page. Which really isn't any different from today where you have to go directly to the shared secrets URL.

I'm not sure how else we would approach getting a list of shared secrets without abusing the GitHub api. One idea would be to collect some metadata about the user on login and store it in vela; it would be stale as soon as they made a change in GitHub. Maybe a refresh metadata route, that could be triggered on demand would alleviate that.

Metadata could include `orgs` the user belongs to, `teams` they are connected to along with `permissions`.

JordanSussman: > Wouldn't an org admin inherit admin for the teams? I think in GitHub org admins have full ability to manage teams under their org.

We could utilize the [/orgs/{org}/teams](https://docs.github.com/en/rest/reference/teams) to determine which team(s) the given user has access.

> I'm not sure how else we would approach getting a list of shared secrets without abusing the GitHub api.

Do we think that the average organization has many hundreds or thousands of teams? Looks like the previously mentioned endpoint can return 100 teams within the first request.

jbrockopp: > Wouldn't an `org` admin inherit admin for the teams?

No, I don't believe this is the case.

IIRC, we've designed this as intentional functionality in the system for specific use-cases.

> I think in GitHub `org` admins have full ability to manage teams under their `org`.

Yes I believe this is correct 👍 

An `org` admin can manage any `team` under that `org`.

---

Now I understand the above two statements might seem conflicting so I'm going to elaborate.

You can be an `org` admin, but still not have access to directly manage `shared` secrets for a `team` in that `org`.

Whenever a user attempts to access `shared` secrets, Vela specifically checks if they're an `admin` of the `team`:

https://github.com/go-vela/server/blob/e4801f81021eba1ab3d4909a57979c9a4b32365f/router/middleware/perm/perm.go#L89-L104

If they don't have access to the `team`, but they're an `org` admin, they'll face the following error:

```
user <user> does not have 'admin' permissions for the team <org>/<team>
```

However, like you pointed out above, an `org` admin can manage any `team` in that `org`.

This means that a user who is an `org` admin but isn't on a `team` in that `org` could go add themselves to the `team`.

Then, this would grant them the required access level to manage `shared` secrets for that `team` in Vela.

jbrockopp: @JayCeeJr and I hopped into a quick Slack call to discuss some of this 👍 

Based off what @JordanSussman mentioned above, we think we've found a good approach to enable this functionality.

This approach may use either the endpoint Jordan mentioned or the [`/user/teams`](https://docs.github.com/en/rest/reference/teams#list-teams-for-the-authenticated-user) endpoint.

Thinking this will help check what `teams` in an `org` the user has access to.

Then, we only have to list the `shared` secrets in those specific `teams`.

JordanSussman: > This approach may use either the endpoint Jordan mentioned or the /user/teams endpoint.

I'm thinking the `/user/teams` endpoint will be the most performant option.

source/github/access.go
```diff
@@ -59,7 +59,7 @@ func (c *client) RepoAccess(u *library.User, org, repo string) (string, error) {
 // TeamAccess captures the user's access level for a team.
 func (c *client) TeamAccess(u *library.User, org, team string) (string, error) {
 	logrus.Tracef("Capturing %s access level to team %s/%s", u.GetName(), org, team)
-
+	team = strings.ToLower(team)
```
jbrockopp, line 62: Do we need this `strings.ToLower()`?

source/github/access.go
```diff
@@ -87,7 +87,7 @@ func (c *client) TeamAccess(u *library.User, org, team string) (string, error) {
 	// iterate through each element in the teams
 	for _, t := range teams {
 		// skip the team if does not match the team we are checking
-		if !strings.EqualFold(team, t.GetName()) {
+		if !strings.EqualFold(team, strings.ToLower(t.GetName())) {
```
jbrockopp, line 90: Do we need this `strings.ToLower()`?

AFAIK, `strings.EqualFold()` implicitly handles case-insensitivity 👍 

So, I don't believe we have to worry about upper or lower casing the team name:

https://pkg.go.dev/strings#EqualFold

And a sample from Go Playground:

https://play.golang.org/p/5qGWfV6yzap

jbrockopp (changes requested):

source/github/access.go
JayCeeJr: Needed that for the database query, I moved the lowercase down to where the query is run and removed it from this function.

router/middleware/perm/perm.go
```diff
@@ -87,21 +88,40 @@ func MustSecretAdmin() gin.HandlerFunc {
...
+				logrus.Debugf("Verifying user %s has 'admin' permissions for org %s", u.GetName(), o)
+
+				// TODO: Get list of teams user is a part of and filter by team in []
+				teams, err := source.FromContext(c).ListUsersTeamsForOrg(u, o)
+				if err != nil {
+					logrus.Errorf("unable to get users %s teams for org %s: %v", u.GetName(), o, err)
+				}
+
+				if len(teams) == 0 {
+					// nolint: lll // ignore long line length due to error message
```
jbrockopp, line 101: According to the linter, looks like we can remove this line:

https://github.com/go-vela/server/actions/runs/1120266414

secret/native/list.go
```diff
@@ -11,11 +11,11 @@ import (
 )
 
 // List captures a list of secrets.
-func (c *client) List(sType, org, name string, page, perPage int) ([]*library.Secret, error) {
+func (c *client) List(sType, org, name string, page, perPage int, teams []string) ([]*library.Secret, error) {
```
jbrockopp, line 14: Think we'll have to add a `nolint` comment here:

```suggestion
//
// nolint: lll // ignore long line length
func (c *client) List(sType, org, name string, page, perPage int, teams []string) ([]*library.Secret, error) {
```

jbrockopp (changes requested):

jbrockopp (approved):
LGTM

wass3r (approved):
lgtm

PR #460, opened 2021-08-06
fix(deps): address vulnerable jwt-go library

jbrockopp:
Closes https://github.com/go-vela/community/issues/352

This updates the library the server uses to handle [JWT](https://jwt.io/) token processing.

Currently, we're using `github.com/dgrijalva/jwt-go`.

After this change, we'll be using `github.com/golang-jwt/jwt`.

This resolves the known vulnerability in the old JWT library reported by Dependabot:

https://github.com/go-vela/server/security/dependabot/go.mod/github.com%2Fdgrijalva%2Fjwt-go/open

> **CVE-2020-26160** - `high severity`
>
> **Vulnerable versions**: `<= 3.2.0`
>
> **Patched version**: `No fix`
>
> [jwt-go](https://github.com/dgrijalva/jwt-go) allows attackers to bypass intended access restrictions in situations with []string{} for m["aud"] (which is allowed by the specification). Because the type assertion fails, "" is the value of aud. This is a security problem if the JWT token is presented to a service that lacks its own audience check. There is no patch available and users of jwt-go are advised to migrate to [golang-jwt](https://github.com/golang-jwt/jwt) at version 3.2.1

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,13 +6,13 @@ require (
 	github.com/DATA-DOG/go-sqlmock v1.5.0
 	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/aws/aws-sdk-go v1.40.16
-	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/fatih/color v1.10.0 // indirect
 	github.com/gin-gonic/gin v1.7.3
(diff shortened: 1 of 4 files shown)


KellyMerrick (approved):
lgtm

wass3r (approved):

PR #463, opened 2021-08-10
fix: correct build number on restart

wass3r:
on restart we incremented the build counter ***after*** setting the link property which is utilized in the UI, resulting in the link pointing to the previous build.

this PR moves the build increment logic up to be called before we set the link for the build within the handler for the restart endpoint.

in my testing, this solved the issue for the below:

fixes: https://github.com/go-vela/community/issues/353
fixes: https://github.com/go-vela/community/issues/322

### DIFF
--- a/api/build.go
+++ b/api/build.go
@@ -755,6 +755,12 @@ func RestartBuild(c *gin.Context) {
 		return
 	}
 
+	// update the build numbers based off repo counter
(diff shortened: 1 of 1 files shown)


jbrockopp (approved):
LGTM

meinenec (approved):

PR #469, opened 2021-08-17
fix(database): sort by created instead of started

wass3r:
modification of https://github.com/go-vela/server/pull/464 that utilizes `created` instead of `started` for the sorting.

intent is to prevent items jumping in the list since it's more likely for `started` to be `0` (top of the list) and then updated, whereas `created` is typically set when added to the db.


jbrockopp (approved):
LGTM

meinenec (approved):

PR #470, opened 2021-08-17
fix: add second order to ensure order consistency

wass3r:
While https://github.com/go-vela/server/pull/469 helped, it's possible for the DB to return data in an inconsistent order when the `created` value is the same for two different builds. This PR adds a second order using the primary key (`id`) to help deliver a consistent order.


meinenec (approved):

KellyMerrick (approved):
lgtm

jbrockopp (approved):
LGTM

PR #471, opened 2021-08-19
feat: deployments with builds

JayCeeJr:
Return the last 3 builds related to the deployment in the deployment list.

Related Issue: https://github.com/go-vela/community/issues/364
Depends on: https://github.com/go-vela/types/pull/193


jbrockopp (changes requested):
Looks like there are conflicts in the `go.sum` file.

Could we get those resolved?

https://github.com/go-vela/server/pull/471/conflicts

go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/go-vela/server
 
+replace github.com/go-vela/types => github.com/JayCeeJr/types v0.9.1-beta
```
jbrockopp, line 3: Should this line be removed?

Looks like https://github.com/go-vela/types/pull/193 was merged 👍 

So I think you could vendor in the latest changes from that repo instead of this line.

jbrockopp (changes requested):

go.mod
JayCeeJr: I'll pull that line out

jbrockopp (approved):
LGTM

jbrockopp (approved):
LGTM

matt-fevold (changes requested):
LGTM, minor question on why 3 builds and not say 1 or 10, but that's just my own curiosity. also looks like there's more go.mod conflicts so once those are resolved I'll approve.

PR #478, opened 2021-09-02
refactor(api/pipeline): collapsed common code snippets into helper functions

GregoryDosh:
This refactors the API for the Pipelines to extract reused code bits into something slightly more maintainable and less error prone.  Almost all the end points used the exact same initial steps of pulling the repo and compiling the information together with slight nuances. There is still more refactor work that could be done in other parts of the API but those can come in other future PRs


GregoryDosh:
There are probably better design patterns I could use instead of `getBasePipeline` and returning a ton of (sometimes unused) instances. Perhaps the `getBasePipeline` should instead take in some parameters (expand true, or compile true etc) and only return the pipeline or something else?

GregoryDosh:
Alrighty, took another pass at breaking things apart and I believe it's a lot easier to understand what each endpoint is actually doing now. The code before was a bit harder to sift through to understand the nuances and differences between the endpoints but now it's fairly straightforward. Would love more feedback and eyes though if possible. The diff viewer in GitHub seems to get confused on which lines went where because of how much refactoring happened.

wass3r (approved):
nice deduping and cleanup - thanks!

api/pipeline.go
```diff
@@ -558,130 +324,77 @@ func ValidatePipeline(c *gin.Context) {
...
+	// capture query parameters
+	ref := ctx.DefaultQuery("ref", repo.GetBranch())
 
-			return
-		}
+	// send API call to capture the repo owner
+	user, err := database.FromContext(ctx).GetUser(repo.GetUserID())
+	if err != nil {
+		retErr := fmt.Errorf("unable to get owner for %s: %w", repo.GetFullName(), err)
+		return nil, nil, retErr
```
jbrockopp, line 362: I wonder if there is any value in keeping the `retErr` variable here?

Perhaps we condense this to a one liner?

```suggestion
		return nil, nil, fmt.Errorf("unable to get owner for %s: %w", repo.GetFullName(), err)
```

api/pipeline.go
```diff
@@ -558,130 +324,77 @@ func ValidatePipeline(c *gin.Context) {
...
 
-	// validate the yaml configuration
-	err = comp.Validate(p)
+	// send API call to capture the pipeline configuration file
+	config, err := source.FromContext(ctx).ConfigBackoff(user, repo, ref)
 	if err != nil {
-		// nolint: lll // ignore long line length due to error message
-		retErr := fmt.Errorf("unable to validate pipeline configuration for %s@%s: %w", r.GetFullName(), ref, err)
+		retErr := fmt.Errorf("unable to get pipeline configuration for %s: %w", repoName(ctx), err)
+		return nil, nil, retErr
```
jbrockopp, line 369: Perhaps we remove the `retErr` here and condense to one line?

```suggestion
		return nil, nil, fmt.Errorf("unable to get pipeline configuration for %s: %w", repoName(ctx), err)
```

api/pipeline.go
```diff
@@ -558,130 +324,77 @@ func ValidatePipeline(c *gin.Context) {
...
+	comp := compiler.FromContext(ctx).
+		WithMetadata(meta).
+		WithRepo(repo).
+		WithUser(user)
 
-		return
+	pipeline, err := comp.Parse(config)
+	if err != nil {
+		retErr := fmt.Errorf("unable to parse pipeline configuration for %s: %w", repoName(ctx), err)
+		return nil, nil, retErr
```
jbrockopp, line 381: Perhaps we remove the `retErr` here and condense to one line?

api/pipeline.go
```diff
@@ -692,23 +405,20 @@ func setTemplateLinks(c *gin.Context, u *library.User, templates yaml.TemplateSl
 		cl, err := github.New("", "")
 		if err != nil {
 			retErr := fmt.Errorf("unable to create compiler github client: %w", err)
-
 			return nil, retErr
```
jbrockopp, line 408: Perhaps we remove the `retErr` here and condense to one line?

api/pipeline.go
```diff
@@ -692,23 +405,20 @@ func setTemplateLinks(c *gin.Context, u *library.User, templates yaml.TemplateSl
...
 		}
 
 		// parse template source
 		src, err := cl.Parse(tmpl.GetSource())
 		if err != nil {
 			retErr := fmt.Errorf("unable to parse source for %s: %w", tmpl.GetSource(), err)
-
 			return nil, retErr
```
jbrockopp, line 415: Perhaps we remove the `retErr` here and condense to one line?

api/pipeline.go
```diff
@@ -692,23 +405,20 @@ func setTemplateLinks(c *gin.Context, u *library.User, templates yaml.TemplateSl
...
 		}
 
 		// retrieve link to template file from github
-		link, err := source.FromContext(c).GetHTMLURL(u, src.Org, src.Repo, src.Name, src.Ref)
+		link, err := source.FromContext(ctx).GetHTMLURL(u, src.Org, src.Repo, src.Name, src.Ref)
 		if err != nil {
 			retErr := fmt.Errorf("unable to get html url for %s/%s/%s/@%s: %w", src.Org, src.Repo, src.Name, src.Ref, err)
-
 			return nil, retErr
```
jbrockopp, line 422: Perhaps we remove the `retErr` here and condense to one line?

jbrockopp (changes requested):

api/pipeline.go
GregoryDosh: I can do a quick pass and clean-up and consolidate a lot of those to one liners. I was following the convention from the rest of the repo and figured it was because the err was being masked and to make it more apparent that the underlying error was being manipulated before being passed back.

jbrockopp (approved):
LGTM

KellyMerrick:
Looks like there are conflicts to resolve.

GregoryDosh:
@jbrockopp & @KellyMerrick I've rebased on master to pull in the source -> scm updates. Should be good to go now but needs another review

jbrockopp (approved):
LGTM

KellyMerrick (approved):
lgtm

wass3r:
you're good to merge it @GregoryDosh 👍🏼

PR #485, opened 2021-09-28
feat: allow setting a webhook-specific server address

cognifloyd:
`VELA_SOURCE_WEBHOOK_ADDR` (or `SOURCE_WEBHOOK_ADDR`) allows setting an alternative to `VELA_ADDR`/`VELA_HOST` so that github and other source providers use this alternate, generally externally facing, address to send webhooks. This allows `VELA_ADDR` to match `VELA_API` in the UI so that the oauth login flow makes sense.


source/flags.go
```diff
@@ -74,4 +74,14 @@ var Flags = []cli.Flag{
 		Usage:    "OAuth scopes to be used for the version control system",
 		Value:    cli.NewStringSlice("repo", "repo:status", "user:email", "read:user", "read:org"),
 	},
+	&cli.StringFlag{
+		EnvVars:  []string{"VELA_SOURCE_WEBHOOK_ADDR", "SOURCE_WEBHOOK_ADDR"},
+		FilePath: "/vela/source/webhook_addr",
+		Name:     "source.webhook.addr",
+		Usage: "Alternative or proxy server address as a fully qualified url (<scheme>://<host>). " +
+			"Use this when the Vela server address that the source provider can send webhooks to " +
+			"differs from the server address the UI and oauth flows use, such as when the server" +
```
meinenec, line 83: missing a space
```suggestion
			"differs from the server address the UI and oauth flows use, such as when the server " +
```

meinenec (changes requested):

cognifloyd:
I'm not sure what to do with reviewdog's complaints as they have nothing to do with what I've changed. Can I ignore them?

cognifloyd:
All tests are green!

cognifloyd:
I just finished deploying a build of this in my on-prem k8s cluster. I was able to successfully auth with github, and receive a github webhook. Here's my test setup:

I have these resources an on-prem k8s cluster:
- vela-server: 1 service (type: ClusterIP) + 1 deployment (1 replica for testing; more replicas should work)
- vela-ui: 1 service (type: ClusterIP) + 1 deployment (1 replica for testing; more replicas should work)
- vela-worker: no service, 1 deployment (1 replica for testing; more replicas should work - each worker uses its pod IP as hostname for registration)

I have these nodes outside the k8s cluster:
- nginx load balancer inside the firewall that handles connections to vela-server and vela-ui (ssl termination + load balance) each on a separate internal domain. This proxy_passes to the kubernetes services for vela-server and vela-ui (below)
- nginx load balancer that handles external connections (ie webhooks from github) and then proxy_pass them to the internal domain (ie the internal load balancer)

wass3r:
generally speaking.. don't worry about golangci-lint unless it complains about something you are adding/changing specifically. we'll be going through to clean up in all the repos at some point.

if you don't mind, could you undo the `init`, `clone` change for now (since that occurs elsewhere also, it might end up in `go-vela/types` potentially). feel free to undo any of the other changes you made to satisfy the linter for code you didn't touch.

also, very exciting and cool to hear that you got something going in your cluster. fair to assume that you used the helm chart you shared?

cognifloyd:
K. I dropped the unrelated changes.

Yes, I used charts from https://github.com/cognifloyd/vela-charts
https://cognifloyd.github.io/vela-charts

meinenec (approved):

kneal (approved):
LGTM 🐬

PR #489, opened 2021-10-01
feat(log streaming): add ability to stream step and service logs

KellyMerrick:
For https://github.com/go-vela/community/issues/366


wass3r (approved):

router/service.go
```diff
@@ -39,6 +40,8 @@ func ServiceHandlers(base *gin.RouterGroup) {
 			service.PUT("", perm.MustPlatformAdmin(), middleware.Payload(), api.UpdateService)
 			service.DELETE("", perm.MustPlatformAdmin(), api.DeleteService)
 
+			service.POST("/stream", api.PostServiceStream)
```
jbrockopp, line 43: I wonder if we should explore locking down this endpoint behind some permission layer?

Thinking for the other endpoints we call `perm.MustPlatformAdmin()` so perhaps we adopt that?

```suggestion
			service.POST("/stream", perm.MustPlatformAdmin(), api.PostServiceStream)
```

router/step.go
```diff
@@ -39,6 +40,8 @@ func StepHandlers(base *gin.RouterGroup) {
 			step.PUT("", perm.MustPlatformAdmin(), middleware.Payload(), api.UpdateStep)
 			step.DELETE("", perm.MustPlatformAdmin(), api.DeleteStep)
 
+			step.POST("/stream", api.PostStepStream)
```
jbrockopp, line 43: I wonder if we should explore locking down this endpoint behind some permission layer?

Thinking for the other endpoints we call `perm.MustPlatformAdmin()` so perhaps we adopt that?

```suggestion
			step.POST("/stream", perm.MustPlatformAdmin(), api.PostStepStream)
```

jbrockopp (changes requested):

router/service.go
wass3r: 👍🏼

router/step.go
wass3r: 👍🏼

wass3r (approved):

jbrockopp (approved):
LGTM

PR #526, opened 2021-11-03
feat(secrets)!: adding new metadata to secrets

ecrupper:
- Adding new columns in SQL databases for `created_at`, `created_by`, `updated_at`, `updated_by`, and `last_build_id`. 
- Updating tests so that all these new fields are accounted for
- Adding to api/secret.go to interact with these new fields when a user creates/updates a secret
- Needs to be paired with go-vela/types PR, which has the field changes


wass3r (approved):
thanks for this addition!

marked it as a breaking change due to DB updates needed.

KellyMerrick (approved):
lgtm

PR #527, opened 2021-11-05
feat(webhook): set new pusher fields on tag events

ecrupper:
This PR will be connected to another PR in go-vela/types in order to expose tagger information in the environment variables


source/github/webhook.go
```diff
@@ -77,8 +77,12 @@ func (c *client) VerifyWebhook(request *http.Request, r *library.Repo) error {
 // processPushEvent is a helper function to process the push event.
 func processPushEvent(h *library.Hook, payload *github.PushEvent) (*types.Webhook, error) {
 	logrus.Tracef("processing push GitHub webhook for %s", payload.GetRepo().GetFullName())
-
+	logrus.Tracef("THIS IS THE COMMENT BEFORE THE GOOD STUFF")
+	logrus.Tracef("Payload: %s", payload)
 	repo := payload.GetRepo()
+	pusherName := payload.GetPusher().Name
+	pusherEmail := payload.GetPusher().Email
```
JordanSussman, line 84: You likely want to utilize the [sender](https://github.com/google/go-github/blob/master/github/event_types.go#L824) since the [pusher](https://github.com/google/go-github/blob/master/github/event_types.go#L823) can incorrectly point to a different user if someone sends a commit with a different email/name with valid credentials. You can test the difference between pusher and sender by sending commits with `git commit --author="John Doe <john@doe.org>"`.

ecrupper: Interesting... the sender does not have an email field. Do you think it is fine to leave that out?

JordanSussman: Do you have an issue that describes what this is trying to solve? I've confirmed that tag events within Vela 0.9 have this data available within the environment variables of `VELA_BUILD_SENDER`, `VELA_BUILD_AUTHOR`, and `VELA_BUILD_AUTHOR_EMAIL`.

ecrupper: [https://github.com/go-vela/community/issues/436](url) Sorry it was an internal issue that wasn't duplicated yet. So since the `VELA_BUILD_SENDER` already exists, the only missing information is the email, which as I mentioned does not come with the payload. Any ideas? Should we just inform the user to use the `VELA_BUILD_SENDER` for the name and scrap the email? The `BUILD_AUTHOR_EMAIL` will not work since it is tied to the last person to commit. Not sure if it's a good idea to set the `BUILD_AUTHOR` and `BUILD_AUTHOR_EMAIL` to the tagger on tag events, but that is an option. Let me know!

ecrupper:
Closing this PR in favor of using the `sender` field in the GitHub webhook instead of `pusher`. No library/database changes necessary - just added environment variable in go-vela/types

PR #534, opened 2021-11-15
enhance: setting version information

jbrockopp:
This change slightly modifies two pieces from the version output.

This first part is updating the version output to use a fallback value for the semantic tag of `v0.0.0`.

This should prevent the possibility of a panic happening if the tag info isn't captured properly:

https://github.com/go-vela/server/blob/aff96b30eefd6d9425758088f97e42be1fd80243/Makefile#L17-L21

The second part is using Go's `runtime.Version()` function to capture the version of `go` the app is built with:

https://pkg.go.dev/runtime#Version

Since we're relying on that function from the `runtime` library, we no longer need to capture that info in the `Makefile`.

Previously, we used to run the `go version` command and parse the value from the output:

https://github.com/go-vela/server/blob/aff96b30eefd6d9425758088f97e42be1fd80243/Makefile#L23-L27

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,14 +20,8 @@ ifndef GITHUB_TAG
 	GITHUB_TAG = $(shell git describe --tag --abbrev=0)
 endif
 
-# check if a go version is already set
-ifndef GOLANG_VERSION
-	# capture the current go version we build the application from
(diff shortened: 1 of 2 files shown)


kneal (approved):
LGTM 🐬

KellyMerrick (approved):
lgtm

PR #536, opened 2021-11-18
refactor: rename the source package to scm

kneal:
PR Includes the following changes:

- Rename the `source` package to `scm`
- Update source to scm messaging

**This PR has changes to the environment variables used to configure the server.** Its the same spirit as https://github.com/go-vela/community/issues/395 and https://github.com/go-vela/community/issues/394

The goal is to just improve clarity for new users by using a more industry identifiable term. 

_This is a prelude for me starting: https://github.com/go-vela/community/issues/441_


JordanSussman:
> This PR has changes to the environment variables used to configure the server. Its the same spirit as go-vela/community#395 and go-vela/community#394

Can the old environment variables be left in the urfave slice to avoid the breaking change?

kneal:
@JordanSussman Yeah, I can add them back with a `TODO` to remove for a future release

scm/github/github.go
```diff
@@ -53,7 +53,7 @@ type client struct {
 	AuthReq *github.AuthorizationRequest
 }
 
-// New returns a Source implementation that integrates with
+// New returns a SCM implementation that integrates with
```
wass3r, line 56: super nitpicky, but this is the first instance i found where "SCM" is actually all caps so it jumped out at me for being inconsistent 🤷🏼‍♂️

wass3r (approved):

scm/github/github.go
kneal: Umm... it should be in all instances `S` was capitalized for source.

kneal: For example here it is in another file:
https://github.com/go-vela/server/pull/536/files#diff-20b08cad4aeb414832e6f834e33e07c46c1be25d5b3fc3d55d602a8fa0d925a5L23


Can definitely modify it so there is no all caps tho, if that is preferred.

wass3r: like i said, nitpicky.. i approved the PR so do what you will 🤪

cognifloyd: Gramatically, I prefer all caps here, but if we're going that far it would be:
```suggestion
// New returns an SCM implementation that integrates with
```
:zany_face:

KellyMerrick: +1 for @cognifloyd

jbrockopp (approved):
LGTM

PR #542, opened 2021-11-26
feat(scm): add new native scm driver

kneal:
**PR includes the following changes:**

- add a Bitbucket option for using the native driver
- add a native driver implentation based on the [jenkins-x/go-scm](https://pkg.go.dev/github.com/jenkins-x/go-scm) library

_closes https://github.com/go-vela/community/issues/441_

**Notes:**

- The SCM package has been functionally tested with a set of pipelines against GitHub and Bitbucket. 
- I will not be adding docs yet because I do not recommend using the package until further testing is completed
- The new library has a lot of testing downsides I would like to re-think in next pr. (Documented with TODOs)


cognifloyd:
Why "native"?
Other possibilities: "generic", "abstract", "shim"

kneal:
I was originally thinking _generic_ for the package name but we actually already use the _native_ nomenclature in a few package interfaces. The thought was _native_ might sync up with that thought better and do like the standard OAuth esque integrations and then other packages could focus on being more 1st class citizens to the individual SCMs if we want that. 

Definitely not stuck on the name by any means, I'm largely indifferent on what we call it. I just thought in the moment it made sense when I was first structuring out the package/files.

scm/native/access.go
```diff
@@ -0,0 +1,164 @@
+// Copyright (c) 2021 Target Brands, Inc. All rights reserved.
```
jbrockopp, line 1: Could we update the copyright statement to use the current year? 👍 

I think this change will need to be applied throughout the files being added here.

scm/native/access.go
```diff
@@ -0,0 +1,164 @@
...
+
+	"github.com/go-vela/types/library"
+	"github.com/jenkins-x/go-scm/scm"
+
+	"github.com/sirupsen/logrus"
+)
+
+// OrgAccess captures the user's access level for an org.
+func (c *client) OrgAccess(u *library.User, org string) (string, error) {
+	logrus.Tracef("Capturing %s access level to org %s", u.GetName(), org)
```
jbrockopp, line 19: Could we update the logging statements to reflect current state in the codebase? 👍 

i.e.

https://github.com/go-vela/server/blob/3fe3d7eaabc3f1ee03d3a7636749bdc8009ff0ee/scm/github/access.go#L16-L21

I think this change will need to be applied throughout the functions and files being added here.

jbrockopp (changes requested):

scm/native/access.go
kneal: Not sure what you mean by "current state" in this context? Could you elaborate?

kneal: Ohh, I'm sorry I see it now. You mean using the client logger. Sorry, for some reason my eyes weren't picking up the difference. Yeah, super easy change!

kneal: Better to get it out of the way now tho. So good call out!

PR #545, opened 2021-11-30
feat: add index on builds.created

dtanner:
Used for reporting - e.g. `where created > $some_epoch` to quickly and cheaply limit querying recent build metrics.

### DIFF
--- a/database/postgres/ddl/build.go
+++ b/database/postgres/ddl/build.go
@@ -62,4 +62,13 @@ IF NOT EXISTS
 builds_status
 ON builds (status);
 `
+
+	// CreateBuildCreatedIndex represents a query to create an
+	// index on the builds table for the created column.
+	CreateBuildCreatedIndex = `
+CREATE INDEX CONCURRENTLY
+IF NOT EXISTS
+builds_created
+ON builds (created);
+`
 )
--- a/database/postgres/postgres.go
+++ b/database/postgres/postgres.go
@@ -257,6 +257,12 @@ func createIndexes(c *client) error {
 		return fmt.Errorf("unable to create builds_status index for the %s table: %v", constants.TableBuild, err)
 	}
 
+	// create the builds_created index for the builds table
+	err = c.Postgres.Exec(ddl.CreateBuildCreatedIndex).Error
+	if err != nil {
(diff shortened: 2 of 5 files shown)


database/postgres/ddl/build.go
```diff
@@ -61,5 +61,11 @@ CREATE INDEX
 IF NOT EXISTS
 builds_status
 ON builds (status);
+`
```
wass3r, line 64: super picky but for consistency, can we add blank line after this and add a comment before the new query?

database/postgres/postgres.go
```diff
@@ -305,5 +305,10 @@ func createIndexes(c *client) error {
 		return fmt.Errorf("unable to create workers_hostname_address index for the %s table: %v", constants.TableWorker, err)
 	}
 
+	err = c.Postgres.Exec(ddl.CreateBuildCreatedIndex).Error
```
wass3r, line 308: missing comment before this for consistency

database/sqlite/ddl/build.go
```diff
@@ -61,5 +61,12 @@ CREATE INDEX
 IF NOT EXISTS
 builds_status
 ON builds (status);
+`
+
+	CreateBuildCreatedIndex = `
```
wass3r, line 66: missing comment right before this new query.

database/sqlite/sqlite.go
```diff
@@ -256,6 +256,11 @@ func createIndexes(c *client) error {
 		return fmt.Errorf("unable to create builds_status index for the %s table: %v", constants.TableBuild, err)
 	}
 
+	err = c.Sqlite.Exec(ddl.CreateBuildCreatedIndex).Error
```
wass3r, line 259: missing comment before index creation.

additionally, i noticed between here and the postgres addition that the index is inserted in different spots. in the postgres version it was inserted at the end of the indexes and here after the other index on the build table. if i had to pick, i like the approach here better, but feel like it should be consistent between the two no matter where it happens.

wass3r (changes requested):
thanks for the additions! left some comments. super minor tweaks - mostly for consistency.

dtanner:
i think all your comments were addressed @wass3r

wass3r (approved):

jbrockopp (approved):
LGTM

PR #550, opened 2021-12-06
chore: put sync and sync all in a new scm folder

ecrupper:
For consistency, I think putting the sync repo functions in their own scm folder. Since the base of the API call starts with `scm/`, it would match the current practice of repo, build, etc.


wass3r (approved):

kneal (approved):
LGTM 🐬

jbrockopp (approved):
LGTM

PR #552, opened 2021-12-08
feat(stage): add environment map

ecrupper:
[Related Issue](https://github.com/go-vela/community/issues/420)
[Related Types PR](https://github.com/go-vela/types/pull/215)

These changes allow for users to configure an environment for a specific stage. The stage environment inherits the environment set globally and passes that environment (along with any values set in stage) to the steps. It will also pass environment variables to templates.


JordanSussman:
Can you confirm that setting an environment at the stage level does not override a platform injected variable like `VELA_BUILD_AUTHOR`? Example yaml file I'm concerned about:

```yaml
stages:
  test:
    environment:
        VELA_BUILD_AUTHOR: fakeuser
    steps:
      - name: test
        commands:
          - echo $VELA_BUILD_AUTHOR
        image: alpine
```

The response of the `test` step should not return `fakeuser`.

ecrupper:
> Can you confirm that setting an environment at the stage level does not override a platform injected variable like `VELA_BUILD_AUTHOR`? Example yaml file I'm concerned about:
> 
> ```yaml
> stages:
>   test:
>     environment:
>         VELA_BUILD_AUTHOR: fakeuser
>     steps:
>       - name: test
>         commands:
>           - echo $VELA_BUILD_AUTHOR
>         image: alpine
> ```
> 
> The response of the `test` step should not return `fakeuser`.

Just verified this! It indeed maintains environment variables that are not meant to be overwritten by a user

compiler/native/environment.go
```diff
@@ -16,31 +16,69 @@ import (
 	"github.com/go-vela/types/yaml"
 )
 
-// EnvironmentStages injects environment variables
-// for each step in every stage in a yaml configuration.
+// EnvironmentSteps injects environment variables
```
KellyMerrick, line 19: The function for this comment looks to still be named `EnvironmentStages`.

compiler/native/environment.go
```diff
@@ -16,31 +16,69 @@ import (
...
 		if err != nil {
 			return nil, err
 		}
+	}
 
-		stage.Steps = steps
+	return s, nil
+}
+
+// EnvironmentStages injects environment variables
```
KellyMerrick, line 34: The function name for this comment doesn't match.

KellyMerrick (changes requested):

compiler/native/environment.go
```diff
@@ -17,30 +17,68 @@ import (
...
+		env[k] = v
+	}
+
+	// inject the default environment
+	// variables to the build stage
+	// we do this after injecting the declared environment
+	// to ensure the default env overrides any conflicts
+	for k, v := range defaultEnv {
+		env[k] = v
 	}
```
ecrupper, line 59: I'm curious as to whether or not this is necessary.  Since the stage will never have a commands block, I don't see a possible scenario where a user overwrites a default environment. However, it might be nice to have this so that if we were to ever introduce some sort of functionality with environment variables, whoever implemented that change would not have to worry about changing this. Would love to hear more thoughts!

wass3r: i'm good with leaving it in as a precautionary measure.

compiler/native/environment.go
```diff
@@ -17,30 +17,68 @@ import (
...
+	env := make(map[string]string)
+	// gather set of default environment variables
+	defaultEnv := environment(c.build, c.metadata, c.repo, c.user)
+
+	// inject the declared global environment
+	// WARNING: local env can override global
+	env = appendMap(env, globalEnv)
+
+	// inject the declared environment
+	// variables to the build step
```
wass3r, line 48:
```suggestion
	// variables to the build stage
```

wass3r (commented):
one super minor edit

KellyMerrick (approved):
lgtm

kneal (approved):
LGTM 🐬

PR #558, opened 2021-12-20
feat(repo)!: add build rate limiting

jbrockopp:
Dependent on https://github.com/go-vela/types/pull/219

Part of the effort for https://github.com/go-vela/community/issues/27 and https://github.com/go-vela/community/issues/162

Based off of https://github.com/go-vela/community/blob/master/proposals/2020/12-16_rate-limiting.md

Related to https://github.com/go-vela/server/tree/matt/rate-limiting

## New Field

This adds support for the `BuildLimit` repo field introduced in the above [go-vela/types](https://github.com/go-vela/types) PR.

This includes creating a new repo with the field set:

https://github.com/go-vela/server/blob/5958b154e95a167f67567e08c1d01c19d0e00fe7/api/repo.go#L126-L135

As well as, updating an existing repo with the field set:

https://github.com/go-vela/server/blob/5958b154e95a167f67567e08c1d01c19d0e00fe7/api/repo.go#L669-L680

> NOTE:
>
> This logic can be customized with a new `VELA_DEFAULT_BUILD_LIMIT` flag for the server:
>
> https://github.com/go-vela/server/blob/5958b154e95a167f67567e08c1d01c19d0e00fe7/cmd/vela-server/main.go#L98-L103
>
> And additionally with a new `VELA_MAX_BUILD_LIMIT` flag for the server:
>
> https://github.com/go-vela/server/blob/5958b154e95a167f67567e08c1d01c19d0e00fe7/cmd/vela-server/main.go#L104-L109

## Rate Limiting

Additionally, this adds a basic form of rate limiting to the following endpoints:

* `POST /webhook` - used for standard webhook workflow
* `POST /api/v1/repos/:org/:repo/builds` - API endpoint for creating a new build for a repository
* `POST /api/v1/repos/:org/:repo/builds/:build` - API endpoint for restarting an existing build for a repository

To accomplish this, we now send a query to the database to count the `pending` and `running` builds for a repo:

https://github.com/go-vela/server/blob/5958b154e95a167f67567e08c1d01c19d0e00fe7/api/webhook.go#L275-L290

After we query those results, we check to see if the count of builds exceeds the `BuildLimit` set for the repo:

https://github.com/go-vela/server/blob/5958b154e95a167f67567e08c1d01c19d0e00fe7/api/webhook.go#L292-L302

In the event the database query fails or the count exceeds the limit, an error is returned:

```
unable to process webhook: repo foo/bar has exceeded the concurrent build limit of 10
```

> NOTE:
>
> I had to modify the existing `GetRepoBuildCount()` database function to accept a `map[string]interface{}`:
>
> https://github.com/go-vela/server/blob/5958b154e95a167f67567e08c1d01c19d0e00fe7/database/service.go#L49-L51
>
> Previously, this function accepted a `map[string]string` which didn't work for this use case.
>
> I also noticed an used `GetRepoBuildCountByEvent()` database function so I removed it:
>
> https://github.com/go-vela/server/blob/c6bebb55a4e08695209e090c89dceda5ece9c3f9/database/service.go#L55-L57


kneal (commented):
Honestly, [10 as the default](https://github.com/go-vela/types/pull/219/files#diff-3866a32a7c1f41b5da432ca3634351fbe40e59447776a231a17a666c4160d8e5R16) does feel pretty small to me for repos that have large amounts of users or mono repos but it is adjustable. So I guess, we'll find out. The rule of thumb I always hear in agile training for team size is two large pizzas. So, like 6-10 people an agile team and if in cases all those people are in a single repo that does PR/push builds I could see teams hitting that minimum quite frequently. All of that is of course speculation so take the information how you would like but IMO it's on the smaller side. I'd be in favor of leaving it until we see community concern. 

On a different note, how do you see this working with a multi-worker build? I generally see this feature as a way to limit a single repo from taking to many worker resources but a multi-worker build (when it does exist) would be a single work across multiple workers. So, I'd imagine you'd want to weight that somehow. 

I could also see people wanting an easier way to replay the hook event. The way this is implemented would not add a build to the builds table so there's no easy way from the Vela UI to replay the build. The user would need to create a new event or replay from the VCS system which is some cumbersome.

jbrockopp:
@kneal 

> Honestly, [10 as the default](https://github.com/go-vela/types/pull/219/files#diff-3866a32a7c1f41b5da432ca3634351fbe40e59447776a231a17a666c4160d8e5R16) does feel pretty small to me for repos that have large amounts of users or mono repos but it is adjustable.

Perhaps this discussion is worth having over in `types`? 😃 

https://github.com/go-vela/types/pull/219#pullrequestreview-837695253

The limit is probably small for repos with a large amount of users, which often tend to be mono repos.

However, like you mentioned, the limit is adjustable like the build timeout functionality so hopefully it isn't a problem.

FWIW:

It's worth mentioning that, to some extent, this is intended behavior for this functionality 👍 

Rate limiting builds will prevent a user from accidentally or maliciously DDOSing Vela with a surge of new builds.

Additionally, like you mentioned, it can also be used to prevent specific repos from monopolizing all the resources.

To be clear, this means preventing one repo from controlling an unfair amount of workers with running builds.

From what I've seen before, the repos that tend to do this are often the ones you mentioned above 😅 

> On a different note, how do you see this working with a multi-worker build?

I think it might be too soon to tell because it would depend on how the multi-worker functionality is written 🤷 

Once that feature is written (or maybe before), there will likely need to be some changes to improve this functionality.

A great enhancement would be to push builds exceeding the limit to a low priority route instead of failing to create it 👍 

> I could also see people wanting an easier way to replay the hook event.

Agreed 😄 

Adding something like https://github.com/go-vela/community/issues/459 should address this concern

kneal:
> It's worth mentioning that, to some extent, this is intended behavior for this functionality 👍
> 
> Rate limiting builds will prevent a user from accidentally or maliciously DDOSing Vela with a surge of new builds.
> 
> Additionally, like you mentioned, it can be used to prevent specific repos from monopolizing all the resources.
> 
> To be clear, this means preventing one repo from controlling an unfair amount of workers with running builds.

I do understand the point of the feature is preventing accidentally or maliciously DDOSing Vela. I feel like that is fairly obvious and didn't need to be stated. I'm was trying to call out that I don't think people using those patterns should feel penalized for workflows that make them the most productive for their applications but I understand gotta be a balance that doesn't affect the Vela server. 

I do think the level of customization should be enough with maybe a minor tweak but per your comment, I can move that conversation to the other PR.

jbrockopp:
> I'm was trying to call out that I don't think people using those patterns should feel penalized for workflows that make them the most productive for their applications but I understand gotta be a balance that doesn't affect the Vela server.

Yeah that's a tough scenario to account for confidently 😞 

Hopefully, this won't be the case since we allow users to customize the limit, to some extent, without admin involvement.

Ultimately, this is likely going to be one of those features where community feedback will help us greatly!

api/repo.go
```diff
@@ -77,6 +77,7 @@ func CreateRepo(c *gin.Context) {
 	// capture middleware values
 	u := user.Retrieve(c)
 	allowlist := c.Value("allowlist").([]string)
+	defaultLimit := c.Value("defaultLimit").(int64)
```
JordanSussman, line 80: Should this be updated to `defaultBuildLimit` instead? I know `defaultTimeout` is present elsewhere, but adding the additional context to additions feels like a good change.

jbrockopp: Yeah I was trying to be consistent with `defaultTimeout` because I have an idea for refactor effort

That being said, I agree with you so I'll make this change now 😃

jbrockopp: This should be addressed 👍

kneal (approved):
LGTM 🐬

kneal (approved):
LGTM 🐬

api/repo.go
```diff
@@ -650,6 +670,35 @@ func UpdateRepo(c *gin.Context) {
...
+				util.MaxInt(
+					constants.BuildLimitMin,
+					util.MinInt(
+						int(input.GetBuildLimit()),
+						int(maxBuildLimit),
+					), // clamp max
+				), // clamp min
+			),
+		)
+	}
```
JordanSussman, line 700: This seems redundant since `maxBuildLimit` defaults to `constants.BuildLimitMax`

jbrockopp: Yeah I can definitely see where you're coming from on that 😅 

In most cases, the code above will be used since `maxBuildLimit` **should** contain a value above `0`.

However, it's possible for someone to accidentally configure `maxBuildLimit` with a value of `0`.

In this scenario, the above code wouldn't executed and the following would run instead:

https://github.com/go-vela/server/blob/3c1b13d8471a7f460d6cd6ecba259b8d18df7668/api/repo.go#L674-L687

So, you could view the redundant code as providing an implicit fallback for the build limit on the server?

The only reason I can come up with to keep it has to do with misconfiguration on the server.

If we're thinking we don't want the extra code to handle that scenario, then I can remove it?

JordanSussman: Feels like we should add some validation logic for the `maxBuildLimit` value to account for that scenario.

jbrockopp: Sure 👍  I'll add that validation logic now.

FWIW:

We may have to remove this validation logic in the future if the use case comes up.

It's doubtful, but at some point a user running Vela may want to set no max build limit for repos.

A pattern I've seen in the Go community (and outside of it) is `0` == unlimited when setting values like this.

jbrockopp: This should be addressed 👍

JordanSussman (approved):

wass3r (commented):
Thanks for this. Can we either add a story for documentation or even better add a draft PR. Would probably live here: https://go-vela.github.io/docs/administration/server/reference/?

jbrockopp:
> Thanks for this. Can we either add a story for documentation or even better add a draft PR. Would probably live here: https://go-vela.github.io/docs/administration/server/reference/?

@wass3r Documentation will be added as a part of the original stories:

* https://github.com/go-vela/community/issues/27
* https://github.com/go-vela/community/issues/162

The draft PR has been opened:

https://github.com/go-vela/docs/pull/270

However, it's incomplete as I'm waiting for the rest of this functionality to be accepted:

* https://github.com/go-vela/ui/pull/494
* https://github.com/go-vela/cli/pull/317

wass3r (approved):

PR #560, opened 2021-12-22
fix(api/sync): add favorites handling to sync repos

ecrupper:
I noticed that the current sync repo API doesn't update the user's favorites list. Quick fix to handle that so the user doesn't have to unfavorite all deleted repos.


api/scm.go
```diff
@@ -107,6 +107,11 @@ func SyncRepos(c *gin.Context) {
 
 				return
 			}
+			e = updateUserFavorites(c, repo)
+			if e != nil {
```
jbrockopp, line 111: Can you update `e` -> `err` here?

```suggestion
			err = updateUserFavorites(c, repo)
			if err != nil {
```

In Go, it is commonplace and idiomatic to use `err` as a variable name to represent an `error`.

api/scm.go
```diff
@@ -172,6 +177,42 @@ func SyncRepo(c *gin.Context) {
 
 			return
 		}
+		e = updateUserFavorites(c, r)
+		if e != nil {
```
jbrockopp, line 181: Can you update `e` -> `err` here?

api/scm.go
```diff
@@ -172,6 +177,42 @@ func SyncRepo(c *gin.Context) {
...
 	c.JSON(http.StatusOK, fmt.Sprintf("repo %s synced", r.GetFullName()))
 }
+
+// updateUserFavorites is a helper function that takes a repo
+// that has been deleted from the SCM and ensures that it does
+// not appear in the user's favorites after its deletion.
+func updateUserFavorites(c *gin.Context, r *library.Repo) error {
+	// get user from the database
+	u, e := database.FromContext(c).GetUser(r.GetUserID())
+	if e != nil {
```
jbrockopp, line 195: Can you update `e` -> `err` here?

jbrockopp (changes requested):

api/scm.go
```diff
@@ -99,14 +99,19 @@ func SyncRepos(c *gin.Context) {
...
-			if e != nil {
+			err := database.FromContext(c).UpdateRepo(repo)
+			if err != nil {
 				retErr := fmt.Errorf("unable to update repo for org %s: %w", org, err)
 
 				util.HandleError(c, http.StatusInternalServerError, retErr)
 
 				return
 			}
+			err = updateUserFavorites(c, repo)
```
jbrockopp, line 110: I think we're missing a space here:

```suggestion
			}

			err = updateUserFavorites(c, repo)
```

api/scm.go
```diff
@@ -164,14 +169,50 @@ func SyncRepo(c *gin.Context) {
...
-		if e != nil {
+		err := database.FromContext(c).UpdateRepo(r)
+		if err != nil {
 			retErr := fmt.Errorf("unable to update repo for org %s: %w", org, err)
 
 			util.HandleError(c, http.StatusInternalServerError, retErr)
 
 			return
 		}
+		err = updateUserFavorites(c, r)
```
jbrockopp, line 180: I think we're missing a space here:

```suggestion
		}

		err = updateUserFavorites(c, r)
```

api/scm.go
```diff
@@ -164,14 +169,50 @@ func SyncRepo(c *gin.Context) {
...
+
+// updateUserFavorites is a helper function that takes a repo
+// that has been deleted from the SCM and ensures that it does
+// not appear in the user's favorites after its deletion.
+func updateUserFavorites(c *gin.Context, r *library.Repo) error {
+	// get user from the database
+	u, err := database.FromContext(c).GetUser(r.GetUserID())
+	if err != nil {
+		retErr := fmt.Errorf("%s: failed to retrieve user in database", baseErr)
+		util.HandleError(c, http.StatusBadRequest, retErr)
```
jbrockopp, line 197: AFAICT, if this function returns an `error`, then we call `util.HandleError()` with that error:

https://github.com/go-vela/server/blob/20f7ffd20dcbb07f207227ae46a401b3a84e6176/api/scm.go#L180-L184

AFAIK, that function will "abort" the current request preventing any subsequent calls from proceeding:

https://github.com/go-vela/server/blob/master/util/util.go#L18

https://pkg.go.dev/github.com/gin-gonic/gin#Context.AbortWithStatusJSON

I mention all of this, because I think we can remove this line.

api/scm.go
```diff
@@ -164,14 +169,50 @@ func SyncRepo(c *gin.Context) {
...
+			favorites = append(favorites, fav)
+		}
+	}
+	u.SetFavorites(favorites)
+
+	// update the user favorites slice
+	err = database.FromContext(c).UpdateUser(u)
+	if err != nil {
+		retErr := fmt.Errorf("%s: failed to update repo %s: %v", baseErr, r.GetFullName(), err)
+		util.HandleError(c, http.StatusBadRequest, retErr)
```
jbrockopp, line 214: Like above, remove this line because `util.HandleError()` is called if `updateUserFavorites()` returns an `error`.

api/scm.go
```diff
@@ -164,14 +169,50 @@ func SyncRepo(c *gin.Context) {
...
+		return retErr
+	}
+	return nil
```
jbrockopp, line 217: I think we're missing a space here:

```suggestion
	}
	
	return nil
```

jbrockopp (changes requested):

api/scm.go
```diff
@@ -191,7 +197,43 @@ func SyncRepo(c *gin.Context) {
...
+
+// updateUserFavorites is a helper function that takes a repo
+// that has been deleted from the SCM and ensures that it does
+// not appear in the user's favorites after its deletion.
+func updateUserFavorites(c *gin.Context, r *library.Repo) error {
+	// get user from the database
+	u, err := database.FromContext(c).GetUser(r.GetUserID())
+	if err != nil {
+		retErr := fmt.Errorf("%s: failed to retrieve user in database", baseErr)
+		return retErr
```
jbrockopp, line 219: Thoughts on condensing this into one line?

```suggestion
		return fmt.Errorf("%s: failed to retrieve user in database", baseErr)
```

api/scm.go
```diff
@@ -191,7 +197,43 @@ func SyncRepo(c *gin.Context) {
...
+			favorites = append(favorites, fav)
+		}
+	}
+	u.SetFavorites(favorites)
+
+	// update the user favorites slice
+	err = database.FromContext(c).UpdateUser(u)
+	if err != nil {
+		retErr := fmt.Errorf("%s: failed to update repo %s: %v", baseErr, r.GetFullName(), err)
+		return retErr
```
jbrockopp, line 235: Thoughts on condensing this into one line?

```suggestion
		return fmt.Errorf("%s: failed to update repo %s: %v", baseErr, r.GetFullName(), err)
```

jbrockopp (changes requested):

api/scm.go
```diff
@@ -191,7 +197,41 @@ func SyncRepo(c *gin.Context) {
...
+		}
 	}
 
 	c.JSON(http.StatusOK, fmt.Sprintf("repo %s synced", r.GetFullName()))
 }
+
+// updateUserFavorites is a helper function that takes a repo
+// that has been deleted from the SCM and ensures that it does
+// not appear in the user's favorites after its deletion.
+func updateUserFavorites(c *gin.Context, r *library.Repo) error {
```
jbrockopp, line 214: @ecrupper the change looks good to me so I'll approve 👍 

One point of clarification I wanted to inquire about for my own understanding.

This change doesn't update `Favorites` for all users in Vela, right?

AFAICT, it appears it would update `Favorites` for the owner of the repo in Vela.

So any users who had the repo in their `Favorites`, but wasn't the owner of that repo, would still see it?

jbrockopp (approved):
LGTM

dtanner (approved):

ecrupper:
Closing this PR. While it's nice to have the user who ran `sync repos` have an updated favorites, it is confusing that this is not reflected across all users who have favorited the orphaned repos.

PR #562, opened 2021-12-30
fix(enqueue): add db updates for enqueue events

wass3r:
fixes https://github.com/go-vela/community/issues/458

mini PR that will address the following scenarios (as touched on in the issue):

1. queue is down (or temporarily unavailable); the build will now error and be updated in the DB instead of being perpetually pending
2. there is an issue adding the build to the queue; the build will now error and be updated in the DB instead of being perpetually pending
3. there is a build up in the queue, waiting for a worker to pick up the build; the `enqueued` field will now be updated in the DB **before** a worker picks it up. prior to this change, the field/build wasn't updated in the DB until after a worker picked it up.

to note, neither the CLI nor the UI do particularly well with handling duration in cases where a build never starts.

image
image

if i'm not mistaken the change introduced in https://github.com/go-vela/types/pull/222 will handle the CLI issue once it's in use there. the UI will need a fix however.


kneal (approved):
LGTM 🐬

jbrockopp (approved):
LGTM

PR #566, opened 2022-01-07
initial work for signature api and build checking in webhook

ecrupper:
This is very much a draft but I was hoping it could provide some more detail on how the pipeline signing might work server-side. For example purposes, I made `Trusted` the field in question, put the signature in a separate file `.vela.sig`, and use the entire contents of the `.vela.yml` file as the key for the signature.

A few things I'm missing: expanded pipeline since templates could get manipulated, SHA checksum of `.vela.yml` file, and obviously the accompanying CLI changes. I mainly wanted to put this out here as a draft as a supplement to the [proposal](https://github.com/go-vela/community/pull/461)

### DIFF
--- a/api/build.go
+++ b/api/build.go
@@ -213,7 +213,7 @@ func CreateBuild(c *gin.Context) {
 	}
 
 	// send API call to capture the pipeline configuration file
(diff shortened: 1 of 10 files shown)


ecrupper:
Closing

PR #567, opened 2022-01-07
initial work for acp solution

ecrupper:
This is very much a draft but I was hoping it could provide some more detail on how the ACP idea might work server-side. For example purposes, I made the access levels `write` to run the build, `admin` to approve of a build with a comment. I also made the trigger comment the ✅ emoji just for now. 

Hopefully these changes can potentially act as a supplement to the [proposal](https://github.com/go-vela/community/pull/461) for added pipeline security. I'm sure there are some details I'm missing in this implementation, but hopefully it makes a clearer picture of what it would look like.


ecrupper:
Closing

PR #568, opened 2022-01-10
enhance: allow user to query build via commit hash

srinunagulapalli:
For implementing [124](https://github.com/go-vela/community/issues/124).
Added request parameter `commit` that allows the user to query builds based on the commit hash. 
A `GET` request on `/api/v1/repos/:org/:repo/builds?commit=40charactersOfTheCommitHash` returns the list of builds for that particular commit.

### DIFF
--- a/api/build.go
+++ b/api/build.go
@@ -417,6 +417,8 @@ func GetBuilds(c *gin.Context) {
 	event := c.Query("event")
 	// capture the status type parameter
 	status := c.Query("status")
+	// capture the commit hash parameter
+	commit := c.Query("commit")
 
 	// check if branch filter was provided
 	if len(branch) > 0 {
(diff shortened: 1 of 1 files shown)


jbrockopp (dismissed):
LGTM

dtanner (dismissed):

wass3r (changes requested):
thank you for the addition. can you please add the new parameter to the swagger annotation that starts here: https://github.com/go-vela/server/blob/0e82ed5f14c0272d64c8afbe272365b6dd9f338c/api/build.go#L357

also, i see that `branch` and `status` are missing from the annotation as well. if you don't mind add those, that would be great - otherwise, we can add in after this PR.

srinunagulapalli:
@wass3r Added the parameters `commit`, `branch`, `status` to the swagger annotation.

wass3r (approved):
Thank you!

jbrockopp (approved):
LGTM

PR #574, opened 2022-01-19
feat(database): add support for pipelines

jbrockopp:
Dependent on https://github.com/go-vela/types/pull/223

Part of the effort for https://github.com/go-vela/community/issues/460

> NOTE: I had to modify `hooks` to account for https://github.com/go-vela/types/pull/239

This adds support for a new `pipelines` resource to the `github.com/go-vela/server/database` package.

As a part of this, I used a new strategy to add this logic which attempts to solve a few things:

* reduce code complexity
* reduce code duplication
* improve ability to add support for other database drivers
* provider a cleaner abstraction for database resources

If this new approach is liked and folks feel the pros outweigh the cons, I'll replicate this for the other resources.

If not, then I'll modify this PR accordingly to follow our existing pattern.

## Structure

One distinguishable change with this approach is all code for `pipelines` is under a `database/pipeline` package:

```shell
$ tree -d database
database
├── pipeline
├── postgres
│   ├── ddl
│   └── dml
└── sqlite
    ├── ddl
    └── dml

7 directories
```

You'll notice that no `dml` package is nested under the `pipeline` package.

This new approach would no longer require raw SQL queries for integrating with the `pipelines` table.

This is great because we reduce our total LOC and complexity when we need to add/update functionality.

To be specific, when we want to add new code for `pipelines`, ideally I should only have to update the one package.

We no longer have to update code specific to `postgres`, `sqlite` or any future drivers we may support for Vela.

However, it comes with a cost of becoming **reliant on the GORM library** which can be a downside.

## Resource Specific Interface

Inside the package exists a `PipelineService` interface that declares all functions necessary for this code:

https://github.com/go-vela/server/blob/093f2e1e46438b2ffdecb32f0eb0991b542c1d4e/database/pipeline/service.go#L11-L49

Most of the naming for these functions was [adopted from the existing code](https://github.com/go-vela/server/blob/master/database/service.go#L68-L93) barring the `Get` prefix.

And then we create an `engine` which implements that service interface:

https://github.com/go-vela/server/blob/093f2e1e46438b2ffdecb32f0eb0991b542c1d4e/database/pipeline/pipeline.go#L25-L39


database/pipeline/table.go
```diff
@@ -0,0 +1,76 @@
...
+
+const (
+	// CreatePostgresTable represents a query to create the Postgres pipelines table.
+	CreatePostgresTable = `
+CREATE TABLE
+IF NOT EXISTS
+pipelines (
+	id               SERIAL PRIMARY KEY,
+	repo_id          INTEGER,
+	number           INTEGER,
```
kneal, line 17: Is this the build number?

kneal (commented):
Overall it looks fantastic, one question what is the goal of `SkipCreation` for the pipeline endpoints?

jbrockopp:
> Overall it looks fantastic, one question what is the goal of `SkipCreation` for the pipeline endpoints?

To answer your question directly, the `SkipCreation` logic won't be used for the API pipeline endpoints 👍 

In case others have the same or similar questions, I'll provide some context on `SkipCreation` in general 😃 

Then I'll explain a bit more on why we're referencing it in the `pipeline` package.

This functionality was first added here: https://github.com/go-vela/server/pull/455

By default, Vela will always attempt to create all tables and indexes in the database 👍 

However, we enable Vela installation admins to disable this behavior by setting this flag to `true` 😄 

The reason why we're using `SkipCreation` in the `pipeline` package is because of how the new code is structured.

When we create the pipeline `engine` which implements the `PipelineService`, we create the table and indexes too:

https://github.com/go-vela/server/blob/093f2e1e46438b2ffdecb32f0eb0991b542c1d4e/database/pipeline/pipeline.go#L62-L79

So, when we create the database client for `postgres` or `sqlite`, we need to pass this flag to that engine:

https://github.com/go-vela/server/blob/093f2e1e46438b2ffdecb32f0eb0991b542c1d4e/database/postgres/postgres.go#L94-L98

->

https://github.com/go-vela/server/blob/093f2e1e46438b2ffdecb32f0eb0991b542c1d4e/database/postgres/postgres.go#L358-L369

database/pipeline/table.go
jbrockopp: No, but it follows the same train of logic 👍 

Like `builds` today, each `pipeline` in the table will have a `number` field that increments based off the `repo_id`.

That field doesn't directly map to the build number because pipelines have a one-to-many relation with builds 😃 

You can have multiple builds all stemming from the same pipeline i.e. a build that gets restarted

ecrupper (approved):
I really like the design change. + 1 for getting it applied to the rest of the resources. 

Everything else LGTM.

JordanSussman (approved):

kneal (approved):
LGTM 🐬

PR #578, opened 2022-01-27
fix(secrets): actually set updated_at and updated_by fields

ecrupper:
Actually setting the metadata fields when a secret is updated. I also made the existing test actually test for an update so this hopefully won't happen in the future.


wass3r (approved):

jbrockopp (approved):
LGTM

FWIW:

In the future, we should consider removing this code in the `secret/native/update.go` file 😅 

It feels like we're already handling this logic in the API handler:

https://github.com/go-vela/server/blob/master/api/secret.go#L546-L547

PR #586, opened 2022-02-15
enhance(api/build): adding the ability to search builds within time constraints

ecrupper:
Adding `before` and `after` query parameters which take in unix epoch timestamps and fetches builds within those time constraints (if provided). Closes [issue 187](https://github.com/go-vela/community/issues/187).


kneal:
Saw [comment about offline conversation](https://github.com/go-vela/community/issues/187#issuecomment-765699475), however, would it make more sense to just expose the created, started, finished, etc in query params, then the user can just pick the ranges for the specific timestamp they are interested in.

`before` and `after` is easy to read but it is somewhat limiting but again I'm not sure what was discussed in that offline conversation.

ecrupper:
I'm not sure what the offline conversation was either since it was a year ago, but I chose `started` because it seemed the most reliable of the timestamps. If we exposed all the timestamp fields, we would still need a way to utilize some sort of `before` and `after` capability I would imagine. I'm not sure how gin handles ranges. I did a brief look through the documentation and didn't find much, but I could be looking in the wrong place.

kneal:
Good point, yeah, might be better to go with this design and expand when needed then. Unless @wass3r remembers the other context, since he mentioned the offline conversation.

wass3r:
For additional context, the pagination was talked about in the offline conversation, so nothing in regards to the time range selection - at least that i can recall.

I'm personally ok with the `before` and `after` or `from` to `to` approach and have it limited to `created`. Like Neal said, should be easy to expand upon if needed.

api/build.go
```diff
@@ -539,7 +549,33 @@ func GetBuilds(c *gin.Context) {
 	// nolint: gomnd // ignore magic number
 	perPage = util.MaxInt(1, util.MinInt(100, perPage))
 
-	b, t, err = database.FromContext(c).GetRepoBuildList(r, filters, page, perPage)
+	// capture before query parameter if present, default to now
+	//
+	// nolint: gomnd, lll // ignore magic number and long line length
+	before, err := strconv.ParseInt(c.DefaultQuery("before", strconv.FormatInt(time.Now().UTC().Unix(), 10)), 10, 64)
```
dtanner, line 555: question - what are the last two arguments?

database/postgres/build_list.go
```diff
@@ -159,6 +159,8 @@ func (c *client) GetRepoBuildList(r *library.Repo, filters map[string]interface{
 	err = c.Postgres.
 		Table(constants.TableBuild).
 		Where("repo_id = ?", r.GetID()).
+		Where("started < ?", before).
```
dtanner, line 162: Something to check: Can you look at this query out in production to see how it behaves from a performance and cost perspective? It's probably ok assuming there's an index on `repo_id` which would narrow the rows down a lot, but since there's not an index on `started`, I'm curious how fast those queries and what their cost is.

I'm assuming you know how to grab the generated query.?  And then you could throw that in a query tool and run `explain analyze $your_query`.  That'll show how long it takes and its cost.  It's probably fine, especially if people don't use this often, but as the database grows over time, this kind of stuff can add up and need to be considered for database tuning and maintenance.

database/postgres/build_list_test.go
```diff
@@ -396,27 +399,31 @@ func TestPostgres_Client_GetRepoBuildList(t *testing.T) {
...
 	// ensure the mock expects the query
 	_mock.ExpectQuery(`SELECT count(*) FROM "builds" WHERE repo_id = $1`).WillReturnRows(_rows)
 
 	// create expected return in mock
 	_rows = sqlmock.NewRows(
 		[]string{"id", "repo_id", "number", "parent", "event", "status", "error", "enqueued", "created", "started", "finished", "deploy", "deploy_payload", "clone", "source", "title", "message", "commit", "sender", "author", "email", "link", "branch", "ref", "base_ref", "head_ref", "host", "runtime", "distribution", "timestamp"},
-	).AddRow(1, 1, 1, 0, "", "", "", 0, 0, 0, 0, "", nil, "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", 0).
-		AddRow(2, 1, 2, 0, "", "", "", 0, 0, 0, 0, "", nil, "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", 0)
+	).AddRow(1, 1, 1, 0, "", "", "", 0, 0, 1, 0, "", nil, "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", 0).
+		AddRow(2, 1, 2, 0, "", "", "", 0, 0, 2, 0, "", nil, "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", 0)
```
dtanner, line 411: nothing to do here right now, but a comment...this many params is hard to understand and prone to errors. named arguments would be better, or a type passed that contains the parameters. something to think about for future development.

dtanner (dismissed):

api/build.go
jbrockopp: Thats the base (`10`) and the size in bits (`64`):

https://pkg.go.dev/strconv#ParseInt

database/postgres/build_list_test.go
kneal: Named params isn't quite a thing in Go, you can pass a struct or call a function on a struct to access it's fields but in this particular instance we are handicapped by the [library](https://pkg.go.dev/github.com/DATA-DOG/go-sqlmock). If you look up the [NewRow in godoc](https://pkg.go.dev/github.com/DATA-DOG/go-sqlmock#NewRows) you'll see it takes that array of strings to define what the columns are and then AddRow takes an array of a custom type which narrows down it [down to primitives](https://pkg.go.dev/database/sql/driver#Value).

database/postgres/build_list.go
GregoryDosh: I agree that performance should be explored and validated. A LIMIT doesn't do much if it's paired up with an ORDER statement since the DB will still have to do a full sort on that ORDER column before it can even consider LIMITing the return rows. If you didn't have the order, then a limit on its own would probably be fine, but we have them both together.

https://dba.stackexchange.com/questions/110636/poor-performance-on-query-with-limit-when-i-add-an-order-by

ecrupper: We indeed have `repo_id` as an index in the DB. I ran a few test queries as well as the `explain analyse` tool. It seems like the query is fairly performant. I got results that varied from a couple tenths of a millisecond to two milliseconds.

kneal: It's probably worth calling out the [indexes](https://github.com/go-vela/server/blob/master/database/postgres/ddl/repo.go#L41-L47) are visible from the application code.

database/postgres/build_list.go
```diff
@@ -159,6 +159,8 @@ func (c *client) GetRepoBuildList(r *library.Repo, filters map[string]interface{
...
+		Where("started > ?", after).
```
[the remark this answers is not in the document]
kneal: Yeah, I think `created` would be better personally just because it's more full representation of start time including it being in the queue, but I'm not sure about the customer use case. 

I think either would be fine, just have a preference towards created.

api/build.go
```diff
@@ -407,6 +407,16 @@ func skipEmptyBuild(p *pipeline.Build) string {
...
 //   maximum: 100
 //   default: 10
+// - in: query
+//   name: before
+//   description: filter builds before a certain time
+//   type: integer
+//   default: 1
+// - in: query
+//   name: after
+//   description: filter builds after a certain time
```
jbrockopp, line 417: Do you think it's worth clarifying this is for builds `created`?

```suggestion
//   description: filter builds created after a certain time
```

jbrockopp (changes requested):

jbrockopp (approved):
LGTM

wass3r (approved):

PR #588, opened 2022-02-15
fix(secrets): reject whitespace value secrets

ecrupper:
The UI already prevents this by trimming, but you can still create secrets that are solely whitespace via the API endpoint.


dtanner (approved):

jbrockopp (approved):
LGTM

wass3r (approved):

PR #598, opened 2022-02-26
fix(api/secret): get rid of whitespace check for secret updating

ecrupper:
When trying to update a secret to handle more events, I received a `400` because I wanted to keep the value of the secret the same. In other words, we already handle whitespace updates by simply keeping the original secret value — we don't need this check; in fact, it's just a bug if we leave it.


KellyMerrick (approved):
lgtm

PR #607, opened 2022-03-07
feat(repo): accommodate last update field in server

ecrupper:
Part of the effort for https://github.com/go-vela/community/issues/343

Setting the `lastUpdate` field within the `planBuild` function as well as whenever a user enables a repo


jbrockopp (changes requested):
Based off the definition of done in the linked issue:

> Anytime a build is ran, that repo will go the top of the list for that org.

That use case could be satisfied with a `JOIN` database query with the `builds` and `repos` tables.

What are your thoughts on using that approach vs adding another field to the `repos` table?

ecrupper:
> That use case could be satisfied with a JOIN database query with the builds and repos tables.
> What are your thoughts on using that approach vs adding another field to the repos table?

Yeah I was thinking about that as a potential solution. The only reason I didn't pursue it was because I suspected it may be too time intensive to sort through all the builds rather than the repos. If you don't believe that would be an issue, then I'm happy to revert those changes and pursue the table join route.

ecrupper:
Closing due to the pursuit of the DB query route instead of adding a field

PR #611, opened 2022-03-15
feat(api/repo): add ability to sort by latest

ecrupper:
Part of the effort for https://github.com/go-vela/community/issues/343

- Adds new query parameter to the `GetOrgRepos` function: `sort_by`
- Utilize SQL query to return list of all org repos sorted by their latest build's created timestamp in descending order


cognifloyd (dismissed):
LGTM.

I'm not sure what you can do (if anything) about Github's "Code scanning" warning: "Database query built from user-controlled sources" since that's how it was before.

jbrockopp (dismissed):
LGTM

api/repo.go
```diff
@@ -427,6 +427,11 @@ func GetRepos(c *gin.Context) {
 //   type: integer
 //   maximum: 100
 //   default: 10
+// - in: query
+//   name: sort_by
+//   description: How to sort the results
+//   type: string
+//   default: name
```
wass3r, line 434: should we add `enum` here? it can really only be `name` or `latest` right?

see example here: https://github.com/go-vela/server/blob/61358d2df68207f980a7845270193f50e0eb5709/api/secret.go#L43-L46

wass3r (commented):
super minor suggestion. otherwise, looks good

jbrockopp (approved):
LGTM

wass3r (approved):

cognifloyd (approved):

PR #615, opened 2022-03-18
feat(api)!: add support for pipelines

jbrockopp:
Dependent on https://github.com/go-vela/server/pull/574, https://github.com/go-vela/server/pull/626, and https://github.com/go-vela/server/pull/627

Part of the effort for https://github.com/go-vela/community/issues/460

This **BREAKING CHANGE** adds first class support for `pipelines` in the system.

## API

This adds CRUD support for `pipelines` to the `github.com/go-vela/server/api` package.

Previously, we hosted a suite of API endpoints for `pipelines` that enabled various interactions with them:

https://github.com/go-vela/server/blob/29562a46eb6d88e85cf037f6a073b6a83d50cae5/router/pipeline.go#L17-L21

A `ref` query param (usually with a commit SHA as the value) was used to tell Vela which pipeline to fetch for a repo.

If no `ref` query param was provided, then Vela would use the default branch for the repo.

Now, these endpoints have been modified to require the commit SHA in the path (`:pipeline`):

https://github.com/go-vela/server/blob/da3142e1d282e3b7dae8153c271ed9b9bb5713cd/router/pipeline.go#L19-L27

This means the lookup for `pipelines` stored in our system is based off the commit SHA provided in the path.

If a pipeline hasn't been stored in the Vela system with that commit SHA, we'll now return a `404 Not Found` error.

## Compiler

Both the `Compile()` and `CompileLite()` functions have been updated to return a `*library.Pipeline` type:

https://github.com/go-vela/server/blob/da3142e1d282e3b7dae8153c271ed9b9bb5713cd/compiler/engine.go#L20-L28

This returned object will have most of its fields set excluding `repo_id`, `commit` and `ref` (set from the `build`):

https://github.com/go-vela/server/blob/da3142e1d282e3b7dae8153c271ed9b9bb5713cd/compiler/native/compile.go#L50-L53

The `Parse()` function was also updated to return a `[]byte`:

https://github.com/go-vela/server/blob/da3142e1d282e3b7dae8153c271ed9b9bb5713cd/compiler/engine.go#L34-L36

This contains the raw pipeline, meaning whatever file thats fetched from the source provider.

This is used to set the `data` field which will be compressed before storing in the database.

## Webhook

The webhook workflow has been modified to start storing `pipelines` fetched during this process in the Vela system.

For every received webhook, we make a call to the database to see if that pipeline already exists:

https://github.com/go-vela/server/blob/da3142e1d282e3b7dae8153c271ed9b9bb5713cd/api/webhook.go#L386-L402

If it doesn't, then we fallback to fetching the pipeline configuration file from the SCM.

After we've ensured its not an empty build (only `init` and/or `clone`), we store the pipeline in the database:

https://github.com/go-vela/server/blob/da3142e1d282e3b7dae8153c271ed9b9bb5713cd/api/webhook.go#L506-L536


api/build.go
```diff
@@ -274,6 +274,48 @@
...
+	pipeline.SetRef(input.GetRef())
+
+	if lastPipeline != nil {
+		pipeline.SetNumber(lastPipeline.GetNumber() + 1)
+	}
+
+	// send API call to create the pipeline
+	err = database.FromContext(c).CreatePipeline(pipeline)
+	if err != nil {
+		retErr := fmt.Errorf("%s: failed to create pipeline for %s: %v", baseErr, r.GetFullName(), err)
```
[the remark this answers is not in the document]
jbrockopp: fixed 👍

api/build.go
```diff
@@ -274,6 +274,48 @@
...
+		util.HandleError(c, http.StatusBadRequest, retErr)
+
+		return
+	}
+
+	// send API call to capture the created pipeline
+	pipeline, err = database.FromContext(c).GetPipelineForRepo(pipeline.GetNumber(), r)
+	if err != nil {
+		// nolint: lll // ignore long line length due to error message
+		retErr := fmt.Errorf("%s: failed to get new pipeline %s/%d: %v", baseErr, r.GetFullName(), pipeline.GetNumber(), err)
```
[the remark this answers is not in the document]
jbrockopp: fixed 👍

api/build.go
```diff
@@ -978,47 +1020,134 @@
...
+		pipeline.SetRef(b.GetRef())
+
+		if lastPipeline != nil {
+			pipeline.SetNumber(lastPipeline.GetNumber() + 1)
+		}
+
+		// send API call to create the pipeline
+		err = database.FromContext(c).CreatePipeline(pipeline)
+		if err != nil {
+			retErr := fmt.Errorf("%s: failed to create pipeline for %s: %v", baseErr, r.GetFullName(), err)
```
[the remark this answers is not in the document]
jbrockopp: fixed 👍

cognifloyd (approved):
It might be large, but it is repetitive (follows a pattern), so it wasn't that bad to review. Thanks for the overview. Looks great to me!

api/pipeline/validate.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// swagger:operation POST /api/v1/pipelines/{org}/{repo}/{pipeline}/validate pipelines ValidatePipeline
+//
+// Get, expand and validate a pipeline from the configured backend
+//
+// ---
+// produces:
+// - application/json
```
wass3r, line 29: add yaml here as well?

api/pipeline/validate.go
```diff
@@ -0,0 +1,113 @@
...
+//   type: string
+// - in: path
+//   name: pipeline
+//   description: Commit SHA for pipeline to retrieve
+//   required: true
+//   type: string
+// - in: query
+//   name: output
+//   description: Output string for specifying output format
+//   type: string
```
wass3r, line 49: enum again with default

api/pipeline/compile.go
```diff
@@ -0,0 +1,106 @@
...
+//   type: string
+// - in: path
+//   name: pipeline
+//   description: Commit SHA for pipeline to retrieve
+//   required: true
+//   type: string
+// - in: query
+//   name: output
+//   description: Output string for specifying output format
+//   type: string
```
wass3r, line 50: add enum with default

api/pipeline/create.go
```diff
@@ -0,0 +1,121 @@
...
+	// update engine logger with API metadata
+	//
+	// https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Entry.WithFields
+	logger := logrus.WithFields(logrus.Fields{
+		"org":  o,
+		"repo": r.GetName(),
+		"user": u.GetName(),
+	})
+
+	logger.Infof("creating new pipeline for repo %s", r.GetFullName())
```
wass3r, line 83: with the given fields, the contents of `%s` don't provide any new information. should we drop it?

api/pipeline/delete.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// update engine logger with API metadata
+	//
+	// https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Entry.WithFields
+	logrus.WithFields(logrus.Fields{
+		"org":      o,
+		"pipeline": p.GetCommit(),
+		"repo":     r.GetName(),
+		"user":     u.GetName(),
+	}).Infof("deleting pipeline %s", entry)
```
wass3r, line 79: with the given fields, the contents of `%s` don't provide any new information. should we drop it?

api/pipeline/expand.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// update engine logger with API metadata
+	//
+	// https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Entry.WithFields
+	logrus.WithFields(logrus.Fields{
+		"org":      o,
+		"pipeline": p.GetCommit(),
+		"repo":     r.GetName(),
+		"user":     u.GetName(),
+	}).Infof("expanding templates for pipeline %s", entry)
```
wass3r, line 88: with the given fields, the contents of `%s` don't provide any new information. should we drop it?

api/pipeline/get.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// update engine logger with API metadata
+	//
+	// https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Entry.WithFields
+	logrus.WithFields(logrus.Fields{
+		"org":      o,
+		"pipeline": p.GetCommit(),
+		"repo":     r.GetName(),
+		"user":     u.GetName(),
+	}).Infof("reading pipeline %s/%s", r.GetFullName(), p.GetCommit())
```
wass3r, line 69: with the given fields, the contents of `%s` don't provide any new information. should we drop it?

api/pipeline/list.go
```diff
@@ -0,0 +1,139 @@
...
+	u := user.Retrieve(c)
+
+	// update engine logger with API metadata
+	//
+	// https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Entry.WithFields
+	logrus.WithFields(logrus.Fields{
+		"org":  o,
+		"repo": r.GetName(),
+		"user": u.GetName(),
+	}).Infof("listing pipelines for repo %s", r.GetFullName())
```
wass3r, line 91: with the given fields, the contents of `%s` don't provide any new information. should we drop it?

api/pipeline/template.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// update engine logger with API metadata
+	//
+	// https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Entry.WithFields
+	logrus.WithFields(logrus.Fields{
+		"org":      o,
+		"pipeline": p.GetCommit(),
+		"repo":     r.GetName(),
+		"user":     u.GetName(),
+	}).Infof("reading templates from pipeline %s", entry)
```
wass3r, line 90: with the given fields, the contents of `%s` don't provide any new information. should we drop it?

api/pipeline/update.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// update engine logger with API metadata
+	//
+	// https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Entry.WithFields
+	logrus.WithFields(logrus.Fields{
+		"org":      o,
+		"pipeline": p.GetCommit(),
+		"repo":     r.GetName(),
+		"user":     u.GetName(),
+	}).Infof("updating pipeline %s", entry)
```
wass3r, line 86: with the given fields, the contents of `%s` don't provide any new information. should we drop it?

api/pipeline/validate.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// update engine logger with API metadata
+	//
+	// https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Entry.WithFields
+	logrus.WithFields(logrus.Fields{
+		"org":      o,
+		"pipeline": p.GetCommit(),
+		"repo":     r.GetName(),
+		"user":     u.GetName(),
+	}).Infof("validating pipeline %s", entry)
```
wass3r, line 86: with the given fields, the contents of `%s` don't provide any new information. should we drop it?

api/pipeline/expand.go
```diff
@@ -0,0 +1,107 @@
...
+//   type: string
+// - in: path
+//   name: pipeline
+//   description: Commit SHA for pipeline to retrieve
+//   required: true
+//   type: string
+// - in: query
+//   name: output
+//   description: Output string for specifying output format
+//   type: string
```
wass3r, line 50: enum with default?

api/pipeline/get.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// swagger:operation GET /api/v1/pipelines/{org}/{repo}/{pipeline} pipelines GetPipeline
+//
+// Get a pipeline from the configured backend
+//
+// ---
+// produces:
+// - application/json
```
wass3r, line 22: add yaml here?

wass3r (changes requested):
nice work. first pass - see comments

one additional nitpick.. not sure if it's just me, but i'd almost like to see `:pipeline` changed to `:sha` in the paths (eg. `/api/v1/pipelines/:org/:repo/:pipeline`) or something. maybe that's leftover from when you used IDs? for documentation purposes that seems more clear to me.

api/build.go
```diff
@@ -256,6 +275,8 @@ func CreateBuild(c *gin.Context) {
 
 		return
 	}
+	// reset the pipeline type for the repo
+	r.SetPipelineType(pipelineType)
```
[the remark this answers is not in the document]
jbrockopp: fixed 👍

api/pipeline/get.go
```diff
@@ -0,0 +1,72 @@
...
+//   type: string
+// - in: path
+//   name: pipeline
+//   description: Commit SHA for pipeline to retrieve
+//   required: true
+//   type: string
+// - in: query
+//   name: output
+//   description: Output string for specifying output format
+//   type: string
```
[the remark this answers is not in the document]
jbrockopp: removed `output` so this is fixed 👍

api/pipeline/get.go
```diff
@@ -0,0 +1,72 @@
...
+	//
+	// https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Entry.WithFields
+	logrus.WithFields(logrus.Fields{
+		"org":      o,
+		"pipeline": p.GetCommit(),
+		"repo":     r.GetName(),
+		"user":     u.GetName(),
+	}).Infof("reading pipeline %s/%s", r.GetFullName(), p.GetCommit())
+
+	writeOutput(c, p)
```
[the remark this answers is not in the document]
jbrockopp: fixed 👍

api/pipeline/template.go
```diff
@@ -0,0 +1,146 @@
...
+//   type: string
+// - in: path
+//   name: pipeline
+//   description: Commit SHA for pipeline to retrieve
+//   required: true
+//   type: string
+// - in: query
+//   name: output
+//   description: Output string for specifying output format
+//   type: string
```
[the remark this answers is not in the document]
jbrockopp: fixed 👍

api/pipeline/validate.go
jbrockopp: fixed 👍

api/pipeline/validate.go
jbrockopp: fixed 👍

api/pipeline/expand.go
jbrockopp: fixed 👍

api/pipeline/get.go
jbrockopp: removed `output` so this is fixed 👍

api/pipeline/compile.go
```diff
@@ -0,0 +1,106 @@
...
+
+	// update engine logger with API metadata
+	//
+	// https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Entry.WithFields
+	logrus.WithFields(logrus.Fields{
+		"org":      o,
+		"pipeline": p.GetCommit(),
+		"repo":     r.GetName(),
+		"user":     u.GetName(),
+	}).Infof("compiling pipeline %s", entry)
```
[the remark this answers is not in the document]
jbrockopp: We certainly could 👍 I'm leaning towards leaving it in there since it doesn't hurt.

This pattern is adopted from existing log entires we already established for other API endpoints.

So if we did want to switch this up, I think we'd have to go back and apply the same changes for consistency.

api/pipeline/expand.go
jbrockopp: responded in earlier comment 👍

api/pipeline/get.go
jbrockopp: responded in earlier comment 👍

api/pipeline/template.go
jbrockopp: responded in earlier comment 👍

api/pipeline/update.go
jbrockopp: responded in earlier comment 👍

api/pipeline/validate.go
jbrockopp: responded in earlier comment 👍

cognifloyd (approved):

api/pipeline/compile.go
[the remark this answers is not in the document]
wass3r: Fair point, I'd vote for removing it maybe in another PR. Seems like a waste of bytes that can add up - then again, this is more compact than logging as separate JSON fields :disappear:

wass3r (approved):
thanks for addressing feedback. i see there were no hot takes on `:pipeline` vs `:sha` or something similar in URL path for docs. it's not crucial, just something I felt took extra brain cycles to understand what is expected, though the description for the param define it more.

api/pipeline/compile.go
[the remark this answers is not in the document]
jbrockopp: Yeah I see where you're coming from on that 👍 

To me, the message for the log entry is designed for human processing

i.e. user who manually reviews the log entries or using a tool such as `grep` or `awk`

In those scenarios, it can add a bit of complexity when trying to search for multiple fields but not impossible:

```shell
# searching for specific message
cat log.txt | grep '<org>/<repo>/<commit>'

# searching for specific fields
cat log.txt | grep '"org": <org>' | grep '"repo": <repo>' | grep '"pipeline": <commit>'
```

While the fields for the log entry are designed for system processing

i.e. most apps/tools consuming logs enable filtering on one or more fields which improves this experience

However, if folks feel strongly enough about it, then I could get behind shortening the message

wass3r:
That's fair. I think with `:repo` (and most other path fragments), it's relatively intuitive to infer that you're looking for either "name" or "number", but `:pipeline` being something totally different kind of throws it for a loop. But I digress - I prefaced with it being a nitpick and not something that ought to stop the merge. Thanks for the additional comment.

PR #619, opened 2022-03-24
enhance!: modify CI environment value to true

dtanner:
https://github.com/go-vela/community/issues/554


compiler/native/environment_test.go
```diff
@@ -6,6 +6,7 @@ package native
 
 import (
 	"flag"
+	"github.com/urfave/cli/v2"
```
cognifloyd, line 9: nit: revert import order change.

cognifloyd (approved):
I have a minor nit, but otherwise LGTM.

compiler/native/environment_test.go
wass3r: yea, we should configure one of the linters to catch this for us. i think we have discrepancies on import order in a few places.

wass3r (approved):

ecrupper (approved):
LGTM

wass3r (approved):

PR #622, opened 2022-03-29
feat(scm/api/hook): create in-house webhook redelivery method

ecrupper:
Part of the effort for https://github.com/go-vela/community/issues/459

Includes:
- New endpoint to redeliver a webhook using the `go-github` API
- Setting `hook.WebhookID` field during the `ProcessWebhook` function
- Some updates to DB tests due to some parallel changes to the pipeline story, see: https://github.com/go-vela/types/pull/223

Why is this a draft?
- Currently using non-versioned instance of the `go-github` library. Will update to latest version when they release it.

Things to note:
- It is necessary that whichever users adopt this have a version of GitHub that is `3.2` or higher. Otherwise, there will be `415` errors when trying to execute. I added some error handling for that in this PR.


plyr4 (approved):

KellyMerrick (approved):
lgtm

PR #625, opened 2022-04-15
feat(hook): set webhook_id in process webhook function

ecrupper:
Since https://github.com/go-vela/server/pull/622 will not be merged relatively soon, the piece of that PR that is absolutely necessary is the part where the `webhook_id` is set during `ProcessWebhook`. Redelivery can wait.


scm/github/webhook.go
```diff
@@ -27,6 +28,14 @@ func (c *client) ProcessWebhook(request *http.Request) (*types.Webhook, error) {
 	h := new(library.Hook)
 	h.SetNumber(1)
 	h.SetSourceID(request.Header.Get("X-GitHub-Delivery"))
+
+	hookID, err := strconv.Atoi(request.Header.Get("X-GitHub-Hook-ID"))
+	if err != nil {
+		retErr := fmt.Errorf("unable to convert hook id to int64: %w", err)
+		return nil, retErr
```
jbrockopp, line 35: Thoughts on condensing this into one line?

```suggestion
		return nil, fmt.Errorf("unable to convert hook id to int64: %w", err)
```

jbrockopp (changes requested):
one quick change

jbrockopp (approved):
LGTM

plyr4 (approved):

PR #630, opened 2022-04-26
feat(compiler/database/scm)!: add event action build field and use for rulesets

ecrupper:
Part of the effort of https://github.com/go-vela/community/issues/159

PR accomplishes the following:
- Set event action field in `scm/github/webhook.go` if the event is `pull_request` or `comment`. Otherwise, this field remains `nil`. 
- Use the event action field, if not `nil`, to create a scoped `<event>:<action>` combination to utilize for determining whether or not a step should execute. There will be a separate PR to address this at runtime in the worker code.
- Adds the `event_action` field to the ddl


scm/github/webhook.go
```diff
@@ -184,9 +184,20 @@ func (c *client) processPREvent(h *library.Hook, payload *github.PullRequestEven
...
 	}
 
-	// skip if the pull request action is not opened or synchronize
+	// if the edits to the PR are not related to the title or base branch, ignore them
+	if strings.EqualFold(payload.GetAction(), "edited") {
+		title := payload.GetChanges().GetTitle()
+		base := payload.GetChanges().GetBase()
+
+		if title == nil && base == nil {
+			return &types.Webhook{Hook: h}, nil
```
cognifloyd, line 193: Why? What if someone had a workflow that validated the PR description followed some format, would we not support that?

cognifloyd (commented):
I have a question about restricting the PR edited event. Otherwise this looks good

cognifloyd (dismissed):

wass3r:
conflicts :(

cognifloyd (approved):

PR #642, opened 2022-05-18
fix: restaring builds before pipeline support

jbrockopp:
Closes https://github.com/go-vela/community/issues/593


cognifloyd (approved):
Will there be a time that we won't need to fall back to the legacy behavior? Eg for a v1.0 release?

wass3r (approved):

wass3r:
@cognifloyd this particular change is forward and backward compatible, so if we wanted to drop legacy support, this edit wouldn't change.

PR #644, opened 2022-05-22
fix(api): always set Strict-Transport-Security header

jbrockopp:
This change enforces the API to utilize HTTP Strict-Transport-Security (HSTS):

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security

To accomplish this, we set a header to inform browsers that the site should only be accessed using HTTPS:

https://github.com/go-vela/server/blob/8bf333d1e18e6dec3ee7e530412760b6a7453e29/router/middleware/header.go#L56

> NOTE: This is the same value we set for the [go-vela/ui](https://github.com/go-vela/ui/blob/master/nginx/default.conf#L19):
>
> ```nginx
>     add_header Strict-Transport-Security "max-age=63072000; includeSubDomains; preload" always;
> ```

This was flagged, as a low finding, by Cargill's security team during a penetration test of the system.

Their argument is by not enforcing secure connections, via HSTS, enables someone to exploit the product.

If an adversary were positioned in the middle of the endpoint and the server, they could use that against the system.


JordanSussman (approved):

router/middleware/header.go
```diff
@@ -51,12 +51,9 @@ func Secure(c *gin.Context) {
...
 	c.Header("X-Content-Type-Options", "nosniff")
 	c.Header("X-XSS-Protection", "1; mode=block")
-
-	// Also consider adding Content-Security-Policy headers
+	// TODO: consider adding Content-Security-Policy headers
 	// c.Header("Content-Security-Policy", "script-src 'self' https://cdnjs.cloudflare.com")
-	if c.Request.TLS != nil {
-		c.Header("Strict-Transport-Security", "max-age=31536000")
-	}
+	c.Header("Strict-Transport-Security", "max-age=63072000; includeSubDomains; preload")
```
cognifloyd, line 56: Since you're changing the max-age here, does this need to be configurable somehow?

wass3r (approved):

router/middleware/header.go
cognifloyd: Cool. That makes sense. Making it configurable is probably overkill then.

cognifloyd (approved):

kneal (approved):
LGTM 🐬

PR #646, opened 2022-05-23
feat(api/build): add endpoint to get build by id

ecrupper:
Getting build by ID will be very helpful in a number of ways, but the reason at the top of mind is using this endpoint to retrieve a build that corresponds with a hook on the hooks page in the UI.

One endpoint is for users who have at least `read` access to the repo in question, and the other is for admins which bypasses that repo check.

Part of the effort for https://github.com/go-vela/community/issues/610 as well.


router/search.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"github.com/gin-gonic/gin"
+	"github.com/go-vela/server/api"
+)
+
+// SearchHandlers is a function that extends the provided base router group
+// with the API handlers for source code management functionality.
+//
+// GET    /api/v1/search/build/:id .
```
jbrockopp, line 15: plural?

jbrockopp (changes requested):

jbrockopp (changes requested):
I've reviewed the code and it LGTM! However, I had a question that may lead to a discussion.

> NOTE: I'm posting review as request for changes for visibility purposes.

I noticed that today, to query a build for a repo, we require the user has at least `read` level access:

https://github.com/go-vela/server/blob/8e8d0d456b0f57a978e996b85a20402841fabc5e/router/build.go#L56

Did we intentionally make this endpoint a higher access level (`write`)?

https://github.com/go-vela/server/blob/1de361fa72c98423be23ad3f3e7648b0800e1406/api/build.go#L492-L501

If so, could you share some of the thinking around why?

FWIW - I don't have strong concerns about the difference in access level requirements between the two endpoints.

More so, I'm trying to understand the reasoning behind it.

ecrupper:
Yeah, I'm not really sure why I chose `write`. I think complete parity with the current form of getting a build is the way we want to go. I'll change that now

jbrockopp (approved):
LGTM

api/admin/build.go
```diff
@@ -20,6 +20,55 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// swagger:operation GET /api/v1/admin/builds/:id admin AdminSingleBuild
```
wass3r, line 23: with the latest changes, this needs to be updated to `/api/v1/admin/build/:id`

wass3r (changes requested):

api/build.go
```diff
@@ -411,6 +411,96 @@ func skipEmptyBuild(p *pipeline.Build) string {
...
+func GetBuildByID(c *gin.Context) {
+	// Variables that will hold the library types of the build and repo
+	var (
+		b *library.Build
+		r *library.Repo
+	)
+
+	// Capture user from middleware
+	u := user.Retrieve(c)
+
```
plyr4, line 454: can we add a simple log here like `GetBuild` ?
https://github.com/go-vela/server/blob/c565a1d762bbd835b4f05f3d7318b9eda296757d/api/build.go#L955-L963

api/build.go
```diff
@@ -411,6 +411,96 @@ func skipEmptyBuild(p *pipeline.Build) string {
...
+	}
+
+	// Capture user access from SCM
+	perm, err := scm.FromContext(c).RepoAccess(u, u.GetToken(), r.GetOrg(), r.GetName())
+	if err != nil {
+		logrus.Errorf("unable to get user %s access level for repo %s", u.GetName(), r.GetFullName())
+	}
+
+	// Ensure that user has at least read access to repo to return the build
+	if perm == "none" {
```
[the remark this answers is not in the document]
ecrupper: The middleware won't capture the repo from the request, so it's necessary to have this perm check after retrieving the build. This preserves the integrity of the permission barrier.

api/admin/build.go
```diff
@@ -20,6 +20,55 @@ import (
...
+//     schema:
+//       "$ref": "#/definitions/Build"
+//   '500':
+//     description: Unable to retrieve build from the database
+//     schema:
+//       "$ref": "#/definitions/Error"
+
+// SingleBuild represents the API handler to
+// capture a single build stored in the database by id.
+func SingleBuild(c *gin.Context) {
```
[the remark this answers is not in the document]
ecrupper: There used to be an admin endpoint called `AllBuilds`, so I thought I'd follow that pattern. Now that it's no longer a pattern, I can change this.

plyr4 (approved):

wass3r (approved):

PR #653, opened 2022-06-01
refactor: update substitute tests to use table syntax 

JordanSussman:
## changes

- Updates the existing substitute tests to utilize table tests
- Update the SubstituteSteps test to include another test case for templates


wass3r:
looks like the test is failing :(

JordanSussman:
@wass3r this was originally created to showcase that the default variable substitution syntax `${var:-default}` does not appear to be work outside of `commands`. Talking with @jbrockopp we realized the following:

- Some variables do not get set until a build runs on the worker
- The server attempts to make best efforts to substitute all variables available to it at the time
- If a variable isn't set, then assume it may be set by the worker so it "puts it back"

The existing behavior is fine for most use-case but if you're trying to do variable substitution for `vars` passed to a template, you're going have a bad time. Thinking about opening up a new issue to talk through possible changes to this behavior to all default syntax to be used.

compiler/native/substitute_test.go
```diff
@@ -94,91 +198,60 @@ func TestNative_SubstituteStages(t *testing.T) {
...
+						Variables: map[string]interface{}{
+							"build_author": "testauthor",
+							"unknown":      "${DEPLOYMENT_PARAMETER_API_IMAGE}",
+						},
+					},
+					Environment: map[string]string{
+						"BUILD_AUTHOR": "testauthor",
+					},
+				},
+			},
```
plyr4, line 236: should we set `wantErr` explicitly like the other test cases?
```
			wantErr: false,
```

compiler/native/substitute_test.go
```diff
@@ -6,86 +6,190 @@ package native
...
+							Image:       "alpine:latest",
+							Name:        "not_found",
+							Pull:        "always",
+						},
+					},
 				},
 			},
+			wantErr: false,
 		},
+		// TODO: Add test cases.
```
jbrockopp, line 114: I believe this line can be removed.

JordanSussman: Resolved.

compiler/native/substitute_test.go
JordanSussman: Resolved.

jbrockopp (approved):
LGTM

plyr4 (approved):

wass3r (approved):

PR #659, opened 2022-06-13
fix: webhook race condition on pipeline creation

jbrockopp:
closes https://github.com/go-vela/community/issues/608

Similar to https://github.com/go-vela/server/pull/191


kneal (approved):
LGTM 🐬

wass3r (approved):
was able to replicate the issue consistently locally with the old code and the changes here are taking care of the issue. thanks for the fix.

PR #663, opened 2022-06-22
refactor(database): move user logic into separate package

jbrockopp:
Based off of https://github.com/go-vela/server/pull/574

This change continues the refactor effort initially introduced in the above PR.

This adds a new `user` package to the `github.com/go-vela/server/database` package.

This contains a `UserService` interface declaring all functions necessary for user based interactions with the database:

https://github.com/go-vela/server/blob/b2350a95513fedd31ac4f1915b5b2d91821226d1/database/user/service.go#L11-L45

https://github.com/go-vela/server/blob/b2350a95513fedd31ac4f1915b5b2d91821226d1/database/user/user.go#L25-L39

This `engine` contains no raw SQL queries for integrating with the `users` table.


cognifloyd (commented):
Maybe point out in the docstrings ListUsers and ListLiteUsers differ (what makes it "Lite").
It looks like Lite just slips decrypting fields, right? Anything else?

Other than that, it looks good to me.

database/user/list_lite.go
```diff
@@ -0,0 +1,61 @@
...
+
+package user
+
+import (
+	"github.com/go-vela/types/constants"
+	"github.com/go-vela/types/database"
+	"github.com/go-vela/types/library"
+)
+
+// ListLiteUsers gets a lite list of users from the database.
```
cognifloyd, line 13:
```suggestion
// ListLiteUsers gets a lite (only: id, name) list of users from the database.
```

cognifloyd (commented):
Maybe something like this?

cognifloyd (approved):

ecrupper (approved):
LGTM - really like this new pattern

PR #667, opened 2022-06-29
feat(logger): censor build author email for security compliance

ecrupper:
Addresses security concern that build author emails, which can be changed within GitHub enterprise to be a non-work email, were being logged. With this change, all http logs with the build as the body will now censor the email field. 

Abstracted this routine into a `sanitize` function, which can be used in case there are future concerns with logging practices.


plyr4 (dismissed):

wass3r (changes requested):
the [lack of coverage](https://app.codecov.io/gh/go-vela/server/compare/667/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%20comments&utm_term=go-vela#diff-cm91dGVyL21pZGRsZXdhcmUvbG9nZ2VyLmdv) in the report threw me off for a min.

when you do `sanitizeBuild := b` and modify a field, it will modify for both `b` and `santizeBuild`, since it's a pointer. you can get past it by dereferencing like `sanitizeBuild := *b` but you will see the tests fail then. turns out you never get past the `if m, ok := body.(map[string]interface{}); ok` check and always just return the original `body`.

if we can fix up the tests, this is good to go, i think.

ecrupper:
> the [lack of coverage](https://app.codecov.io/gh/go-vela/server/compare/667/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%20comments&utm_term=go-vela#diff-cm91dGVyL21pZGRsZXdhcmUvbG9nZ2VyLmdv) in the report threw me off for a min.
> 
> when you do `sanitizeBuild := b` and modify a field, it will modify for both `b` and `santizeBuild`, since it's a pointer. you can get past it by dereferencing like `sanitizeBuild := *b` but you will see the test fail then. turns out you never get past the `if m, ok := body.(map[string]interface{}); ok` check and always just return the original `body`.
> 
> if we can fix up the tests, this is good to go, i think.

Good catch!

plyr4 (approved):

wass3r (approved):

PR #671, opened 2022-07-14
fix(webhook): avoid panics

wass3r:
closes https://github.com/go-vela/community/issues/643

most details on this can be found in the issue above. i commented some of the changes. 

we also added some TODOs for things to look at in lieu of creating new issues to start some conversation and not lose track of it.

we are adding to top of HEAD since there were only two minor merges that only impact tests.

### DIFF
--- a/api/build.go
+++ b/api/build.go
@@ -7,6 +7,7 @@ package api
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -1461,6 +1462,7 @@ func planBuild(database database.Service, p *pipeline.Build, b *library.Build, r
 
 	// send API call to create the build
 	err := database.CreateBuild(b)
+	err = errors.New("error creating build")
 	if err != nil {
 		// clean up the objects from the pipeline in the database
 		cleanBuild(database, b, nil, nil)
--- a/api/build.go
+++ b/api/build.go
@@ -7,7 +7,6 @@ package api
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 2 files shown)


api/webhook.go
```diff
@@ -402,14 +402,6 @@ func PostWebhook(c *gin.Context) {
...
 				retErr := fmt.Errorf("%s: unable to get pipeline configuration for %s: %w", baseErr, r.GetFullName(), err)
 
-				// check if the retry limit has been exceeded
-				if i < retryLimit {
-					logrus.WithError(retErr).Warning("retrying")
-
-					// continue to the next iteration of the loop
-					continue
-				}
-
```
wass3r, line 412: opted to remove this retry logic since ConfigBackoff is already retrying 5x. if that fails it's likely this is just failing, period.

api/build.go
```diff
@@ -1455,21 +1455,36 @@ func getPRNumberFromBuild(b *library.Build) (int, error) {
...
 		cleanBuild(database, b, nil, nil)
 
 		return fmt.Errorf("unable to create new build for %s: %w", r.GetFullName(), err)
 	}
 
 	// send API call to capture the created build
-	b, _ = database.GetBuild(b.GetNumber(), r)
+	// TODO: this can be dropped once we return
+	// the created build above
+	b, err = database.GetBuild(b.GetNumber(), r)
```
wass3r, line 1484: if we don't handle any potential errors here, we will get issues downstream from here

api/webhook.go
```diff
@@ -618,6 +639,8 @@ func PostWebhook(c *gin.Context) {
 
 		h.SetStatus(constants.StatusFailure)
 		h.SetError(retErr.Error())
+
+		return
```
wass3r, line 643: we probably don't want to return a build if there was an error. this also creates a conflict because this sets a 500 return code that later gets changed to a 200.

cognifloyd (approved):

jbrockopp (approved):
LGTM

plyr4 (approved):

ecrupper (approved):
lgtm

KellyMerrick (approved):
lgtm

PR #678, opened 2022-07-21
fix(build/ruleset): handle legacy PR ruleset for restarted build

ecrupper:
Any pull request build ran before 0.14 will not have the action field populated, thus upon restarting, it will not run properly against steps with a `event: pull_request` ruleset.

### DIFF
--- a/api/build.go
+++ b/api/build.go
@@ -990,6 +990,14 @@ func RestartBuild(c *gin.Context) {
 	b.SetRuntime("")
 	b.SetDistribution("")
 
+	// update the PR event if action was never set
+	if b.GetEvent() == constants.EventPull && b.GetEventAction() == "" {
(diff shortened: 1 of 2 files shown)


api/build.go
```diff
@@ -990,6 +990,14 @@ func RestartBuild(c *gin.Context) {
 	b.SetRuntime("")
 	b.SetDistribution("")
 
+	// update the PR event if action was never set
```
wass3r, line 993:
```suggestion
	// update the PR event action if action was never set
	// for backwards compatibility with pre-0.14 releases
```
a little nitpicky, but thought we can be more explicit. thoughts?

wass3r (approved):

plyr4 (approved):

PR #679, opened 2022-08-01
enhance(scm/repo): mirror allowed events with sent events

ecrupper:
Closes https://github.com/go-vela/community/issues/539

Now that we store the `webhook_id` in the hooks table, we're able to edit the webhook whenever a user updates their repo settings. With this functionality, we will consume far less webhook payloads from the SCM and users won't have to deal with error messages for events they do not want enabled.

A thing to note: I added an `initialize` webhook whenever a repo is enabled. This ensures that when a user adjusts settings right after enabling their repo, we have record of the webhook_id to edit.

Would love to hear thoughts!


cognifloyd:
Nice! From the issue description, we really need this.

I don't understand what the issue with push vs tag is. If you have time, could you post a few links to where the code needs to be disambiguated? I can also go looking next time I have some time to look.

ecrupper:
> I don't understand what the issue with push vs tag is. If you have time, could you post a few links to where the code needs to be disambiguated? I can also go looking next time I have some time to look.

For sure! The way GitHub is set up, they send tag events as a sub-action of a push ([Docs](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#push)). We distinguish this in the code in [scm/webhook.go](https://github.com/go-vela/server/blob/66ef7d75000d03c51b4ca23689dfdc07979cddcc/scm/github/webhook.go#L178-L189). So if a user chose to unsubscribe to `push` events but not `tag` events, then we still need to handle all push events. I think a cleaner way would be to do some sort of `if r.GetAllowPush() || r.GetAllowTag()`, which I think I'll change now.

cognifloyd (approved):
Thanks for the links. That makes more sense.

Maybe update the PR description for future reviewers since that question has been resolved. 😄

jbrockopp (dismissed):
LGTM

wass3r (changes requested):
in my testing against github.com this does not seem to work. when enabling a repo i get
```
Status 500 (unable to create initialization webhook for wass3r/vela-test: empty webhook repo_id provided)
```
we will need to set repo id or otherwise solve for this. the error comes from: https://github.com/go-vela/types/blob/ef95941428d9195503c3db067a9f27447b4dcc16/database/hook.go#L162

wass3r (changes requested):
little bit of whack-a-mole.. now getting this error:
```
Status 500 (unable to create initialization webhook for wass3r/vela-test: empty webhook source_id provided)
```
😭

ecrupper:
Sorry about that! All required hook fields have been set now. I also included some extra testing in the scm `enable` function

wass3r (dismissed):
thanks for the fixes. tested on github.com and works great. nice work.

too bad we can't force update existing hooks in github.

jbrockopp (dismissed):
LGTM

plyr4 (dismissed):

plyr4 (approved):

api/repo/update.go
```diff
@@ -238,6 +239,47 @@ func UpdateRepo(c *gin.Context) {
...
+	// if repo has no hook deliveries, skip webhook update
+	if lastHook.GetWebhookID() != 0 {
+		// if user is platform admin, fetch the repo owner token to make changes to webhook
+		if u.GetAdmin() {
+			// log admin override update repo hook
+			logrus.WithFields(logrus.Fields{
+				"org":  o,
+				"repo": r.GetName(),
+				"user": u.GetName(),
+			}).Infof("platform admin updating repo webhook events for repo %s", r.GetFullName())
```
wass3r, line 261: can we move it to _after_ we get the user, because it's not until then, that user is actually getting used?

wass3r (changes requested):

wass3r (approved):

wass3rw3rk (approved):

jbrockopp (approved):
LGTM

PR #682, opened 2022-08-09
enhance(metrics)!: query parameter support on metrics endpoint

plyr4:
### what?
this adds query parameter support to the `/metrics` endpoint.

### why? 

without selective querying, the `/metrics` endpoint will perform expensive and exhaustive queries to retrieve information from the database

### breaking change!

the default values for each query parameter are `false` so that you must selectively request additional information from the database, which means anyone using the `/metrics` endpoint as it works currently will stop receiving data unless they update their query from 

```sh
curl 'localhost:8080/metrics'
```
to
```sh
curl 'localhost:8080/metrics?user_count=t&build_count=true'
```
and include each piece of data they would like to query for.

### docs

once i get approval on the structure and usage of the parameters i will create the documentation before merging


cognifloyd (dismissed):
I don't use the metrics endpoint so far, but this looks sane from reading.

PR #687, opened 2022-08-27
refactor(database): move repo logic into separate package

jbrockopp:
Based off of https://github.com/go-vela/server/pull/574 and https://github.com/go-vela/server/pull/663

This adds a new `repo` package to the `github.com/go-vela/server/database` package.

This contains a `RepoService` interface declaring all functions necessary for user based interactions with the database:

https://github.com/go-vela/server/blob/ca230da3a70cabb0a2b9fae49c7b7891c4cdf461/database/repo/service.go#L11-L51

https://github.com/go-vela/server/blob/ca230da3a70cabb0a2b9fae49c7b7891c4cdf461/database/repo/repo.go#L25-L39

This `engine` contains no raw SQL queries for integrating with the `users` table.


wass3r (commented):
i see you replaced `.Scan()` with `.Take()` in a few places. based on GORM docs (https://gorm.io/docs/query.html#Retrieving-a-single-object), it looks like `Take` already does `LIMIT 1`? Can we remove the preceding `.Limit(1)` in those instances? clearly it passed unit testing so it must account for it somehow?

jbrockopp:
@wass3r good catch! 👍 

To provide some background/context in case others are curious, this is the reason why we're using `Take()`:

https://gorm.io/docs/v2_release_note.html#ErrRecordNotFound

In short, for several of our DB queries, we expect an error if the record doesn't exist.

However, `gorm.io` changed their behavior from V1 -> V2 regarding when that error message is returned:

https://github.com/go-gorm/gorm/issues/4416

In order to account for this today, we have extra code across our DB package that does this:

https://github.com/go-vela/server/blob/2f14f3894de0b7f82bee337b8f042926a4b4385d/database/postgres/repo.go#L37-L40

Links for reference:

* https://github.com/go-vela/server/pull/409
* https://github.com/go-vela/server/pull/410

wass3r (approved):

PR #692, opened 2022-09-11
refactor(database): move worker logic into separate package

jbrockopp:
Based off of https://github.com/go-vela/server/pull/574, https://github.com/go-vela/server/pull/663 and https://github.com/go-vela/server/pull/687

This adds a new `worker` package to the `github.com/go-vela/server/database` package.

This contains a `WorkerService` interface declaring all functions necessary for worker based interactions with the database:

https://github.com/go-vela/server/blob/7991f03903622a820698678037cbe482828d72a2/database/worker/service.go#L11-L43

https://github.com/go-vela/server/blob/7991f03903622a820698678037cbe482828d72a2/database/worker/worker.go#L23-L37

This `engine` contains no raw SQL queries for integrating with the `workers` table.


jbrockopp:
@cognifloyd `GetWorkerByAddress` was not used anywhere in the codebase 😅 

When I discovered this during the refactor, I elected to not include it.

However, if we feel there is a good reason to keep it, I could certainly do so.

wass3r (approved):

plyr4 (approved):

PR #700, opened 2022-09-22
enhance(secrets): verify casing of orgs and repos in SCM before adding secret

ecrupper:
GitHub is case insensitive, while Vela is not. This causes issues when a user attempts to create an org or repo secret with the wrong casing. The secret is created and added to the database, which is unexpected behavior when the user navigates to their secrets page and does not see it there.

Example of the new casing check, given that `EastonCrupper/myvela` is the repo in Vela:
```sh
vela add secret --secret.engine native --secret.type repo --org EastonCrupper --repo MyVela --name example --value blahblahblah
DEBU[0000] parsing Vela client from provided configuration 
DEBU[0000] validating provided configuration for Vela client 
TRAC[0000] creating Vela client for http://localhost:8080 
TRAC[0000] setting token for Vela client                
DEBU[0000] validating secret configuration              
DEBU[0000] executing add for secret configuration       
DEBU[0000] capturing value for secret example           
TRAC[0000] adding secret native/repo/EastonCrupper/MyVela/example 
FATA[0000] unable to retrieve repository MyVela. Did you mean myvela? 
```


wass3r (changes requested):

wass3r (approved):

KellyMerrick (approved):
lgtm

PR #701, opened 2022-09-22
enhance(build): return build on CreateBuild and UpdateBuild

ecrupper:
Part of the effort for https://github.com/go-vela/community/issues/613. It would be helpful to not have to make two calls to the database during our webhook process. This can be remedied by returning the created build object when it's saved to the database.


cognifloyd (approved):
Looks sane

cognifloyd (approved):

database/postgres/build.go
```diff
@@ -156,17 +158,25 @@ func (c *client) CreateBuild(b *library.Build) error {
...
-	return c.Postgres.
+	err = c.Postgres.
 		Table(constants.TableBuild).
 		Create(build.Crop()).Error
+
+	if err != nil {
+		return nil, err
+	}
+
+	return build.Crop().ToLibrary(), nil
```
jbrockopp, line 173: Have we confirmed the `build` object has the ID field populated with this change?

We don't explicitly set that ID field in our codebase which is why I ask.

jbrockopp (changes requested):

database/postgres/build.go
cognifloyd: When gorm saves the object, it populates the ID field, right?

ecrupper:
closing because this has become a mess. Will just use a new branch

PR #707, opened 2022-09-29
fix(webhook): more general error message for redelivery failure

ecrupper:
Sometimes the webhook delivery will no longer be available due to how old it is, not whether it is in the 100 most recent. Upon failure, it's a better idea to give a general error to avoid confusing the user.


wass3r (approved):

KellyMerrick (approved):
lgtm

plyr4 (approved):

jbrockopp (approved):
LGTM

PR #713, opened 2022-10-06
feat(templates): support same-repo templates using build commit

ecrupper:
When working with templates that are found in the same repo as the Vela file, any live edits to the templates requires the users to reference their specific branch in the templates block. Example: 
```yaml
templates:
  - name: sample
    source: https://github.com/<org>/<repo>/path/template.yml@feature-branch
    type: github
```

With this change, users can specify the type `file` in their templates block, which will automatically fetch the template from the commit associated with the build. Example: 
```yaml
templates:
  - name: sample
    source: path/template.yml
    type:  file
```


JordanSussman:
@ecrupper I feel like we should have the `type` be either `github-commit` or `commit-github` incase we want to add additional source code systems in the provider.

compiler/native/expand_test.go
```diff
@@ -287,25 +317,29 @@ func TestNative_ExpandSteps(t *testing.T) {
...
-	if err != nil {
-		t.Errorf("ExpandSteps returned err: %v", err)
-	}
+	compiler.WithBuild(testBuild).WithRepo(testRepo)
 
-	if diff := cmp.Diff(build.Steps, wantSteps); diff != "" {
-		t.Errorf("ExpandSteps() mismatch (-want +got):\n%s", diff)
-	}
+	for _, test := range tests {
+		build, err := compiler.ExpandSteps(&yaml.Build{Steps: steps, Services: yaml.ServiceSlice{}, Environment: globalEnvironment}, test.tmpls)
```
cognifloyd, line 323: Would you please use `t.Run()` here? That makes selecting subtests easier in Goland.

cognifloyd (changes requested):
Very nice!

I think we should stick with `commit` (or something similar) that does not mention github. See my comment.

I have one nitpick about the test. Otherwise, I say this is good to go!

wass3r:
this would be really nice for local pipeline execution via `vela exec pipeline` - even without checking code in. but i think the current implementation doesn't suit itself for that. part of the reason i thought this can be scm agnostic and the type be `relative` or `file`. thoughts around that?

cognifloyd:
> this would be really nice for local pipeline execution via `vela exec pipeline` - even without checking code in. but i think the current implementation doesn't suit itself for that. part of the reason i thought this can be scm agnostic and the type be `relative` or `file`. thoughts around that?

Is that what `c.local` bit does?

https://github.com/go-vela/server/blob/56932446fdc545d44162da499f7318ad34994c77/compiler/native/expand.go#L170-L179

I kind of like `relative` - but that might suggest that it is relative to the current file vs relative to the repo root.
I think I could grow to like `file` even better. The more I think about this one, I think I like it a lot.

wass3r:
> Is that what c.local bit does?

busted! forgot about that. 🫣

wass3r (approved):

wass3r:
@cognifloyd i think your feedback is addressed

@ecrupper we will need to add to docs too

cognifloyd (approved):

PR #720, opened 2022-10-07
feat(templates): allow for compile-time rulesets for template steps

ecrupper:
Closes https://github.com/go-vela/community/issues/547

A couple things to note about this PR:

1) It does not account for runtime rules, such as `continue` and `status`. The reason being: templates are compiled and merged into the parent pipeline, and there is no way of referring to a template-level ruleset once that has happened.

2) This PR struck me as the only way to accommodate rulesets at a template level without re-hauling a bit. Another idea would be to create a `template` field for each `pipeline.Build` which carries the slice of steps (or perhaps references to them) as well as the ruleset. In other words: a lot of work.

That said, I recognize this solution is a bit hacky, and if we'd rather come to a conclusion on whether or not we want this feature in the first place (and if so, probably work on a proposal to establish a design), that would be a viable outcome for this PR! Let me know your thoughts!


cognifloyd (approved):
A little hacky, yes. But I think that's ok here. Let's get the functionality and then figure out if we need a more extensive refactor.

jbrockopp (approved):
LGTM

plyr4 (approved):

wass3rw3rk (commented):
👍🏼

PR #730, opened 2022-11-04
feat(events): add release webhook event

Gradsta:
This is one in a series of PRs that enables Vela users to define pipeline steps for GitHub release events. This feature was originally requested in [#518](https://github.com/go-vela/community/issues/518) in the community repository. 

While a release requires a tag, the opposite cannot be said--tags are independent events and can exist without the presence of a release. As such, it can be useful to distinguish between the different event types. For my particular use case, I'd like to use a Gradle plugin that uses tags to automatically version my repository. In this setup, merging a PR to the main branch would serve as the trigger to increment the version (i.e., tagging the applicable commit) and deploy to stage. Creating a release and selecting the newly-created tag would serve as the trigger to deploy to production. This requires that the release event type be supported by Vela.

### Related PRs

* [types](https://github.com/go-vela/types/pull/272)
* [cli](https://github.com/go-vela/cli/pull/394)
* [ui](https://github.com/go-vela/ui/pull/596)


chrispdriscoll:
Apparently we only followed up in the UI PR. Apologies for not updating all PRs. Please see UI PR for our updates and decision to close. Happy to re-engage if you want to discuss.

https://github.com/go-vela/ui/pull/596#issuecomment-1802092643

PR #754, opened 2023-01-06
refactor(api): move repo logic to separate package

jbrockopp:
Based off of https://github.com/go-vela/server/pull/615

This change continues the refactor efforts initially introduced in the above PR.

This adds a new `repo` package to the `github.com/go-vela/server/api` package.

Also, this change skips checking a user's access for orgs, repos, and teams if that resource is in their personal org.

i.e. Vela no longer check's my access level for those resources when interacting with https://github.com/jbrockopp

This will help improve the user experience when developing locally and reduce unneeded calls to the SCM.


api/repo/update.go
```diff
@@ -0,0 +1,255 @@
...
+//       "$ref": "#/definitions/Error"
+//   '503':
+//     description: Unable to update the repo
+//     schema:
+//       "$ref": "#/definitions/Error"
+
+// UpdateRepo represents the API handler to update
+// a repo in the configured backend.
+//
+//nolint:funlen // ignore line length
```
cognifloyd, line 72: This is just a copy/paste from where it was before.

It might be nice to fix this comment however, as it doesn't seem right. This is ignoring the function length, not the line length, isn't it?

cognifloyd (approved):
One minor nit, but looks good to me anyway. It's nice to see all of this moving into separate modules with shorter files. So much nicer.

api/repo/update.go
jbrockopp: fixed 👍

cognifloyd (dismissed):

api/repo/create.go
```diff
@@ -0,0 +1,327 @@
...
+		util.HandleError(c, http.StatusBadRequest, retErr)
+
+		return
+	}
+
+	// update fields in repo object
+	r.SetUserID(u.GetID())
+
+	// set the active field based off the input provided
+	if input.Active == nil {
```
ecrupper, line 114: Is there a reason we don't use the getter here?

api/repo/repair.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		retErr := fmt.Errorf("unable to delete webhook for %s: %w", r.GetFullName(), err)
+
+		util.HandleError(c, http.StatusInternalServerError, retErr)
+
+		return
+	}
+
+	// check if we should create the webhook
+	if c.Value("webhookvalidation").(bool) {
```
ecrupper, line 79: This isn't an issue, but curious why this webhook validation check isn't for both SCM calls here (disable and enable)?

scm/github/access.go
```diff
@@ -51,6 +56,17 @@ func (c *client) RepoAccess(u *library.User, token, org, repo string) (string, e
 		"user": u.GetName(),
 	}).Tracef("capturing %s access level to repo %s/%s", u.GetName(), org, repo)
 
+	// check if user is accessing repo in personal org
+	if strings.EqualFold(org, *u.Name) {
```
ecrupper, line 60: Any reason we don't use the getter here?

ecrupper (approved):
A couple questions out of curiosity, but otherwise LGTM — I like the refactor

api/repo/create.go
ecrupper: Nvm, not using a getter makes sense in this context

api/repo/repair.go
jbrockopp: fixed 👍

jbrockopp:
@cognifloyd @ecrupper ready for review again 👍

cognifloyd (approved):

ecrupper (approved):

PR #755, opened 2023-01-20
feat(clone-image): make clone image configurable

wass3r:
closes https://github.com/go-vela/community/issues/456

adds a `VELA_CLONE_IMAGE` configuration to the server that allows a Vela administrator to define the default clone image to use for the Vela injected clone step.


wass3r:
also readied https://github.com/go-vela/worker/pull/417 to fix broken tests once this is sourced in worker code.

docs PR incoming as well

wass3r:
docs draft PR opened

plyr4 (approved):

jbrockopp (changes requested):
> NOTE: Thanks for this! Code LGTM so requesting changes for visibility purposes (i.e. Trailer.io)

Do we think we need to "validate" the value provided as the clone image?

I know we already ensure a value is provided which is a good thing 👍 

But I'm wondering if we want to "force" people to use `target/vela-git`?

i.e. `if !strings.Contains(c.String("clone-image"), "target/vela-git") { return err }`

I'm leaning towards NOT forcing and leaving this PR as-is.

However, I thought it was worth opening the discussion for it.

wass3r:
@jbrockopp forgot about the comment on this. i'm also leaning towards not doing that. as suggested, you could trivially bypass by using `not-target/vela-git`. is the thought to limit usage to some sort of "authentic" clone image? i can see doing something with signed images and checking that potentially, if that's the goal. however, not sure if there's a ton of value in limiting flexibility here.

jbrockopp:
@wass3r 

> is the thought to limit usage to some sort of "authentic" clone image?

Yeah, thinking it would be a security enhancement to ensure a "valid" clone image is used.

That being said, I worry it would limit organizations from creating their own customized clone image.

Let's plan to roll forward as is and we can always consider it in the future 👍

jbrockopp (approved):
LGTM

plyr4 (approved):

PR #765, opened 2023-02-21
feat(auth): build token implementation

ecrupper:
Server PR for Epic 1: Build Tokens.

Changes:

- Conversion of `router/middleware/token` package to `router/middleware/auth`, which will now only retrieve auth headers from http requests, rather than generate/parse tokens
- Creation of `internal/token` package, which features a `manager` that is set up in the context using various new flags added to `main.go`. In brief, it's a one-stop-shop for token minting and validating and offers configurability to a degree.
- Additional claims added to `Claims` struct, including information about the token type as well as relevant fields for that token (build tokens have build ids and repo).
- Removal of global permissions for agents
- Worker endpoints used for check in (POST / PUT) have been put behind `MustWorker()` permissions check, which validates the symmetric token shared between server and worker
- Build/Step/Service/Log update endpoints have been put behind `MustBuildAccess()` permissions check, which validates the build token, which is provided as a unique auth scope for each build.
- Add GET endpoint for retrieving a build token, which validates the build's status and creates a build token
- Addition of a private key, which will be seeded in the server environment, much like the secret, which will be used to sign and validate all tokens.
- Updates made to `MustSecretAdmin` which will scope requests with build token auth based on the repo within the token
- Claims will be established in the context for baseAPI endpoints before the user

depends on: https://github.com/go-vela/types/pull/276


.github/workflows/publish.yml
```diff
@@ -1,52 +0,0 @@
-# name of the action
```
plyr4, line 1: revert deletion?

wass3r (approved):

wass3r (dismissed):

internal/token/compose.go
```diff
@@ -0,0 +1,75 @@
...
+	"net/http"
+	"net/url"
+
+	"github.com/gin-gonic/gin"
+	"github.com/go-vela/types"
+	"github.com/go-vela/types/constants"
+	"github.com/go-vela/types/library"
+)
+
+// Compose generates an refresh and access token pair unique
```
plyr4, line 17:
```suggestion
// Compose generates a refresh and access token pair unique
```

plyr4 (changes requested):

internal/token/compose.go
ecrupper: Nice catch!!

wass3r (approved):

jbrockopp (changes requested):
proposal?

ecrupper:
> proposal?

https://github.com/go-vela/community/pull/755

wass3r (approved):

plyr4 (approved):

timhuynh94 (approved):

cognifloyd (approved):

PR #772, opened 2023-02-27
feat(worker-visibility): extend worker table with 5 fields

KellyMerrick:
Worker Visibility Proposal: https://github.com/go-vela/community/pull/813

dependent on https://github.com/go-vela/types/pull/277
needed for https://github.com/go-vela/worker/pull/482


api/worker.go
```diff
@@ -0,0 +1,539 @@
+// Copyright (c) 2022 Target Brands, Inc. All rights reserved.
```
ecrupper, line 1: I believe this file was added by mistake?

database/worker/index_test.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/DATA-DOG/go-sqlmock"
 )
 
-func TestWorker_Engine_CreateWorkerIndexes(t *testing.T) {
+func Indexes(t *testing.T) {
```
ecrupper, line 13: Can we keep the testing name structure like we do [here](https://github.com/go-vela/server/blob/e06d557dd12cb1c6477b7403d7815f091cf25e51/database/build/index_test.go#L13), [here](https://github.com/go-vela/server/blob/e06d557dd12cb1c6477b7403d7815f091cf25e51/database/repo/index_test.go#L13), and [here](https://github.com/go-vela/server/blob/e06d557dd12cb1c6477b7403d7815f091cf25e51/database/pipeline/index_test.go#L13)?

router/middleware/perm/perm.go
```diff
@@ -100,6 +100,15 @@ func MustWorkerAuthToken() gin.HandlerFunc {
...
 
+		// global permissions bypass
+		if cl.IsAdmin {
+			logrus.WithFields(logrus.Fields{
+				"user": cl.Subject,
+			}).Debugf("user %s has platform admin permissions", cl.Subject)
+
+			return
+		}
+
```
ecrupper, line 111: With the introduction of this platform admin bypass, I wonder if we should restrict the fields that they can change (such as running build IDs)?

api/metrics.go
```diff
@@ -401,6 +440,24 @@ func recordGauges(c *gin.Context) {
 			}
 		}
 
+		// idle, available, busy, error counts
+		for _, worker := range workers {
```
ecrupper, line 444: Could this entire loop be brought into the already existing loop above? Under https://github.com/go-vela/server/pull/772/files#diff-ffb09227a043a8b68ade16603539951b8e216f71125de14b6919ab63fee86bc5R433-R435 perhaps?

ecrupper (changes requested):

api/worker.go
timhuynh94: I think it was added by mistake!

router/middleware/perm/perm.go
timhuynh94: @ecrupper can you explain a little bit more on this?

ecrupper: I'm imagining a future case where a platform admin may want to live-update the worker's routes. In that case, because of [this check](https://github.com/go-vela/server/pull/772/files#diff-c760a1133ea01608d485599741679f949ff6f09ff14ba119fb15f67d81fecde4R101-R104) (which will return as `true` because `GetRunningBuildIDs` returns `[]string{}` if it is `nil` rather than `nil`), then the platform admin will also have to know the running build IDs to make this update.

Steps to reproduce:
1. fire up local stack on this branch
2. set user to platform admin
3. manually perform DB update to worker, setting running build ids to an example value
4. call PUT worker as platform admin with the body as: 
```json
{
	"id": 1,
	"hostname": "worker",
	"address": "http://worker:8080",
	"routes": [
		"vela",
		"highway"
	]
}
```
5. worker in DB will now have a running_build_ids value of empty. This is with the `latest` worker version, so no other updates have been made behind the scenes.

So either the platform admin shouldn't even be allowed to update values such as `running_build_ids`, or there should be a more scrutinizing check in the PUT call, which will not accidentally set values on behalf of the platform admin. I have no opinion on which of those two solutions is better, or if there's another option.

KellyMerrick: Addressed by not requiring a running build list in the body.

ecrupper (approved):

timhuynh94 (approved):

plyr4 (approved):

PR #776, opened 2023-02-28
fix(perm): allow workers with build tokens to access MustRead() for private visibility repos

ecrupper:
Initial [build token implementation](https://github.com/go-vela/server/pull/765) failed to address repos with a private visibility status. Since the worker is not treated as a platform admin, it will fail the SCM check to verify read permissions. 

This fix adds logic to allow workers to access `MustRead` endpoints from private repos so long as the build token is still accurate.  I also added a related test.

I double checked the worker's flow and determined that it does not call endpoints behind `MustWrite` or `MustAdmin`, so I've decided to leave the build token logic out of those functions.

### DIFF
--- a/router/middleware/perm/perm.go
+++ b/router/middleware/perm/perm.go
@@ -444,6 +444,12 @@ func MustRead() gin.HandlerFunc {
 			if cl.BuildID == b.GetID() {
 				return
 			}
+
+			retErr := fmt.Errorf("user %s does not have 'read' permissions for repo %s", u.GetName(), r.GetFullName())
+
+			util.HandleError(c, http.StatusUnauthorized, retErr)
+
+			return
 		}
 
 		logger.Debugf("verifying user %s has 'read' permissions for repo %s", u.GetName(), r.GetFullName())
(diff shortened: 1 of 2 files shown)


router/middleware/perm/perm.go
```diff
@@ -437,8 +438,23 @@ func MustRead() gin.HandlerFunc {
 			return
 		}
 
+		// return if request is from worker with build token access
+		if strings.EqualFold(cl.TokenType, constants.WorkerBuildTokenType) {
+			b := build.Retrieve(c)
```
wass3r, line 443: MustRead is sort of resource agnostic. `org` and `repo` are typically present on any endpoint today, especially one that needs perm check. this addition assumes we are on a route that did `build.Establish()`. should there be additional checks or a different solution to account for this?

ecrupper: It only assumes we're on a route that did `build.Establish()` if the token type is `WorkerBuild`, which is an accurate assumption to make. If the build token is used on an endpoint that does not include the `build.Establish()` piece, then it will fail without panic. 

E.g. tried using a build token for GET `api/v1/repos/foo/bar` and was met with 

```
subject vela-worker does not have 'read' permissions for repo foo/bar
```

Could be a more descriptive error, but at the very least it does not panic

plyr4 (approved):

router/middleware/perm/perm.go
wass3r: Agreed on the error message. My concern was two-fold. One was the thought that build would provide `nil`, but since we're using getters we should be shielded from not `nil` checking here. The other concern was more about stuffing route/token specific logic in these, otherwise, route and token agnostic methods. I can see that getting unwieldy if it gets out of hand.

PR #779, opened 2023-03-03
feat: Make InitStep a first-class-citizen to add user-facing reporting around api / compiler

cognifloyd:
Currently, we're abusing the `Step` and `Container` models to allow reporting on build setup (eg when Docker initializes the Network and Volumes). This means that we have to check for the special "init" stage/step/container in many places. This makes "Init" reporting a first-class-citizen, instead of shoe-horning it into steps and containers (which was admittedly an excellent MVP).

An `InitStep` is a report by a "reporter" about a discrete part of the build setup. The "reporter" would be a logical part of the stack, for example "Pipeline Compiler", "Docker Runtime", "Kubernetes Runtime", ... For example, we could have these reporters be discrete initsteps:
- Pipeline Compiler: report info/debug logs
- Docker Runtime: report network setup
- Docker Runtime: report volume setup
- Kubernetes Runtime: Pod YAML

Similar to a `Step`, where the log data is stored in a separate `Log` struct/table, the `InitStep` is also associated with a `Log` that stores the actual log/report. The `InitStep` however can also specify a mimetype for the output, which is meant for eventual consumption by the UI to do syntax highlighting if relevant.

So, this adds:
- `database/initstep`
- `router/initstep`, `router/middleware/initstep`, `api/initstep`, `mock/server/initstep`
- InitStep support in `database/log`
- InitStep support in `router/log`, `router/middleware/log`, `api/log`, `mock/server/log`

And then I also added example logs in and around the compiler (in `api/webhook` and `api/build`). Not everything that compiles a template has a build it can associate a log with, so logs get created only after the build is ready.


cognifloyd:
Closed with https://github.com/go-vela/community/pull/771

PR #781, opened 2023-03-03
feat(compiler): Enables Starlark structs

colindean:
This simple addition enables the `struct` function in Starlark, which simplifies Starlark code by enabling a convenient dot-notation to a dictionary-like collection instead of requiring verbose array access syntax.

starlark-go's enablement instructions:

https://pkg.go.dev/go.starlark.net/starlarkstruct#Make

Bazel's docs on struct:

https://bazel.build/rules/lib/struct#struct


wass3rw3rk (changes requested):
can you go run `make test` or `go mod tidy`

colindean:
`make test` passed, `go mod tidy` in 203953b, `make test` passed after.

wass3rw3rk:
hah, i meant `make clean` (which will run `go mod tidy` as well), but thanks for the quick follow-up. looks like GitHub Actions is having an issue here. I'll wait for a bit but probably have to kick that off again.

plyr4 (approved):
lgtm

colindean:
Yay!

PR #790, opened 2023-03-15
feat(auth): server side changes for new worker auth flow

ecrupper:
Ref: https://github.com/go-vela/community/pull/770

Dependent on: https://github.com/go-vela/types/pull/283


plyr4 (changes requested):
needs server mocks for admin/register-token (probably not validate though)
correction: validate needs a mocked endpoint but probably not a mock response json

api/worker.go
```diff
@@ -272,20 +318,47 @@ func UpdateWorker(c *gin.Context) {
 		w.SetLastCheckedIn(input.GetLastCheckedIn())
 	}
 
-	// send API call to update the worker
-	err = database.FromContext(c).UpdateWorker(w)
```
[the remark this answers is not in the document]
ecrupper: :0 that was definitely not intentional

cognifloyd (commented):
That's closer. I see one more issue with UpdateWorker, and then I can approve.

api/worker.go
```diff
@@ -285,7 +335,50 @@ func UpdateWorker(c *gin.Context) {
...
+	switch cl.TokenType {
+	// if update performed by platform admin user, do not send back token
+	case constants.UserAccessTokenType:
+		c.JSON(http.StatusOK, w)
+	// if symmetric token configured, send back symmetric token
+	case constants.ServerWorkerTokenType:
+		if secret, ok := c.Value("secret").(string); ok {
+			tkn := new(library.Token)
+			tkn.SetToken(secret)
+			c.JSON(http.StatusOK, WorkerCheckIn{Worker: w, Token: tkn})
```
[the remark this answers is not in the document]
cognifloyd: Also, sdk-go is going to need to know about both types that can be returned. Or, maybe the worker needs a different end point to check in and get the new token.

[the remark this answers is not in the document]
ecrupper: The more I think about this, the more I think it should just be a separate endpoint altogether

cognifloyd (dismissed):

api/worker.go
```diff
@@ -288,6 +331,95 @@ func UpdateWorker(c *gin.Context) {
...
+//     schema:
+//       "$ref": "#/definitions/Error"
+//   '500':
+//     description: Unable to refresh worker auth
+//     schema:
+//       "$ref": "#/definitions/Error"
+
+// RefreshWorkerAuth represents the API handler to
+// refresh the auth token for a worker.
+func RefreshWorkerAuth(c *gin.Context) {
```
cognifloyd, line 369: I think this should also update the LastCheckedIn time on the worker.

api/worker.go
```diff
@@ -288,6 +331,95 @@ func UpdateWorker(c *gin.Context) {
...
+			retErr := fmt.Errorf("unable to generate auth token for worker %s: %w", w.GetHostname(), err)
+
+			util.HandleError(c, http.StatusInternalServerError, retErr)
+
+			return
+		}
+
+		tkn.SetToken(wt)
+
+		c.JSON(http.StatusCreated, tkn)
```
cognifloyd, line 419: This is 201. Should it be 200? Or update the swagger comment?

plyr4 (changes requested):

mock/server/server.go
```diff
@@ -128,12 +129,14 @@ func FakeHandler() http.Handler {
...
 	e.POST("/api/v1/workers", addWorker)
 	e.PUT("/api/v1/workers/:worker", updateWorker)
+	e.POST("/api/v1/workers/:worker/refresh", refreshWorkerAuth)
 	e.DELETE("/api/v1/workers/:worker", removeWorker)
 
 	// mock endpoints for authentication calls
 	e.GET("/token-refresh", getTokenRefresh)
 	e.GET("/authenticate", getAuthenticate)
 	e.POST("/authenticate/token", getAuthenticateFromToken)
+	e.POST("/validate-token", validateToken)
```
plyr4, line 139: should match the handler
```suggestion
	e.GET("/validate-token", validateToken)
```

plyr4 (changes requested):

mock/server/authentication.go
```diff
@@ -73,3 +73,17 @@ func getAuthenticateFromToken(c *gin.Context) {
...
 	c.JSON(http.StatusOK, body)
 }
+
+// validateToken returns mock response for a http GET.
+//
+// Don't pass "Token" in header to receive an error message. Pass "0" in "Token" header to receive invalid token resp.
+func validateToken(c *gin.Context) {
+	err := "error"
+
+	token := c.Request.Header.Get("Token")
```
plyr4, line 83: should this mock be checking the "Authorization" header rather than "Token"? 

https://github.com/go-vela/server/pull/790/files#diff-150d061ec4206a8396c1e267a3493c43049fb93cca6b50cb99cf4fa8c1adc2e7R93

api/token.go
```diff
@@ -65,3 +67,38 @@ func RefreshAccessToken(c *gin.Context) {
...
+// ---
+// produces:
+// - application/json
+// security:
+//   - CookieAuth: []
+// responses:
+//   '200':
+//     description: Successfully validated a token
+//     schema:
+//       type: string
```
wass3rw3rk, line 84: question: i don't believe this returns a string on success?

router/middleware/executors/executors.go
```diff
@@ -51,8 +53,26 @@ func Establish() gin.HandlerFunc {
...
 
-		// add the token to authenticate to the worker as a header
-		req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", c.MustGet("secret").(string)))
+		tm := c.MustGet("token-manager").(*token.Manager)
+
+		// set mint token options
+		mto := &token.MintTokenOpts{
+			Hostname:      "vela-server",
+			TokenType:     constants.WorkerAuthTokenType,
+			TokenDuration: time.Minute * 1,
```
wass3rw3rk, line 62: question: should we turn the vela-server token duration into a constant somewhere or since we create this type of token in a couple of places, create a method somewhere?

wass3rw3rk: we'll come back to this if we add more code like this 👍🏼

mock/server/worker.go
```diff
@@ -31,6 +31,23 @@ const (
...
+				"address": "http://vela:8080",
+				"routes": [
+				"large",
+				"docker",
+				"large:docker"
+				],
+				"active": true,
+				"last_checked_in": 1602612590
+			}
+		}`
```
plyr4, line 49:
```suggestion
	UpdateWorkerResp = `
		{
			"id": 1,
			"hostname": "worker_1",
			"address": "http://vela:8080",
			"routes": [
				"large",
				"docker",
				"large:docker"
			],
			"active": true,
			"last_checked_in": 1602612590
		}`
```

plyr4: needs the wrapper field removed.
i might have messed up the whitespace in there via, it needs to be double checked in an editor

plyr4 (changes requested):

wass3rw3rk (approved):

wass3rw3rk (approved):

plyr4 (approved):

PR #793, opened 2023-03-17
fix(compiler)!: webhook payload containing message with special characters causes failure

NickHackman:
Use the implementation made upstream in https://github.com/drone/envsubst/pull/27 to fix escape sequence handling to prevent yaml parsing to fail due to invalid escape sequences.

Issue: https://github.com/go-vela/community/issues/702


go.mod
```diff
@@ -53,6 +53,7 @@ require (
 	github.com/chenzhuoyu/base64x v0.0.0-20221115062448-fe3a3abad311 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.2 // indirect
 	github.com/dgryski/go-rendezvous v0.0.0-20200823014737-9f7001d12a5f // indirect
+	github.com/drone/envsubst v1.0.3 // indirect
```
NickHackman, line 56: This is an indirect dependency

```shell
❯ go mod why github.com/drone/envsubst
# github.com/drone/envsubst
github.com/go-vela/server/api
github.com/go-vela/types/pipeline
github.com/drone/envsubst
```

Going to update this dependency in `types/pipeline` as well

NickHackman: PR to update `types/pipeline` https://github.com/go-vela/types/pull/284

cognifloyd (dismissed):

NickHackman:
"full-review" failed due to other lint checks done by golangci-lint

jbrockopp (approved):
LGTM

plyr4 (dismissed):

NickHackman:
@plyr4 good call out, this would impact users who have tried to hack around the issue in the first place. Hopefully that's a very small subset of users who need remediation here

ecrupper:
@NickHackman Hi Nick! Thanks for the PR. I pulled down both the changes in this PR as well as your changes in the [types PR](https://github.com/go-vela/types/pull/284). I had a few buggy payloads to test, and it does indeed look like the escape character error is resolved, however some new YAML parser error arose, such as `mapping values not allowed in this context`. While I do think we should probably update this library, I'm wondering if there's more to this issue + perhaps another way to solve? Perhaps some sanitization on our end? I will update the issue with a payload I tested that got another parser error. Curious what others think in terms of whether we merge this in as a "step in the right direction" or if we prefer a catch-all sanitization solution?

NickHackman:
@ecrupper Awesome, fire away with examples here. I assume any issue we face is coming from the upstream library. I'll do my best to patch things if we can get cases that fail.

Ideally these issues are all fixed in the upstream library. Vela shouldn't need to patch it, but whatever is seen as best

ecrupper (approved):

wass3rw3rk (approved):

ecrupper:
Will update the original issue with anything else we find related to compiler errors from weird character sequences. For now, we know this fixes a few things and doesn't break anything from what we can tell. Merging! Thanks again @NickHackman !

plyr4 (approved):

PR #794, opened 2023-03-22
feat(templates): add ability to call templates within a template

ecrupper:
Closes https://github.com/go-vela/community/issues/256

Using a configurable max depth, this change would allow Vela users to call templates within templates.

Note: since `compileInline` doesn't call `ExpandSteps` and my first few passes on getting it to work similarly failed, I left support for `render_inline` as TO-DO.

### DIFF
--- a/cmd/vela-server/main.go
+++ b/cmd/vela-server/main.go
@@ -185,6 +185,12 @@ func main() {
 			Usage:   "modification retries, used by compiler, number of http requires that the modification http request will fail after",
 			Value:   5,
 		},
+		&cli.IntFlag{
+			EnvVars: []string{"VELA_MAX_TEMPLATE_DEPTH", "MAX_TEMPLATE_DEPTH"},
+			Name:    "max-template-depth",
+			Usage:   "max template depth, used by compiler, maximum number of templates that can be called in a template chain. Prevents circular depenendcies",
+			Value:   5,
+		},
 		&cli.DurationFlag{
 			EnvVars: []string{"VELA_WORKER_ACTIVE_INTERVAL", "WORKER_ACTIVE_INTERVAL"},
 			Name:    "worker-active-interval",
--- a/compiler/native/compile.go
+++ b/compiler/native/compile.go
@@ -82,7 +82,7 @@ func (c *client) Compile(v interface{}) (*pipeline.Build, *library.Pipeline, err
 
 	switch {
 	case p.Metadata.RenderInline:
-		newPipeline, err := c.compileInline(p, nil, c.TemplateDepth)
+		newPipeline, err := c.compileInline(p, nil)
 		if err != nil {
 			return nil, _pipeline, err
 		}
@@ -117,7 +117,7 @@ func (c *client) CompileLite(v interface{}, template, substitute bool, localTemp
 	_pipeline.SetType(c.repo.GetPipelineType())
 
 	if p.Metadata.RenderInline {
-		newPipeline, err := c.compileInline(p, localTemplates, c.TemplateDepth)
+		newPipeline, err := c.compileInline(p, localTemplates)
 		if err != nil {
 			return nil, _pipeline, err
 		}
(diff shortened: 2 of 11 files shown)


JordanSussman:
We may want to consider adding some caching for the [getTemplate](https://github.com/go-vela/server/blob/main/compiler/native/expand.go#L164) function to avoid fetching the same templates multiple times.

ecrupper:
> We may want to consider adding some caching for the [getTemplate](https://github.com/go-vela/server/blob/main/compiler/native/expand.go#L164) function to avoid fetching the same templates multiple times.
> 
> EDIT: the extra compile duration caused by these changes could be concerning without caching since GitHub expects a response to webhooks [within 10 seconds for github.com](https://docs.github.com/en/rest/guides/best-practices-for-integrators?apiVersion=2022-11-28#favor-asynchronous-work-over-synchronous) or [30 seconds for GHES](https://docs.github.com/en/enterprise-server@3.7/rest/guides/best-practices-for-integrators#favor-asynchronous-work-over-synchronous).

I really like the idea of caching templates. I know we have at least a few templates that have widespread usage and definitely would take up permanent residence in that cache. 

As for the extra compile duration, that is definitely something to consider. In my testing the responses were rather quick, but since template depth is renewed for each individual step, it could definitely get messy with dozens of steps each calling a chain of templates. It would be worth investigating some of those recommendations in the links related to queuing up work and giving the SCM a quick response...

cmd/vela-server/main.go
```diff
@@ -197,6 +197,12 @@ func main() {
 			Usage:   "modification retries, used by compiler, number of http requires that the modification http request will fail after",
 			Value:   5,
 		},
+		&cli.IntFlag{
+			EnvVars: []string{"VELA_MAX_TEMPLATE_DEPTH", "MAX_TEMPLATE_DEPTH"},
+			Name:    "max-template-depth",
+			Usage:   "max template depth, used by compiler, maximum number of templates that can be called in a template chain. Prevents circular depenendcies",
```
timhuynh94, line 203: just a typo for dependencies

ecrupper:
Putting in draft for now. Let's reopen this discussion after `v0.19` is released?

ecrupper:
Bringing this out of draft

ecrupper:
putting this back in draft, as I decided to tackle `render_inline` instead of leaving it as a to-do 👍

compiler/native/compile.go
```diff
@@ -276,12 +291,6 @@ func (c *client) compileInline(p *yaml.Build, localTemplates []string) (*yaml.Bu
 		}
 	}
 
-	// validate the yaml configuration
-	err := c.Validate(&newPipeline)
-	if err != nil {
-		return nil, err
-	}
-
```
wass3rw3rk, line 284: was this intentional?

compiler/native/expand.go
```diff
@@ -47,11 +47,18 @@ func (c *client) ExpandStages(s *yaml.Build, tmpls map[string]*yaml.Template, r
 
 // ExpandSteps injects the template for each
 // templated step in a yaml configuration.
-func (c *client) ExpandSteps(s *yaml.Build, tmpls map[string]*yaml.Template, r *pipeline.RuleData) (*yaml.Build, error) {
+func (c *client) ExpandSteps(s *yaml.Build, tmpls map[string]*yaml.Template, r *pipeline.RuleData, depth int) (*yaml.Build, error) {
 	if len(tmpls) == 0 {
 		return s, nil
 	}
 
+	// if max template depth has been reached, return to avoid circular dependencies
```
wass3rw3rk, line 55:
```suggestion
	// return if max template depth has been reached
```
minor, but it makes it sound like it's exclusively about circular dependencies.

wass3rw3rk (changes requested):
bunch of minor things and a couple of questions.

should we validate the value that was provided for `max-template-depth`?

compiler/native/compile.go
ecrupper: Yes, I saw that both calls to `compileInLine` validate the result: https://github.com/go-vela/server/blob/5a9a92d279e058b75b37e589e1eef2ec63f550a6/compiler/native/compile.go#L89-L93 and https://github.com/go-vela/server/blob/5a9a92d279e058b75b37e589e1eef2ec63f550a6/compiler/native/compile.go#L124-L128. And I want to validate whatever is finally returned by that function rather than fragments as the function rolls back up.

wass3rw3rk (approved):
nice work - would like to see docs pr as well

plyr4 (approved):
nice, i like the tests

PR #795, opened 2023-03-24
refactor(api): moved common `Post*Stream` logic to helper func 

cognifloyd:
`PostServiceStream` and `PostStepStream` have a lot of shared logic to manage streaming logs to the database.

This refactors the `Post*Stream` methods so they can share an implementation in a new `streamLogToDatabase` helper function.

This is pre-work for the InitStep proposal https://github.com/go-vela/community/pull/771, but this is separate and useful whether or not that proposal gets accepted. If the proposal is accepted, this will make it easier to add `PostInitStepStream` in a future PR without adding yet another copy of the shared log streaming logic. If the propsal is not accepted, DRYing this code could help with any performance refactors so that we have one place to update instead of 2 (or 3). And it makes the logic within the API method much easier to follow.

### DIFF
--- a/api/stream.go
+++ b/api/stream.go
@@ -116,7 +116,7 @@ func PostServiceStream(c *gin.Context) {
 		return
 	}
 
-	streamLog(c, logger, _log, "service", entry, r.GetTimeout())
+	streamLogToDatabase(c, logger, _log, "service", entry, r.GetTimeout())
 
 	c.JSON(http.StatusNoContent, nil)
 }
@@ -210,13 +210,13 @@ func PostStepStream(c *gin.Context) {
 		return
 	}
 
-	streamLog(c, logger, _log, "step", entry, r.GetTimeout())
+	streamLogToDatabase(c, logger, _log, "step", entry, r.GetTimeout())
 
 	c.JSON(http.StatusNoContent, nil)
 }
 
-// streamLog handles streaming logs to the database.
-func streamLog(c *gin.Context, logger *logrus.Entry, log *library.Log, entryType, entry string, timeout int64) {
+// streamLogToDatabase handles streaming logs to the database.
+func streamLogToDatabase(c *gin.Context, logger *logrus.Entry, log *library.Log, entryType, entry string, timeout int64) {
 	// create new buffer for uploading logs
 	logs := new(bytes.Buffer)
 	// create new channel for processing logs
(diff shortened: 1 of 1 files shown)


api/stream.go
```diff
@@ -97,22 +97,15 @@ func PostServiceStream(c *gin.Context) {
 	//
 	// https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Entry.WithFields
 	logger := logrus.WithFields(logrus.Fields{
-		"build":   b.GetNumber(),
 		"org":     o,
 		"repo":    r.GetName(),
+		"build":   b.GetNumber(),
```
cognifloyd, line 102: Reordering these doesn't really matter.

It helps me to see them in api path order so that I can make sure I didn't forget one: `.../repos/:org/:repo/build/:build/...`

I guess it is in alphabetical order now, so I can revert this if alphabetical order is preferable.

wass3rw3rk (approved):
lgtm

jbrockopp (changes requested):
Do we think we should remove this endpoints instead of refactoring?

I ask because AFAIK, these aren't being used anymore:

https://github.com/go-vela/worker/pull/233

cognifloyd:
Removing them makes sense. Maybe at some point we can add an endpoint that does one of:
- use an alternative protocol that any load balancers are less likely to interfere with such as websocket or grpc
- make sure the endpoint gracefully handles reconnects (allow append-only streaming)

Shall I remove this? Removing achieves my goal of simplifying what I have to implement to add the InitStep stuff.

jbrockopp:
> use an alternative protocol that any load balancers are less likely to interfere with such as websocket or grpc

Unfortunately, those implementations can still be susceptible to issues based on load balancer configuration:

https://github.com/go-vela/community/issues/427#issuecomment-958112554

Not to say we couldn't work through them though.

IIRC, we intentionally decided to pivot away from them due to the fragility that is added by them.

> Shall I remove this?

That would be my preference since they are no longer used

wass3rw3rk:
i made a reference to that here: https://github.com/go-vela/server/pull/796#issuecomment-1483009308

perhaps if the storage provide was something other than postgres, something like this would be more feasible.

i'm ok to remove for now.

cognifloyd:
PRs to remove the endpoints:
- https://github.com/go-vela/server/pull/799
- https://github.com/go-vela/sdk-go/pull/215

PR #796, opened 2023-03-24
chore: Cleanup endpoints

cognifloyd:
There are a variety of endpoints that have been removed or added (or never existed!). This PR adjusts the following to account for those differences:
- server mock
- router doc comments

The server mock updates are a follow-up for:
- #734

The router doc comment updates are a follow-up for:
- #489
- #765

From git spelunking, it looks like the removed `/api/v1/users/:user/...` endpoints never existed.


wass3rw3rk (approved):
thanks for the cleanup!

cognifloyd:
Not in use, but they are supported. Were the `/stream` endpoints used by the worker in the past, or were they designed for future use by the worker?

Something that streams the logs to the database sounds like a great idea to me. - the implementation probably needs work like in #795.

wass3rw3rk:
fair point. i think the idea was to, more or less, live stream with an open connection some day. the implementation details likely have to change. i haven't looked at this in a while though.

cognifloyd:
Rebased and dropped comments about endpoints removed in #799

wass3r (approved):

jbrockopp (approved):
LGTM

PR #800, opened 2023-03-27
enhance(api/scm): expand sync to align SCM subscribed events with allowed events

ecrupper:
With [this change](https://github.com/go-vela/server/pull/679), Vela will always align the SCM webhook configuration for listened-events with what is allowed as an event in the Vela repo settings. However, this will not apply with repos enabled prior to the release of that change unless there is an update to the repo. 

Expanding `sync` seemed like a viable way to accommodate this scenario without clumsily asking users to make a fake update to their repos. Let me know if you think it should be a separate endpoint altogether.


cognifloyd (approved):

KellyMerrick (approved):
lgtm

PR #802, opened 2023-03-31
enhance(metrics): gather queued builds

ecrupper:
There have been several cases where a build has found itself in a `pending` state due to some error in the webhook flow (typically around `publishToQueue`). These builds never make it to the queue and are still left as `pending`. This additional query to the metrics endpoint gives a much more accurate representation of queue size.


wass3rw3rk (approved):

KellyMerrick (approved):
lgtm

jbrockopp (approved):
LGTM

PR #782, opened 2023-03-04
refactor(database): move secret logic into separate package

jbrockopp:
Based off of https://github.com/go-vela/server/pull/574, https://github.com/go-vela/server/pull/663, https://github.com/go-vela/server/pull/687, https://github.com/go-vela/server/pull/692 https://github.com/go-vela/server/pull/721 and https://github.com/go-vela/server/pull/722

This change continues the refactor efforts initially introduced in the above PRs.

This adds a new `secret` package to the `github.com/go-vela/server/database` package.

This contains a `SecretService` interface declaring all functions necessary for worker based interactions with the database:

https://github.com/go-vela/server/blob/662917ed982532b8a14f289878a3a694c2d75938/database/secret/service.go#L11-L63

This package also contains the `engine` which implements the above service interface:

https://github.com/go-vela/server/blob/662917ed982532b8a14f289878a3a694c2d75938/database/secret/secret.go#L25-L39

This `engine` contains no raw SQL queries for integrating with the `secrets` table.

Instead, we leverage our DB library's (https://gorm.io/) agnostic abstraction for integrating with that table.

PR #721, opened 2022-10-07
refactor(database): move hook logic into separate package

jbrockopp:
Based off of https://github.com/go-vela/server/pull/574, https://github.com/go-vela/server/pull/663, https://github.com/go-vela/server/pull/687 and https://github.com/go-vela/server/pull/692

This change continues the refactor efforts initially introduced in the above PRs.

This adds a new `hook` package to the `github.com/go-vela/server/database` package.

This contains a `HookService` interface declaring all functions necessary for worker based interactions with the database:

https://github.com/go-vela/server/blob/5bec8fa7f7d7c8dfc9c9e20e3bc63146ff2d0253/database/hook/service.go#L11-L49

This package also contains the `engine` which implements the above service interface:

https://github.com/go-vela/server/blob/5bec8fa7f7d7c8dfc9c9e20e3bc63146ff2d0253/database/hook/hook.go#L23-L37

This `engine` contains no raw SQL queries for integrating with the `hooks` table.

Instead, we leverage our DB library's (https://gorm.io/) agnostic abstraction for integrating with that table.

PR #810, opened 2023-04-08
refactor(database): move step logic into separate package

jbrockopp:
Based off of https://github.com/go-vela/server/pull/574, https://github.com/go-vela/server/pull/663, https://github.com/go-vela/server/pull/687, https://github.com/go-vela/server/pull/692, https://github.com/go-vela/server/pull/721, https://github.com/go-vela/server/pull/722 and https://github.com/go-vela/server/pull/782

This adds a new `step` package to the `github.com/go-vela/server/database` package.

This contains a `StepService` interface declaring all functions necessary for worker based interactions with the database:

https://github.com/go-vela/server/blob/cc2aa4bfce66da935a6964dcb2495f41c4ee3f32/database/step/service.go#L11-L49

https://github.com/go-vela/server/blob/cc2aa4bfce66da935a6964dcb2495f41c4ee3f32/database/step/step.go#L23-L38

This `engine` contains no raw SQL queries for integrating with the `steps` table.


database/step/list_image.go
```diff
@@ -0,0 +1,44 @@
...
+	s := []struct {
+		Image sql.NullString
+		Count sql.NullInt32
+	}{}
+	images := make(map[string]float64)
+
+	// send query to the database and store result in variable
+	err := e.client.
+		Table(constants.TableStep).
+		Select("image", " count(image) as count").
```
cognifloyd, line 27: Why the space?
```suggestion
		Select("image", "count(image) as count").
```

database/step/list_status.go
```diff
@@ -0,0 +1,50 @@
...
+		"failure": 0,
+		"killed":  0,
+		"running": 0,
+		"success": 0,
+	}
+
+	// send query to the database and store result in variable
+	err := e.client.
+		Table(constants.TableStep).
+		Select("status", " count(status) as count").
```
cognifloyd, line 33:
```suggestion
		Select("status", "count(status) as count").
```

cognifloyd (commented):
Just a few nits.

database/step/delete.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/go-vela/types/database"
+	"github.com/go-vela/types/library"
+	"github.com/sirupsen/logrus"
+)
+
+// DeleteStep deletes an existing step from the database.
+func (e *engine) DeleteStep(s *library.Step) error {
+	e.logger.WithFields(logrus.Fields{
+		"step": s.GetNumber(),
+	}).Tracef("deleting step %s from the database", s.GetName())
```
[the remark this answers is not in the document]
jbrockopp: I agree that we should want to introduce a layer of consistency across the packages.

However, the `library.Step` type doesn't have the repo and build immediately accessible (only their `ID`):

https://github.com/go-vela/types/blob/main/library/step.go#L16-L36

This is one (of many) reasons why I want to introduce something like:

https://github.com/go-vela/community/pull/639

In a scenario where the above proposal was implemented, then we could easily access those objects.

database/step/list_image.go
jbrockopp: The space is intentional and my OCD required I add it to make the query look prettier 😅 

Without the space, the query looks like:

```sql
SELECT "image",count(image) as count FROM "steps"
```

With the space, the query looks like:

```sql
SELECT "image", count(image) as count FROM "steps"
```

database/step/delete.go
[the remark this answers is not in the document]
cognifloyd: Oh. Yes of course. That makes sense. Thank you.

database/step/list_image.go
cognifloyd: LOL. It was on purpose, so it works for me.

cognifloyd (approved):
Nits resolved. No changes needed. LGTM!

Thanks again for doing all this refactoring. The database and API layers are feeling much cleaner.

wass3rw3rk (approved):
super minor copy pasta error, but i don't want to hold back merge

database/postgres/postgres.go
```diff
@@ -382,6 +381,18 @@ func createServices(c *client) error {
 		return err
 	}
 
+	// create the database agnostic step service
+	//
+	// https://pkg.go.dev/github.com/go-vela/server/database/repo#New
```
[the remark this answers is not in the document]
jbrockopp: good catch 😄 I've addressed this within https://github.com/go-vela/server/pull/816/commits/19c554eb3827509a038d8d63e3e644b9afa49629 as a part of https://github.com/go-vela/server/pull/816

My line of thinking is we can merge this one as-is and that PR will resolve the typo

PR #811, opened 2023-04-10
enhance(webhook)!: verify repo owner has at least write access to SCM repo

ecrupper:
Closes https://github.com/go-vela/community/issues/742

Also helps in cases where the repo owner has been suspended.

Marked as BREAKING because some users may not care about the build status reporting to the SCM, and thus the repo owner having write access isn't a sticking point for them — with this change, it will have to be.


timhuynh94 (dismissed):

wass3r (commented):
does that mean we don't have to do this later on: https://github.com/go-vela/server/blob/dc5eed10839c1a2de77c3aac3121a8060652be81/api/webhook.go#L351-L374 seems like we're doubling up on db calls for the same thing? also, i see we are checking for valid userid there, do we need to adopt or move that into this new section?

ecrupper:
@wass3r Good catch, totally missed that. I'll just move this access check into that section

wass3rw3rk (approved):

timhuynh94 (approved):

PR #825, opened 2023-04-21
fix(scm): correct init hook status and send accurate events

ecrupper:
Noticed that `init` webhooks were set to a failed status.

Also saw that when we already have a record of a repo and are in the process of re-enabling it, we should make sure the events specified for that record match what we send the SCM, rather than the default events.


KellyMerrick (approved):
lgtm

wass3r (approved):

PR #830, opened 2023-04-24
fix(queue/metrics): include default route in queued build count

ecrupper:
When gathering queued builds, if an admin supplies specific queue routes, they get overwrite the default route from the channel slice.


ecrupper:
nvm

PR #834, opened 2023-04-28
feat(database): add support for schedules

jbrockopp:

Dependent on https://github.com/go-vela/types/pull/289

This adds a new `schedule` package to the `github.com/go-vela/server/database` package.

This contains a `ScheduleInterface` interface declaring all functions necessary for schedule based interactions with the database:

https://github.com/go-vela/server/blob/32d83df0e137917984c42396e35bee6f91e119a2/database/schedule/interface.go#L12-L49

This package also contains the `engine` which implements the above schedule interface:

https://github.com/go-vela/server/blob/32d83df0e137917984c42396e35bee6f91e119a2/database/schedule/schedule.go#L23-L37

This `engine` contains no raw SQL queries for integrating with the `schedules` table.


JordanSussman:
@jbrockopp can we mark this as Ready for review?

database/schedule/get_repo.go
```diff
@@ -0,0 +1,37 @@
...
+package schedule
+
+import (
+	"github.com/go-vela/types/constants"
+	"github.com/go-vela/types/database"
+	"github.com/go-vela/types/library"
+	"github.com/sirupsen/logrus"
+)
+
+// GetScheduleForRepo gets a schedule by repo ID and number from the database.
```
ecrupper, line 14:
```suggestion
// GetScheduleForRepo gets a schedule by repo ID and name from the database.
```

database/schedule/interface.go
```diff
@@ -0,0 +1,49 @@
...
+	CountSchedules() (int64, error)
+	// CountSchedulesForRepo defines a function that gets the count of schedules by repo ID.
+	CountSchedulesForRepo(*library.Repo) (int64, error)
+	// CreateSchedule defines a function that creates a new schedule.
+	CreateSchedule(*library.Schedule) error
+	// DeleteSchedule defines a function that deletes an existing schedule.
+	DeleteSchedule(*library.Schedule) error
+	// GetSchedule defines a function that gets a schedule by ID.
+	GetSchedule(int64) (*library.Schedule, error)
+	// GetScheduleForRepo defines a function that gets a schedule by repo ID and number.
```
ecrupper, line 39:
```suggestion
	// GetScheduleForRepo defines a function that gets a schedule by repo ID and name.
```

database/schedule/list.go
```diff
@@ -0,0 +1,52 @@
...
+	// variables to store query results and return value
+	count := int64(0)
+	s := new([]database.Schedule)
+	schedules := []*library.Schedule{}
+
+	// count the results
+	count, err := e.CountSchedules()
+	if err != nil {
+		return nil, err
+	}
```
ecrupper, line 26: I'm curious if this is necessary. It doesn't look like we use it other than short circuiting an empty table, and the `COUNT` query is fairly taxing (at least in Postgres): https://wiki.postgresql.org/wiki/Introduction_to_VACUUM,_ANALYZE,_EXPLAIN,_and_COUNT#COUNT.28.2A.29

database/schedule/list_active_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/go-vela/types/library"
+)
+
+func TestSchedule_Engine_ListActiveSchedules(t *testing.T) {
```
ecrupper, line 15: Very minor, but I would like to see either the test for `ListActiveSchedules` or `CountActiveSchedules` actually leverage the filter by setting one of the schedules to inactive.

database/schedule/list_repo.go
```diff
@@ -0,0 +1,63 @@
...
+	// variables to store query results and return value
+	count := int64(0)
+	s := new([]database.Schedule)
+	schedules := []*library.Schedule{}
+
+	// count the results
+	count, err := e.CountSchedulesForRepo(r)
+	if err != nil {
+		return nil, 0, err
+	}
```
ecrupper, line 30: Same idea as above with `ListSchedules` and the necessity of this count request

ecrupper (changes requested):
A couple random typos / nits as well as a question related to the count request for those list functions.

database/schedule/list.go
jbrockopp: To provide context, this was adopted from an existing pattern we already use for `builds`:

https://github.com/go-vela/server/blob/3aa414e46ced00ce3019f0d58120801a7c4cd233/database/postgres/build_list.go#L89-L113

 And this code follows the same pattern used in the other database functions to list resources e.g.

* https://github.com/go-vela/server/blob/main/database/hook/list.go#L14
* https://github.com/go-vela/server/blob/main/database/repo/list.go#L14
* https://github.com/go-vela/server/blob/main/database/user/list.go#L14

In general, this need for the count stems from usage for API pagination.

Before I started the DB refactor efforts, we still called the count for API but did so in a separate function i.e.

https://github.com/go-vela/server/blob/8c2fdb27d4a8ab8eb55e272a9135c6939c1ec28b/api/service.go#L241-L259

I decided it was worth adopting the pattern setup for `builds` since it reduces the need to call them separately.

jbrockopp: Also, I thought it was worth highlighting a statement from the document you linked:

> If you are using count(*), the database is free to use any column to count, which means it can pick the smallest covering index to scan (note that this is why count(*) is much better than count(some_field), as long as you don't care if null values of some_field are counted). Since indexes often fit entirely in memory, this means count(*) is often very fast.

In general, calling `count(*)` for a specific table isn't very taxing. Could you elaborate on your concern?

Are you facing issues with this for other tables and that's why you're mentioning it here?

ecrupper: To be clear, I don't really have any _issues_ with the implementation, nor are we experiencing performance problems. Though, `COUNT` queries across a specific table without any `WHERE` clause related to an index are relatively slow (or slower than you'd expect is probably a better descriptor): https://wiki.postgresql.org/wiki/Slow_Counting. 

I'm fine moving this conversation elsewhere as a potential reconsideration of design since it's endemic across the code base, as you pointed out 👍 

Just trying to give all my thoughts so I don't forget 😃

ecrupper (approved):
lgtm

database/schedule/list.go
jbrockopp: sure 👍 perhaps it's worth opening an issue for this so it  doesn't get forgotten?

ecrupper: https://github.com/go-vela/community/issues/810

plyr4 (approved):

PR #833, opened 2023-04-28
feat(types): add support for schedules

jbrockopp:
xref: https://github.com/go-vela/community/issues/538

Part of https://github.com/go-vela/community/pull/772

This change adds the `structs` (a.k.a. types) for `schedules` to Vela.

This adds a new `types` package to the following packages:

* `github.com/go-vela/server/api` - code that would normally be in `github.com/go-vela/types/library`
* `github.com/go-vela/server/database` - code that would normally be in `github.com/go-vela/types/database`

I'm currently piggy backing off of a [previous repo restructure proposal](https://github.com/go-vela/community/blob/main/proposals/2021/08-25_repo-structure.md) for this.

The end goal is to remove https://github.com/go-vela/types in favor of code within this repo.

> NOTE: If folks would like me to open up a new proposal, please let me know and I will do so.

Also, you'll find that this implementation of `types` for schedules also incorporates this proposal:

https://github.com/go-vela/community/pull/639

PR #836, opened 2023-05-03
feat(api): add support for schedules

JordanSussman:

Dependent on https://github.com/go-vela/server/pull/833 and https://github.com/go-vela/server/pull/834

The code that manages the execution of schedules will added in a separate code change.


jbrockopp (changes requested):

api/schedule/create.go
```diff
@@ -0,0 +1,181 @@
...
+
+	// send API call to capture the repo from the database
+	dbRepo, err := database.FromContext(c).GetRepoForOrg(r.GetOrg(), r.GetName())
+	if err == nil && !dbRepo.GetActive() {
+		retErr := fmt.Errorf("unable to create schedule: %s repo %s is disabled", input.GetName(), dbRepo.GetFullName())
+
+		util.HandleError(c, http.StatusConflict, retErr)
+
+		return
+	}
```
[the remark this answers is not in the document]
JordanSussman: done

jbrockopp:
@JordanSussman can we mark this as Ready for review?

api/schedule/create.go
```diff
@@ -0,0 +1,230 @@
...
+	// update engine logger with API metadata
+	//
+	// https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Entry.WithFields
+	logrus.WithFields(logrus.Fields{
+		"org":  r.GetOrg(),
+		"repo": r.GetName(),
+		"user": u.GetName(),
+	}).Infof("creating new schedule %s", input.GetName())
+
+	// ensure repo is allowed to be activated
```
jbrockopp, line 119: Thoughts on updating the wording here slightly?

```suggestion
	// ensure repo is allowed to create new schedules
```

api/schedule/create.go
```diff
@@ -0,0 +1,230 @@
...
+		retErr := fmt.Errorf("unable to create schedule: %s repo %s is disabled", input.GetName(), r.GetFullName())
+
+		util.HandleError(c, http.StatusConflict, retErr)
+
+		return
+	}
+
+	// if the repo exists but is inactive
+	if len(r.GetOrg()) > 0 && !dbSchedule.GetActive() && input.GetActive() {
+		// update the repo owner
```
jbrockopp, line 167: Thoughts on updating the wording here slightly?

```suggestion
		// update the user who created the schedule
```

api/schedule/create.go
```diff
@@ -0,0 +1,230 @@
...
+		dbSchedule.SetCreatedBy(u.GetName())
```
jbrockopp, line 168: Do you think this should set `UpdatedBy` instead of `CreatedBy`?

```suggestion
		dbSchedule.SetUpdatedBy(u.GetName())
```

api/schedule/create.go
```diff
@@ -0,0 +1,230 @@
...
+		// activate the schedule
+		dbSchedule.SetActive(true)
+
+		// send API call to update the repo
```
jbrockopp, line 172: Thoughts on updating the wording here slightly?

```suggestion
		// send API call to update the schedule
```

api/schedule/create.go
```diff
@@ -0,0 +1,230 @@
...
+		err = database.FromContext(c).UpdateSchedule(dbSchedule)
+		if err != nil {
+			retErr := fmt.Errorf("unable to set schedule %s to active: %w", dbSchedule.GetName(), err)
+
+			util.HandleError(c, http.StatusInternalServerError, retErr)
+
+			return
+		}
+
+		// send API call to capture the updated repo
```
jbrockopp, line 182: Thoughts on updating the wording here slightly?

```suggestion
		// send API call to capture the updated schedule
```

api/schedule/create.go
```diff
@@ -0,0 +1,230 @@
...
+		err = database.FromContext(c).CreateSchedule(s)
+		if err != nil {
+			retErr := fmt.Errorf("unable to create new schedule %s: %w", r.GetName(), err)
+
+			util.HandleError(c, http.StatusInternalServerError, retErr)
+
+			return
+		}
+
+		// send API call to capture the created repo
```
jbrockopp, line 195: Thoughts on updating the wording here slightly?

```suggestion
		// send API call to capture the created schedule
```

api/schedule/update.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// send API call to capture the schedule
+	s, err := database.FromContext(c).GetScheduleForRepo(r, scheduleName)
+	if err != nil {
+		retErr := fmt.Errorf("unable to get schedule %s: %w", scheduleName, err)
+
+		util.HandleError(c, http.StatusNotFound, retErr)
+
+		return
+	}
```
jbrockopp, line 106: We should replace this with `schedule.Retrieve()` middleware like we do in other `Update*` functions:

https://github.com/go-vela/server/blob/main/api/pipeline/update.go#L70-L74

This will help reduce an extra DB call since we already fetch the schedule and set it in the context.

jbrockopp (changes requested):

api/schedule/create.go
```diff
@@ -0,0 +1,230 @@
...
+	if !r.GetActive() {
+		retErr := fmt.Errorf("unable to create schedule: %s repo %s is disabled", input.GetName(), r.GetFullName())
+
+		util.HandleError(c, http.StatusConflict, retErr)
+
+		return
+	}
+
+	// if the repo exists but is inactive
+	if len(r.GetOrg()) > 0 && !dbSchedule.GetActive() && input.GetActive() {
```
ecrupper, line 166: I'm curious why the `len(r.GetOrg())` check here? All repos are validated before DB creation to have an org: https://github.com/go-vela/types/blob/f538de06bbf67a5c9ab5a48eeca0d459f05e3bd8/database/repo.go#L252-L255.

ecrupper (changes requested):

api/schedule/create.go
JordanSussman: This was unintentionally left when I copy/pasted code from the api/repo/create.go code.

JordanSussman:
@jbrockopp @ecrupper feedback has been addressed.

jbrockopp (approved):
LGTM

ecrupper (changes requested):

ecrupper:
I also want to point out that I was kind of disappointed with github.com/adhocore/gronx error handling with invalid entries. For example, I provided an entry of `0 0 0 * *`, which is invalid due to `day_of_month` being 0, but the error output was

```json
{
	"error": "schedule of  is invalid: tried so hard"
}
```

I expect users to be able to diagnose an invalid cron expression, but "tried so hard" isn't exactly a great push in the right direction.

api/schedule/update.go
```diff
@@ -0,0 +1,140 @@
...
+
+	if input.GetName() != "" {
+		// update name if defined
+		s.SetName(input.GetName())
+	}
+
+	if input.GetEntry() != "" {
+		err = validateEntry(minimumFrequency, input.GetEntry())
+		if err != nil {
+			retErr := fmt.Errorf("schedule of %s is invalid: %w", input.GetName(), err)
```
ecrupper, line 115:
```suggestion
			retErr := fmt.Errorf("schedule of %s is invalid: %w", input.GetEntry(), err)
```

api/schedule/create.go
```diff
@@ -0,0 +1,230 @@
...
+
+		util.HandleError(c, http.StatusBadRequest, retErr)
+
+		return
+	}
+
+	// ensure the entry is valid
+	err = validateEntry(minimumFrequency, input.GetEntry())
+	if err != nil {
+		retErr := fmt.Errorf("schedule of %s is invalid: %w", input.GetName(), err)
```
ecrupper, line 98:
```suggestion
		retErr := fmt.Errorf("schedule of %s is invalid: %w", input.GetEntry(), err)
```
I think this will pinpoint the issue better than the schedule name

ecrupper (changes requested):

api/schedule/update.go
```diff
@@ -0,0 +1,140 @@
...
+		util.HandleError(c, http.StatusBadRequest, retErr)
+
+		return
+	}
+
+	// update schedule fields if provided
+	if input.GetActive() {
+		// update active if set to true
+		s.SetActive(input.GetActive())
+	}
```
plyr4, line 105: should we allow deactivation?

```suggestion
	if input.Active != nil {
		// update active if provided
		s.SetActive(input.GetActive())
	}
```

api/schedule/create.go
```diff
@@ -0,0 +1,230 @@
...
+		if err != nil {
+			retErr := fmt.Errorf("unable to set schedule %s to active: %w", dbSchedule.GetName(), err)
+
+			util.HandleError(c, http.StatusInternalServerError, retErr)
+
+			return
+		}
+
+		// send API call to capture the updated schedule
+		s, _ = database.FromContext(c).GetScheduleForRepo(r, dbSchedule.GetName())
```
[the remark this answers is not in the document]
JordanSussman: I'll be honest that I didn't think about handling the error since so much of the existing code base does not... How about we open up a separate GitHub issue to track updating all of these database calls?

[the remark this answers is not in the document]
jbrockopp: FWIW -

We should plan to have the `Create*` and `Update*` funcs return the type so we don't need these calls

i.e. https://github.com/go-vela/server/pull/701

ecrupper (approved):
Nice

plyr4 (approved):

PR #840, opened 2023-05-08
enhance(build): add ability to cancel pending builds

ecrupper:
Part of the effort for https://github.com/go-vela/community/issues/485

This change will give users the ability to cancel builds that are in a pending state. The server will perform the necessary updates to the database to ensure the build (and its steps/services) has the correct status. The worker will pop that item off the queue and determine whether or not it has been cancelled. In other words, rather than scan the queue searching for the item to remove, it seems like a better strategy to let the workers do the cleaning.

### DIFF
--- a/api/build.go
+++ b/api/build.go
@@ -1728,10 +1728,92 @@ func CancelBuild(c *gin.Context) {
 		"user":  u.GetName(),
 	}).Infof("canceling build %s", entry)
 
-	// TODO: add support for removing builds from the queue
-	//
-	// check to see if build is not running
-	if !strings.EqualFold(b.GetStatus(), constants.StatusRunning) {
+	switch b.GetStatus() {
+	case constants.StatusRunning:
+		// retrieve the worker info
+		w, err := database.FromContext(c).GetWorkerForHostname(b.GetHost())
+		if err != nil {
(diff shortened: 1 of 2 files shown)


cognifloyd (dismissed):

api/build.go
```diff
@@ -1979,10 +1980,10 @@ func GetBuildToken(c *gin.Context) {
 		"user":  cl.Subject,
 	}).Infof("generating build token for build %s/%d", r.GetFullName(), b.GetNumber())
 
-	// if build is not in a pending state, then a build token should not be needed - bad request
+	// if build is not in a pending state, then a build token should not be needed - conflict
 	if !strings.EqualFold(b.GetStatus(), constants.StatusPending) {
 		retErr := fmt.Errorf("unable to mint build token: build is not in pending state")
-		util.HandleError(c, http.StatusBadRequest, retErr)
+		util.HandleError(c, http.StatusConflict, retErr)
```
cognifloyd, line 1986: Could you add this to the open API spec in the function doc string?

cognifloyd (changes requested):

api/build.go
ecrupper: Good catch — done

plyr4 (approved):

PR #846, opened 2023-05-17
feat: add support for processing schedules

jbrockopp:

Dependent on https://github.com/go-vela/server/pull/836

This adds support for Vela to interact with `schedules` created in the database and spawn builds accordingly.

> NOTE: In order to run a pipeline with support for `schedules`, you'll need https://github.com/go-vela/worker/pull/479.


cmd/vela-server/server.go
```diff
@@ -102,54 +103,90 @@ func server(c *cli.Context) error {
...
-			err := srv.ListenAndServe()
-			if err != nil {
-				tomb.Kill(err)
-			}
-		}()
+		return err
+	})
 
-		//nolint:gosimple // ignore this for now
+	// spawn goroutine for starting the server
```
plyr4, line 172:
```suggestion
	// spawn goroutine for starting the scheduler
```

cmd/vela-server/schedule.go
```diff
@@ -0,0 +1,357 @@
...
+
+			return err
+		}
+
+		// update repo fields with any changes from SCM process
+		r.SetTopics(r.GetTopics())
+		r.SetBranch(r.GetBranch())
+
+		// set the build numbers based off repo counter
+		r.SetCounter(r.GetCounter() + 1)
```
plyr4, line 213: what are your thoughts on updating the repo counter in the database immediately after incrementing it, to reduce the risk of duplicate constrain violations? 

im just worried about compile + plan time for template-heavy pipelines and not making it to [this line](https://github.com/go-vela/server/pull/846/files#diff-6899fce6b909a4c271e1eabdda2f7240ad4ba8776c9a943e7b06b25d132e01f1R329) before an event-driven webhook is processed

jbrockopp: I'm uncertain we can do that since we'd be prematurely updating the `counter` for a repo before a build is triggered.

For the sake of the discussion, let's assume we do adopt this approach.

Let's say we're processing a schedule with the name of `X` for repo `Y` which has `10` builds.

As we're processing `X`, we'd increment the `counter` for `Y` from `10` -> `11` and update the DB.

Then, we attempt to compile the pipeline (let's pretend this works) for build number `11`.

Then, we attempt to add it to the DB (let's pretend this call fails):

https://github.com/go-vela/server/blob/6f740fa783dc2997ffe3617b553272007a3c951f/cmd/vela-server/schedule.go#L296-L307

From the above code, which was adopted from `webhook.go`, we'd retry which would go from `11` -> `12`.

And let's assume the second time we try all of this, everything works i.e. temp network failure

In this scenario, we'd never see a build for `11` since the second time we'd update the `counter` for `Y` to `12`.

plyr4: sure, then ignoring my incomplete solution, do you agree that the risk between this code and webhook events processing the same build number still exists? 

or am i thinking of it incorrect

plyr4: i suppose its in the same vein as the commented out TODO
```
		// TODO:
		// - if a build gets created and something else fails midway,
		//   the next loop will attempt to create the same build,
		//   using the same Number and thus create a constraint
		//   conflict; consider deleting the partially created
		//   build object in the database
```

the risk is small, and hopefully the retries would help.

cmd/vela-server/server.go
```diff
@@ -111,47 +114,92 @@ func server(c *cli.Context) error {
...
-				tomb.Kill(err)
-			}
-		}()
+		return err
+	})
 
-		//nolint:gosimple // ignore this for now
+	// spawn goroutine for starting the scheduler
+	g.Go(func() error {
+		logrus.Info("starting scheduler")
```
wass3rw3rk, line 176: this would kill the whole server (everything in the errgroup) if there was an error during schedule process, correct? is that a bit too extreme?

wass3rw3rk (commented):
couple of comments for now, but looks good overall - thanks for the work!

and a note for @ecrupper to make sure it plays nicely with https://github.com/go-vela/server/pull/859

cmd/vela-server/schedule.go
jbrockopp: @plyr4 yeah I agree with you that the risk is still there 😞 

However, as you pointed out, the issue itself should hopefully be mitigated by retry logic.

FWIW, I did take the liberty of updating the number of retries from `3` -> `5` to help with this:

https://github.com/go-vela/server/blob/46fcc273b3aaea19f0e9fbc5a0db33212481c8f7/api/webhook.go#L414-L415

https://github.com/go-vela/server/blob/46fcc273b3aaea19f0e9fbc5a0db33212481c8f7/cmd/vela-server/schedule.go#L187-L188

To refresh folks memory on why we have it set to `3` in the webhook workflow:

https://github.com/go-vela/server/pull/191

cmd/vela-server/schedule.go
```diff
@@ -0,0 +1,382 @@
...
+				// continue to the next iteration of the loop
+				continue
+			}
+
+			return err
+		}
+
+		// update repo fields with any changes from SCM process
+		r.SetTopics(r.GetTopics())
+		r.SetBranch(r.GetBranch())
```
ecrupper, line 237: These lines are no longer necessary since we aren't getting any extra info from the SCM for this

cmd/vela-server/schedule.go
```diff
@@ -0,0 +1,384 @@
...
+
+		var compiled *library.Pipeline
+		// parse and compile the pipeline configuration file
+		p, compiled, err = compiler.
+			Duplicate().
+			WithBuild(b).
+			WithMetadata(metadata).
+			WithRepo(r).
+			WithUser(u).
+			Compile(config)
```
ecrupper:
```suggestion
		p, compiled, err = compiler.
			Duplicate().
			WithBuild(b).
			WithCommit(commit).
			WithMetadata(metadata).
			WithRepo(r).
			WithUser(u).
			Compile(config)
```
This is new 😅 ([ref](https://github.com/go-vela/server/pull/859)). FWIW, adding this line doesn't actually change anything because an empty commit [here](https://github.com/go-vela/server/blob/ffac676b781bab77d57caccbc12edd33d26e66b4/compiler/native/expand.go#L238) just uses the default branch. But I suppose it's nice to not rely on that being true and add the commit ourselves 🤷‍♂️ .

jbrockopp:
> couple of comments for now, but looks good overall - thanks for the work!
> 
> and a note for @ecrupper to make sure it plays nicely with #859

Apologies as I missed that PR 😅 I've updated this PR to also use `compiler.WithCommit()`.

Please let me know if you have any additional feedback or there is anything else I'm missing 👍

ecrupper (changes requested):
Nice, I like it a lot. Small thing I noticed but that's it 👍

cmd/vela-server/server.go
jbrockopp: fixed 😃

cmd/vela-server/schedule.go
jbrockopp: fixed 😃

ecrupper (approved):

cmd/vela-server/server.go
wass3rw3rk: 👍🏼 thanks

wass3rw3rk (approved):

plyr4 (approved):

PR #851, opened 2023-05-19
refactor(api): move step logic to separate package

ecrupper:
Based off of https://github.com/go-vela/server/pull/615 https://github.com/go-vela/server/pull/754 https://github.com/go-vela/server/pull/809 https://github.com/go-vela/server/pull/828 https://github.com/go-vela/server/pull/829 https://github.com/go-vela/server/pull/847 https://github.com/go-vela/server/pull/848 and https://github.com/go-vela/server/pull/849

This change continues the refactor efforts initially introduced in the above PRs.

This adds a new `step` package to the `github.com/go-vela/server/api` package.

This contains all the same handlers that existed previously but with each of them within their own file.

Moved `planSteps` to the `api/build.go` file with the intent of creating a `api/step/plan.go` file later in this refactor, which will include an exportable `PlanSteps` for use in `RestartBuild` and scheduled builds. Open to other ideas with that too. 

Also changed `GetSteps` to `ListSteps` to stay consistent.

PR #849, opened 2023-05-18
refactor(api): move service logic to separate package

ecrupper:
Based off of https://github.com/go-vela/server/pull/615 https://github.com/go-vela/server/pull/754 https://github.com/go-vela/server/pull/809 https://github.com/go-vela/server/pull/828 https://github.com/go-vela/server/pull/829 https://github.com/go-vela/server/pull/847 and https://github.com/go-vela/server/pull/848.

This change continues the refactor efforts initially introduced in the above PRs.

This adds a new `service` package to the `github.com/go-vela/server/api` package.

This contains all the same handlers that existed previously but with each of them within their own file.

Moved `planServices` to the `api/build.go` file with the intent of creating a `api/service/plan.go` file later in this refactor, which will include an exportable `PlanServices` for use in `RestartBuild` and scheduled builds. Open to other ideas with that too. 

Also changed `GetServices` to `ListServices` to stay consistent.

PR #847, opened 2023-05-17
refactor(api): move deployment logic to separate package

ecrupper:
Based off of https://github.com/go-vela/server/pull/615 https://github.com/go-vela/server/pull/754 https://github.com/go-vela/server/pull/809 https://github.com/go-vela/server/pull/828 and https://github.com/go-vela/server/pull/829

This change continues the refactor efforts initially introduced in the above PR.

This adds a new `deployment` package to the `github.com/go-vela/server/api` package.

This contains all the same handlers that existed previously but with each of them within their own file.

I also updated a few of the previous packages in this refactor effort to have a `doc.go` file.

PR #828, opened 2023-04-22
refactor(api): move worker logic to separate package

jbrockopp:
Based off of https://github.com/go-vela/server/pull/615 https://github.com/go-vela/server/pull/754 and https://github.com/go-vela/server/pull/809

This change continues the refactor efforts initially introduced in the above PR.

This adds a new `worker` package to the `github.com/go-vela/server/api` package.

This contains all the same handlers that existed previously but with each of them within their own file.

PR #853, opened 2023-05-19
refactor(api): move auth logic to separate package

cognifloyd:
Based off of #615 #754 #809 #828 #829 #847 #848 #849 and #851

This adds a new `auth` package to the `github.com/go-vela/server/api` package.

There are several top-level auth-related handlers, so I combined them into the same package, and I renamed some of the handlers.

Here is a summary of handler moves and renames:

| Endpoint | Old File | Old Handler Name | New File | New Handler Name |
|---|---|---|---|---|
| `GET /authenticate` | `api/authenticate.go` | `Authenticate` | `api/auth/get_token.go` | `GetAuthToken` |
| `GET /authenticate/:type[/:port]` | `api/authenticate.go` | `AuthenticateType` | `api/auth/redirect.go` | `GetAuthRedirect` |
| `POST /authenticate` | `api/authenticate.go` | `AuthenticateToken` | `api/auth/post_token.go` | `PostAuthToken` |
| `GET /login` | `api/login.go` | `Login` | `api/auth/login.go` | `Login` |
| `GET /logout` | `api/logout.go` | `Logout` | `api/auth/logout.go` | `Logout` |
| `GET /token-refresh` | `api/token.go` | `RefreshAccessToken` | `api/auth/refresh.go` | `RefreshAccessToken` |
| `GET /validate-token` | `api/token.go` | `ValidateServerToken` | `api/auth/validate.go` | `ValidateServerToken` |

NB: The commits are arranged to preserve git history (git blame) for all files. I suspect that the squash commit will erase that, but oh well - the history is at least preserved in this branch.


api/auth/validate.go
```diff
@@ -0,0 +1,50 @@
...
+//     description: Successfully validated a token
+//     schema:
+//       type: string
+//   '401':
+//     description: Unauthorized
+//     schema:
+//       "$ref": "#/definitions/Error"
+
+// ValidateServerToken will return the claims of a valid server token
+// if it is provided in the auth header.
```
ecrupper, line 37:
```suggestion
// ValidateServerToken will validate if a token was issued by the server
// if it is provided in the auth header.
```

This was my mistake when writing this endpoint, and I'm just now seeing it... 😅

ecrupper (changes requested):
Awesome — I like it. One thing I did notice that was from an old commit that would be nice to change with this refactor

ecrupper (approved):

PR #861, opened 2023-05-25
refactor(db): return hook created or updated

ecrupper:
This refactor is an effort to eliminate several `Get` requests to the DB immediately after creating or updating a resource. Fortunately, with Gorm's [create](https://gorm.io/docs/create.html#Create-Record) and [save](https://gorm.io/docs/update.html#Save-All-Fields) methods, the object is populated with any fields that are generated upon DB insertion (ID namely). In other words, we can safely return the object we use for these operations.


cognifloyd (approved):
Straight forward improvement

jbrockopp (approved):
LGTM

plyr4 (approved):
nice

PR #863, opened 2023-05-26
feat(api/admin): add endpoint to clean pending / running builds

ecrupper:
Oftentimes during upgrades, network blips, or failures in worker-server communication, builds, steps, and or services, never make it out of `pending` or `running` states. This endpoint gives platform admins the power to clean build resources that have been pending or running for longer than a specified time as well as give the builds a specified error message.

### DIFF
--- a/database/build/clean.go
+++ b/database/build/clean.go
@@ -1,0 +1,32 @@
+// Copyright (c) 2023 Target Brands, Inc. All rights reserved.
+//
+// Use of this source code is governed by the LICENSE file in this repository.
+
+package build
+
+import (
+	"github.com/go-vela/types/constants"
+	"github.com/go-vela/types/database"
+	"github.com/go-vela/types/library"
+	"github.com/sirupsen/logrus"
+)
+
+// CleanBuilds updates builds to an error with a provided message with a created timestamp prior to a defined moment.
+func (e *engine) CleanBuilds(msg string, before int64) (int64, error) {
+	logrus.Tracef("cleaning pending or running builds in the database created prior to %d", before)
+
+	b := new(library.Build)
+	b.SetStatus(constants.StatusError)
+	b.SetError(msg)
+
+	build := database.BuildFromLibrary(b)
+
+	// send query to the database
+	result := e.client.
+		Table(constants.TableBuild).
+		Where("created < ?", before).
+		Where("status = 'running' OR status = 'pending'").
+		Updates(build)
+
+	return result.RowsAffected, result.Error
+}
--- a/database/build/clean.go
+++ b/database/build/clean.go
@@ -5,6 +5,8 @@
 package build
 
 import (
+	"time"
+
 	"github.com/go-vela/types/constants"
 	"github.com/go-vela/types/database"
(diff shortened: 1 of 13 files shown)


wass3rw3rk (commented):
i wouldn't mind seeing some INFO-level logging between each table cleanup.

database/build/clean.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// CleanBuilds updates builds to an error with a provided message with a created timestamp prior to a defined moment.
+func (e *engine) CleanBuilds(msg string, before int64) (int64, error) {
+	logrus.Tracef("cleaning pending or running builds in the database created prior to %d", before)
+
+	b := new(library.Build)
+	b.SetStatus(constants.StatusError)
+	b.SetError(msg)
```
cognifloyd, line 20: I think we should set the finished time too.

```suggestion
	b := new(library.Build)
	b.SetStatus(constants.StatusError)
	b.SetError(msg)
	b.SetFinished(time.Now().UTC().Unix())
```

database/service/clean.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/go-vela/types/library"
+	"github.com/sirupsen/logrus"
+)
+
+// CleanServices updates services to an error with a created timestamp prior to a defined moment.
+func (e *engine) CleanServices(before int64) (int64, error) {
+	logrus.Tracef("cleaning pending or running steps in the database created prior to %d", before)
+
+	s := new(library.Service)
+	s.SetStatus(constants.StatusError)
```
cognifloyd, line 19: For service and step do we want to also add the admin's error message?

And should we set the finished time?

```suggestion
	s := new(library.Service)
	s.SetStatus(constants.StatusError)
	s.SetError(msg)
	s.SetFinished(time.Now().UTC().Unix())
```

database/build/clean_test.go
```diff
@@ -0,0 +1,120 @@
...
+	_buildFour.SetNumber(4)
+	_buildFour.SetCreated(5)
+	_buildFour.SetStatus("running")
+
+	_postgres, _mock := testPostgres(t)
+	defer func() { _sql, _ := _postgres.client.DB(); _sql.Close() }()
+
+	// ensure the mock expects the name query
+	_mock.ExpectExec(`UPDATE "builds" SET "status"=$1,"error"=$2,"finished"=$3,"deploy_payload"=$4 WHERE created < $5 AND (status = 'running' OR status = 'pending')`).
+		WithArgs("error", "msg", time.Now().UTC().Unix(), AnyArgument{}, 3).
```
cognifloyd, line 51: Is this going to create a flaky test? Because timestamp is calculated both in the test and in the code under test.

Ditto for the step and service clean tests.

ecrupper: It would if we were comparing the objects, but all we do is check the count in the test

wass3rw3rk (approved):

cognifloyd (approved):

plyr4 (approved):
nice

PR #856, opened 2023-05-22
refactor(api): move secret logic to separate package

ecrupper:
Based off of https://github.com/go-vela/server/pull/615 https://github.com/go-vela/server/pull/754 https://github.com/go-vela/server/pull/809 https://github.com/go-vela/server/pull/828 https://github.com/go-vela/server/pull/829 https://github.com/go-vela/server/pull/847 https://github.com/go-vela/server/pull/848 https://github.com/go-vela/server/pull/849 https://github.com/go-vela/server/pull/851 https://github.com/go-vela/server/pull/853 and https://github.com/go-vela/server/pull/855

This change continues the refactor efforts initially introduced in the above PRs.

This adds a new `secret` package to the `github.com/go-vela/server/api` package.

This contains all the same handlers that existed previously but with each of them within their own file.

Moved the `unique` helper function to the `github.com/go-vela/server/util` package, as it seems well-placed there, but I'm willing to move it back into the secret package if requested.

Also changed `GetSecret` to `ListSecrets` to stay consistent.

PR #855, opened 2023-05-19
refactor(api): move log logic into separate package

ecrupper:
Based off of https://github.com/go-vela/server/pull/615 https://github.com/go-vela/server/pull/754 https://github.com/go-vela/server/pull/809 https://github.com/go-vela/server/pull/828 https://github.com/go-vela/server/pull/829 https://github.com/go-vela/server/pull/847 https://github.com/go-vela/server/pull/848 https://github.com/go-vela/server/pull/849 https://github.com/go-vela/server/pull/851 and https://github.com/go-vela/server/pull/853

This change continues the refactor efforts initially introduced in the above PRs.

This adds a new `log` package to the `github.com/go-vela/server/api` package.

This contains all the same handlers that existed previously but with each of them within their own file.

Also changed `GetBuildLogs` to `ListLogsForBuild` to stay consistent.

PR #809, opened 2023-04-08
refactor(api): move user logic to separate package

jbrockopp:
Based off of https://github.com/go-vela/server/pull/615 and https://github.com/go-vela/server/pull/754

This change continues the refactor efforts initially introduced in the above PR.

This adds a new `user` package to the `github.com/go-vela/server/api` package.

This contains all the same handlers that existed previously but with each of them within their own file.

PR #864, opened 2023-05-26
refactor(api): move build logic to separate package

ecrupper:
Based off of https://github.com/go-vela/server/pull/615 https://github.com/go-vela/server/pull/754 https://github.com/go-vela/server/pull/809 https://github.com/go-vela/server/pull/828 https://github.com/go-vela/server/pull/829 https://github.com/go-vela/server/pull/847 https://github.com/go-vela/server/pull/848 https://github.com/go-vela/server/pull/849 https://github.com/go-vela/server/pull/851 https://github.com/go-vela/server/pull/853 https://github.com/go-vela/server/pull/855 and https://github.com/go-vela/server/pull/856

This adds a new `build` package to the `github.com/go-vela/server/api` package.

It also adds a new `webhook` package in order to export `PublishToQueue` as a function within `build` package. 

Now that build has been moved to its own package, I was also able to put `PlanSteps` and `PlanServices` in their respective packages.


api/step/plan.go
```diff
@@ -0,0 +1,125 @@
...
+func PlanSteps(database database.Interface, p *pipeline.Build, b *library.Build) ([]*library.Step, error) {
+	// variable to store planned steps
+	steps := []*library.Step{}
+
+	// iterate through all pipeline stages
+	for _, stage := range p.Stages {
+		// iterate through all steps for each pipeline stage
+		for _, step := range stage.Steps {
+			// create the step object
+			s := new(library.Step)
```
cognifloyd, line 29: What would you think of refactoring this file just a bit farther to reduce duplication a bit?

I think it would be nice to add a `planStep` function, and then update the `PlanSteps` function so it looks like this:

```go
steps := []*library.Step{}

for _, stage := range p.Stages {
	for _, step := range stage.Steps {
		steps = append(steps, planStep(step))
	}
}

for _, step := range p.Steps {
	steps = append(steps, planStep(step))
}
```

api/build/token.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/go-vela/server/router/middleware/claims"
+	"github.com/go-vela/server/router/middleware/org"
+	"github.com/go-vela/server/router/middleware/repo"
+	"github.com/go-vela/server/util"
+	"github.com/go-vela/types/constants"
+	"github.com/go-vela/types/library"
+	"github.com/sirupsen/logrus"
+)
+
+// swagger:operation GET /api/v1/repos/{org}/{repo}/builds/{build}/token builds GetBuildToken
```
cognifloyd, line 25: I wonder if this endpoint should be in the `api/auth` package instead of `api/build` since it is auth-specific. Then all of the auth stuff is in the same package even if the API spreads it across a variety of endpoints. What do you think?

api/build/create.go
```diff
@@ -0,0 +1,392 @@
...
+//   required: true
+//   type: string
+// - in: path
+//   name: repo
+//   description: Name of the repo
+//   required: true
+//   type: string
+// - in: body
+//   name: body
+//   description: Payload containing the build to update
```
cognifloyd, line 50:
```suggestion
//   description: Payload containing the build to create
```

api/build/get_id.go
```diff
@@ -0,0 +1,118 @@
...
+	// Capture user access from SCM. We do this in order to ensure user has access and is not
+	// just retrieving any build using a random id number.
+	perm, err := scm.FromContext(c).RepoAccess(u, u.GetToken(), r.GetOrg(), r.GetName())
+	if err != nil {
+		logrus.Errorf("unable to get user %s access level for repo %s", u.GetName(), r.GetFullName())
+	}
+
+	// Ensure that user has at least read access to repo to return the build
+	if perm == "none" && !u.GetAdmin() {
+		retErr := fmt.Errorf("unable to retrieve build %d: user does not have read access to repo %s", id, r.GetFullName())
```
cognifloyd, line 110: Hmm. Probably not for this PR - but if the repo is private, the error message probably shouldn't say the name of the repo. 🤔

api/build/list_org.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/go-vela/server/router/middleware/org"
+	"github.com/go-vela/server/router/middleware/user"
+	"github.com/go-vela/server/scm"
+	"github.com/go-vela/server/util"
+	"github.com/go-vela/types/constants"
+	"github.com/go-vela/types/library"
+	"github.com/sirupsen/logrus"
+)
+
+// swagger:operation GET /api/v1/repos/{org} builds ListBuildsForOrg
```
cognifloyd, line 24:
```suggestion
// swagger:operation GET /api/v1/repos/{org}/builds builds ListBuildsForOrg
```

compare with:
https://github.com/go-vela/server/blob/f7c3c6b593305b1f0b1023c55a1519cd136e6d3d/api/repo/list_org.go#L23

Interestingly, this means that a repo named `builds` is not supported by Vela.
https://github.com/go-vela/server/blob/f7c3c6b593305b1f0b1023c55a1519cd136e6d3d/router/repo.go#L23-L24
https://github.com/go-vela/server/blob/f7c3c6b593305b1f0b1023c55a1519cd136e6d3d/router/repo.go#L63-L72

cognifloyd: Though it would not be part of this refactor PR, it might be nice to change the API path for this function to:
```suggestion
// swagger:operation GET /api/v1/repos/*/{org} builds ListBuildsForOrg
```

cognifloyd: Though it would not be part of this refactor PR, it might be nice to change the API path for this function to: 
```suggestion
// swagger:operation GET /api/v1/repos/*/{org}/builds builds ListBuildsForOrg
```

cognifloyd: Though it would not be part of this refactor PR, it might be nice to change the API path for this function to: 
```suggestion
// swagger:operation GET /api/v1/repos/{org}/*/builds builds ListBuildsForOrg
```

api/build/create.go
```diff
@@ -0,0 +1,392 @@
...
+		p,
+		input,
+		r,
+		u,
+	)
+}
+
+// getPRNumberFromBuild is a helper function to
+// extract the pull request number from a Build.
+func getPRNumberFromBuild(b *library.Build) (int, error) {
```
cognifloyd, line 376: this is used for both `CreateBuild` and `UpdateBuild` but is in `create.go` - should it go in a file on its own? 🤔 

It's probably fine here, but I'm asking anyway. 🤷

cognifloyd (changes requested):
Some misc questions and a couple typos.

api/build/token.go
ecrupper: Yeah I was thinking similarly, but landed on sticking with the `build` package since the URL path is very much of the build flavor.

api/build/get_id.go
ecrupper: Hm, yeah I can change that — good call

api/build/create.go
ecrupper: Going to leave it there for now, but I almost wonder if this belongs in types rather than as a helper function

api/build/list_org.go
cognifloyd: At a minimum we need to fix this swagger doc string so that it lists the correct endpoint.
```suggestion
// swagger:operation GET /api/v1/repos/{org}/builds builds ListBuildsForOrg
```

My other suggestion can wait for a follow up PR.

cognifloyd (approved):

plyr4 (approved):
nice

PR #873, opened 2023-06-06
refactor(database): use agnostic engine

jbrockopp:
Dependent on https://github.com/go-vela/server/pull/870

This change updates the `database.New()` function to return the database agnostic `engine{}` type introduced in https://github.com/go-vela/server/pull/868.

The logic for this function was copied from the duplicate code we already have in both supported drivers:

* [postgres](https://github.com/go-vela/server/blob/main/database/postgres/postgres.go#L80)
* [sqlite](https://github.com/go-vela/server/blob/main/database/sqlite/sqlite.go#L79)

This also adds a new `FromCLIContext()` function to the `database` package to create the agnostic `engine{}`.


database/context.go
```diff
@@ -35,3 +38,19 @@ func FromContext(c context.Context) Interface {
 func ToContext(c Setter, d Interface) {
 	c.Set(key, d)
 }
+
+// FromCLIContext creates and returns a database engine from the urfave/cli context.
+func FromCLIContext(c *cli.Context) (Interface, error) {
```
jbrockopp, line 43: This logic was copied from the `setupDatabase()` function in the `cmd/vela-server/database.go` file.

database/context_test.go
```diff
@@ -1,91 +1,152 @@
...
+	"reflect"
 	"testing"
+	"time"
 
 	"github.com/gin-gonic/gin"
-	"github.com/go-vela/server/database/sqlite"
+	"github.com/urfave/cli/v2"
 )
 
 func TestDatabase_FromContext(t *testing.T) {
```
jbrockopp, line 17: This function was refactored to use table driven tests.

database/context_test.go
```diff
@@ -1,91 +1,152 @@
...
+				t.Errorf("FromContext for %s is %v, want %v", test.name, got, test.want)
+			}
+		})
 	}
 }
 
-func TestDatabase_FromContext_WrongType(t *testing.T) {
-	// setup context
-	gin.SetMode(gin.TestMode)
+func TestDatabase_ToContext(t *testing.T) {
```
jbrockopp, line 65: This function was refactored to use table driven tests.

database/database.go
```diff
@@ -66,38 +68,86 @@ type (
...
+		return nil, err
+	}
+
+	return e, nil
+}
+
+// NewTest creates and returns an engine that integrates with an in-memory database provider.
+//
+// This function is ONLY intended to be used for testing purposes.
+func NewTest() (Interface, error) {
```
jbrockopp, line 142: This will be needed for tests all over the place that have a dependency on `database` logic

i.e. `middleware`, `secret` etc.

database/interface.go
```diff
@@ -18,15 +18,23 @@ import (
...
 type Interface interface {
-	// Database Interface Functions
+	// Generic Interface Functions
 
-	// Driver defines a function that outputs
-	// the configured database driver.
+	// TODO: Add this function to the interface once other code has been updated to use the agnostic engine.
+	//
+	// Close defines a function that stops and terminates the connection to the database.
+	// Close() error
```
jbrockopp, line 28: We can't add this function to the interface until we update all code to call `database.NewTest()`.

That will be added in a separate PR because it will touch a lot of code.

cognifloyd (approved):

ecrupper (approved):

PR #877, opened 2023-06-08
fix(schedules): increment repo counter after read

plyr4:
fixes a double-increment bug. scheduled builds get assigned counter+2 

if it helps, here's the webhook builds code https://github.com/go-vela/server/blob/main/api/webhook/post.go#L472-L484


wass3rw3rk (approved):

KellyMerrick (approved):
lgtm

PR #879, opened 2023-06-08
enhance(log)!: do not return log object for POST and PUT requests

ecrupper:
Sort of a revert of https://github.com/go-vela/server/pull/865 but this PR goes further and simply does not bother returning the log object on a successful `POST` or `PUT`.

I did a little bit of research on general HTTP practices, and it appears that this question of returning data on `200` and `201` codes is pretty much context-driven. Given that `CreateStepLog` + `CreateServiceLog` arguably shouldn't exist for user access and `UpdateStepLog` + `UpdateServiceLog` are locked behind worker permissions / platform admins, I think it's safe to say continuously returning the database object is not worth the overhead. 

The [only logic that utilizes this returned value](https://github.com/go-vela/worker/blob/0bac3d636f4bb0a35ea5b2b020d6d0d7db6343f0/executor/linux/step.go#L326-L331) in the entire codebase is a max-log-size check which can be performed with the in-memory `_log` value that is appended every second.

Merging this PR would result in downstream changes to the sdk-go, worker, and CLI

### DIFF
--- a/api/log/create_service.go
+++ b/api/log/create_service.go
@@ -61,8 +61,6 @@ import (
 // responses:
 //   '201':
 //     description: Successfully created the service logs
-//     schema:
-//       "$ref": "#/definitions/Log"
 //   '400':
 //     description: Unable to create the service logs
 //     schema:
@@ -113,7 +111,7 @@ func CreateServiceLog(c *gin.Context) {
 	input.SetRepoID(r.GetID())
 
 	// send API call to create the logs
-	l, err := database.FromContext(c).CreateLog(input)
+	err = database.FromContext(c).CreateLog(input)
 	if err != nil {
 		retErr := fmt.Errorf("unable to create logs for service %s: %w", entry, err)
 
@@ -122,5 +120,5 @@ func CreateServiceLog(c *gin.Context) {
 		return
 	}
 
-	c.JSON(http.StatusCreated, l)
+	c.JSON(http.StatusCreated, nil)
 }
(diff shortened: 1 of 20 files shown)


wass3rw3rk (approved):
let's go to the moon 🌕

cognifloyd:
Returning the log does seem wasteful.

Today, we only use the return to calculate the log's size, so, what if these endpoints returned the size of the log in a header?

ecrupper:
> Today, we only use the return to calculate the log's size, so, what if these endpoints returned the size of the log in a header?

My thought on this was that we [already have the data to check size before we make the call to update](https://github.com/go-vela/worker/blob/0bac3d636f4bb0a35ea5b2b020d6d0d7db6343f0/executor/linux/step.go#L317), and returning really any data would require decompression. I figured just the successful status code would be enough

jbrockopp (commented):
@ecrupper @wass3rw3rk Do you think we need to mark this as a breaking change?

I'm unsure since the impact is solely scoped to server <-> worker communications.

ecrupper:
> @ecrupper @wass3rw3rk Do you think we need to mark this as a breaking change?

Good idea, even though I doubt it's used at all. I will mark it

jbrockopp (approved):
LGTM

cognifloyd:
> > Today, we only use the return to calculate the log's size, so, what if these endpoints returned the size of the log in a header?
> 
> My thought on this was that we [already have the data to check size before we make the call to update](https://github.com/go-vela/worker/blob/0bac3d636f4bb0a35ea5b2b020d6d0d7db6343f0/executor/linux/step.go#L317), and returning really any data would require decompression. I figured just the successful status code would be enough 
> 

Ah. I forgot about compression.

This should simplify logging in the worker slightly, so I'm in favor of this.

PR #881, opened 2023-06-11
refactor(templates): use yaml for tests instead of base64 encoded json

JordanSussman:
Switch from using base64 encoded files to straight yaml files with the `convertFileToGithubResponse` function.

This changes makes it much easier to understand what the underlying file content is since you no longer need to do something like `echo -n 'base64 encoded content' | base64 -D`.


compiler/native/expand_test.go
```diff
@@ -538,7 +539,7 @@ func TestNative_ExpandStepsMulti(t *testing.T) {
 					"auth_method": "token",
 					"username":    "octocat",
 					"items": []interface{}{
-						map[interface{}]interface{}{string("path"): string("docker"), string("source"): string("secret/docker")},
+						map[interface{}]interface{}{"path": "docker", "source": "secret/docker"},
```
JordanSussman, line 542: The `string` is redundant since is able to infer that anything within `""` is already a string.

ecrupper (approved):

PR #883, opened 2023-06-12
enhance(webhook): handle repository transfer events

ecrupper:
Closes https://github.com/go-vela/community/issues/683

Also fixed a bug where the secrets weren't actually getting transferred because we were searching for secrets belonging to the new repo name, rather than the old. 

Also added a util function to split the full name of a repo to its org and repo respectively since that was used twice.


KellyMerrick (approved):
lgtm

scm/github/testdata/hooks/repository_transferred.json
```diff
@@ -0,0 +1,159 @@
+{
+    "action": "transferred",
+    "changes": {
+        "owner": {
+            "from": {
+                "user": {
+                    "login": "Old-Codertocat",
```
plyr4, line 7: are these supposed to be octocat?

api/webhook/post.go
```diff
@@ -714,7 +714,7 @@ func handleRepositoryEvent(c *gin.Context, m *types.Metadata, h *library.Hook, r
 
 	switch h.GetEventAction() {
 	// if action is rename, go through rename routine
-	case constants.ActionRenamed:
+	case constants.ActionRenamed, "transferred":
```
plyr4, line 717: should we make it a constant?

api/webhook/post.go
```diff
@@ -779,34 +779,16 @@ func handleRepositoryEvent(c *gin.Context, m *types.Metadata, h *library.Hook, r
 // renameRepository is a helper function that takes the old name of the repo,
 // queries the database for the repo that matches that name and org, and updates
 // that repo to its new name in order to preserve it. It also updates the secrets
-// associated with that repo.
+// associated with that repo as well as build links for the UI.
 func renameRepository(h *library.Hook, r *library.Repo, c *gin.Context, m *types.Metadata) (*library.Repo, error) {
 	logrus.Debugf("renaming repository from %s to %s", r.GetPreviousName(), r.GetName())
```
wass3rw3rk, line 784: what do you think about increasing log level here? feels like something that should be in audit and not only visibly in debug mode.

wass3rw3rk (approved):

timhuynh94 (approved):

wass3rw3rk (approved):

PR #896, opened 2023-06-28
feat(database): add integration testing

jbrockopp:
This change adds integration tests for the `database` package.

> NOTE: This is already paying off 😄
>
> * https://github.com/go-vela/server/pull/896/commits/95ed34bc76fbc6527b485516e7d3a6ded128d81c
> * https://github.com/go-vela/server/pull/896/commits/387eed094caf803ced96062ea3326c498ef48f43


database/integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+			if err != nil {
+				t.Errorf("unable to close database engine for %s: %v", test.name, err)
+			}
+		})
+	}
+}
+
+func testWorkers(t *testing.T, db Interface) {
+	// used to track the number of methods we call for workers
+	counter := 0
```
cognifloyd, line 88: This should start at 2 instead of subtracting 2 below. (With an appropriate comment)

database/integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+			}
+
+			err = db.Ping()
+			if err != nil {
+				t.Errorf("unable to ping database engine for %s: %v", test.name, err)
+			}
+
+			t.Run("test_workers", func(t *testing.T) {
+				testWorkers(t, db)
+			})
```
cognifloyd, line 76: So, add one of these for each table of the database?

database/integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+	"time"
+
+	"github.com/go-vela/server/database/worker"
+
+	"github.com/go-vela/types/library"
+)
+
+func TestDatabase_Integration(t *testing.T) {
+	// check if we should skip the integration test
+	if os.Getenv("INTEGRATION") == "" {
```
cognifloyd, line 21: Is this a common way to skip a set of tests by default with go tests? Is `INTEGRATION` a common will known env flag? Or is there a native way to "mark"/"tag" tests with go?

jbrockopp: This article should help - https://konradreiche.com/blog/how-to-separate-integration-tests-in-go

The name of the environment variable is arbitrary assuming the `Makefile` reflects the same.

database/integration_test.go
jbrockopp: Yes that is correct 😃

JordanSussman (approved):

wass3rw3rk (approved):
the way we keep track of "coverage" still feels a bit goofy, but i don't have a better alternative. thanks for all this work!

database/hook/table.go
```diff
@@ -51,7 +51,7 @@ hooks (
 	status       TEXT,
 	link         TEXT,
 	webhook_id   INTEGER,
-	UNIQUE(repo_id, build_id)
+	UNIQUE(repo_id, number)
```
wass3rw3rk, line 54: nice catch :)

wass3rw3rk (approved):

PR #899, opened 2023-06-29
feat: otel tracing

JordanSussman:
this PR adds opt-in [OpenTelemetry](https://opentelemetry.io/docs/concepts/signals/traces/) tracing to the server, often referred to as "OTel".

the main difference is the introduction of the `tracing` package which implements a new set of env flags to enable OTel tracing across (nearly) all of the server's executable code.

### New Flags

| Key                                          | Description | Default Value            |
| ----------------------------- | - | --------------------- |
| `VELA_OTEL_TRACING_ENABLE`                      | enables opt-in tracing for the server | `false` |
| `VELA_OTEL_TRACING_SERVICE_NAME`         | defines the 'service name' attached to all traces | vela-server |
| `VELA_OTEL_EXPORTER_OTLP_ENDPOINT` | defines the fullpath url for using tracing with the jaeger all-in-one | http://jaeger:4318 |
| `VELA_OTEL_TRACING_EXPORTER_SSL_CERT_PATH` | defines the filepath to find certs used to forward traces over https, when not supplied, the traces are forwarded insecurely over http | `NONE` |
| `VELA_OTEL_TRACING_TLS_MIN_VERSION` | optional TLS minimum version requirement to set when communicating with the otel exporter | 1.2 |
| `VELA_OTEL_TRACING_RESOURCE_ATTRIBUTES` | defines the static resource attributes attached to all traces, supplied as a list of `key1=<value>,key2=<value>` pairs |  "process.runtime.name=go" |
| `VELA_OTEL_TRACING_RESOURCE_ENV_ATTRIBUTES` |  defines the resource attributes attached to all traces that are fetched from the environment at runtime, supplied as a list of `key1=<env_key>,key2=<env_key>` pairs | `NONE` |
| `VELA_OTEL_TRACING_SPAN_ATTRIBUTES` | set otel tracestate attributes as a list of `key1=<value>,key2=<value>` pairs | `NONE` |
| `VELA_OTEL_TRACING_TRACESTATE_ATTRIBUTES` | set otel tracestate (span) attributes as a list of `key1=<value>,key2=<value>` pairs | `NONE` |
| `VELA_OTEL_TRACING_SAMPLER_RATELIMIT_PER_SECOND` | defines the rate at which traces can be recorded and forwarded to the collector | 100 |

the sampler resource/span attributes offer the flexibility to use restrictive collectors that key off particular fields.

### Instrumentation Libraries

the code takes advantage of official opentelemetry instrumentation libraries such as [otelgin](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin) and [otelgorm](https://pkg.go.dev/github.com/uptrace/opentelemetry-go-extra/otelgorm) to automatically create detailed trace payloads using Go's built in `context` system. 

### Propagating context.Context

meaning, as long as the `ctx context.Context` is used properly, spans will automatically be generated and attached to the parent trace that initiated the code execution.

this means, if we properly pass along `context` as the first parameter of each function then the instrumentation libraries will magically track the timed execution of these functions. then, if you want additional information attached to spans, you would use `SpanFromContext(ctx)` to retrieve the current span, then you can manipulate the attributes attached to it.

### Trace Sampler Configurations

The `tracing` package allows admins to use env flags to customize the way that Vela samples traces. "Sampling" can be done at the beginning (head sampling) or at the end (tail sampling). For now, Vela samples based on a shared rate limit algorithm using `VELA_OTEL_TRACING_SAMPLER_RATELIMIT_PER_SECOND`

In the future, I would like to add more head sampling algorithms like "TraceID Ratio", and a way to ignore no-op endpoints such as `/health`. 

### Example Env Configuration

```env
# opt-in
VELA_OTEL_TRACING_ENABLE=true

# exporter/collector
VELA_OTEL_EXPORTER_OTLP_ENDPOINT="https://jaeger:6418"
VELA_OTEL_TRACING_EXPORTER_SSL_CERT_PATH="/etc/ssl/certs/ca-certificates.crt"
VELA_OTEL_TRACING_TLS_MIN_VERSION="1.2"

# sampler
VELA_OTEL_TRACING_SAMPLER_RATELIMIT_PER_SECOND=0.2

# trace metadata
VELA_OTEL_TRACING_RESOURCE_ATTRIBUTES: 'resource_attr_key=span_attr_val'
VELA_OTEL_TRACING_RESOURCE_ENV_ATTRIBUTES: 'vela.user-refresh-dur=VELA_USER_REFRESH_TOKEN_DURATION'
VELA_OTEL_TRACING_SPAN_ATTRIBUTES: 'span_attr_key=span_attr_val'
VELA_OTEL_TRACING_TRACESTATE_ATTRIBUTES: 'sometracestate=somevalue'
```

### Example Trace Result

by visiting `<ui>/<org>/builds` it results in a trace that looks like this in Jaeger (`http://localhost:16686/trace/0cd1273b6f73a7dd3bdf84a3d28c43d7`)

Screenshot 2024-09-11 at 2 17 42 PM


tracing/tracer.go
```diff
@@ -0,0 +1,46 @@
...
+
+	res, err := resource.New(ctx, resource.WithAttributes(
+		semconv.ServiceName(c.String("tracing.service.name")),
+	))
+	if err != nil {
+		return nil, err
+	}
+
+	tp := sdktrace.NewTracerProvider(
+		sdktrace.WithSampler(sdktrace.AlwaysSample()),
```
plyr4, line 37: should we use RatioBased sampling? AlwaysSample seems like it might cause excessive overhead.

or... we could somehow expose this as something changeable through platform settings to swap it without rebooting. thinking, in the event tracing causes too much overhead and we want to disable it until we need it

router/middleware/tracing.go
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package middleware
```
ecrupper, line 3: Can we get a corresponding test file for these functions?

ecrupper (changes requested):
Looks good and local testing is going well. Did notice a missing test file

router/middleware/tracing.go
plyr4: done, thanks

ecrupper (dismissed):

tracing/flags.go
```diff
@@ -0,0 +1,80 @@
...
+		Value:   false,
+	},
+
+	// Exporter Flags
+
+	&cli.StringFlag{
+		EnvVars: []string{"VELA_OTEL_TRACING_SERVICE_NAME"},
+		Name:    "tracing.service.name",
+		Usage:   "set otel tracing service name",
+		Value:   "go-otel-server",
```
wass3rw3rk, line 25: should this have vela in its name?

tracing/flags.go
```diff
@@ -0,0 +1,80 @@
...
+	&cli.StringFlag{
+		EnvVars: []string{"VELA_OTEL_TRACING_EXPORTER_SSL_CERT_PATH"},
+		Name:    "tracing.exporter.cert_path",
+		Usage:   "set the path to certs used for communicating with the otel exporter. if not set, will use insecure communication",
+	},
+	&cli.StringFlag{
+		EnvVars: []string{"VELA_OTEL_TRACING_ENDPOINT"},
+		Name:    "tracing.exporter.endpoint",
+		Usage:   "set the otel exporter endpoint",
+		Value:   "127.0.0.1:4318",
```
wass3rw3rk, line 36: should have an empty value as the default?

tracing/flags.go
```diff
@@ -0,0 +1,80 @@
...
+		EnvVars: []string{"VELA_OTEL_TRACING_RESOURCE_ATTRIBUTES"},
+		Name:    "tracing.resource.attributes",
+		Usage:   "set otel resource attributes as a list of key=value pairs. each one will be attached to each span as an attribute",
+		Value:   cli.NewStringSlice("process.runtime.name=go"),
+	},
+	&cli.StringSliceFlag{
+		EnvVars: []string{"VELA_OTEL_TRACING_RESOURCE_ENV_ATTRIBUTES"},
+		Name:    "tracing.resource.env_attributes",
+		Usage:   "set otel resource attributes as a list of key=env_variable_key pairs. each one will be attached to each span as an attribute where the value is retrieved from the environment using the pair value",
+		Value:   cli.NewStringSlice("deployment.environment=CLOUD_ENVIRONMENT"),
```
wass3rw3rk, line 51: should we leave this empty - no default?

tracing/flags.go
```diff
@@ -0,0 +1,80 @@
...
+		EnvVars: []string{"VELA_OTEL_TRACING_SAMPLER_PARENT"},
+		Name:    "tracing.sampler.parent",
+		Usage:   "set otel sampler parent attribute attached to each span.",
+		Value:   "sampler.parent",
+	},
+	&cli.StringFlag{
+		EnvVars: []string{"VELA_OTEL_TRACING_SAMPLER_TYPE"},
+		Name:    "tracing.sampler.type",
+		Usage:   "set otel sampler type attribute attached to each span.",
+		Value:   "sampler.type",
```
wass3rw3rk, line 72: is this the right value? should it be one of https://opentelemetry.io/docs/languages/sdk-configuration/general/#otel_traces_sampler by default? or no default?

tracing/config.go
```diff
@@ -0,0 +1,93 @@
...
+				"sampler.type":   c.String("tracing.sampler.type"),
+			},
+			Ratio:     c.Float64("tracing.sampler.ratio"),
+			PerSecond: c.Float64("tracing.sampler.persecond"),
+		},
+	}
+
+	// add resource attributes
+	for _, attr := range c.StringSlice("tracing.resource.attributes") {
+		kv := strings.Split(attr, "=")
```
wass3rw3rk, line 63: the value on these is unlikely to contain `=`? otherwise we might want to use SplitN

tracing/flags.go
```diff
@@ -0,0 +1,80 @@
...
+		EnvVars: []string{"VELA_OTEL_TRACING_SAMPLER_TRACESTATE"},
+		Name:    "tracing.sampler.tracestate",
+		Usage:   "set otel sampler trace state attached to each span as an attribute.",
+		Value:   "sampler.tracestate",
+	},
+	&cli.StringFlag{
+		EnvVars: []string{"VELA_OTEL_TRACING_SAMPLER_PARENT"},
+		Name:    "tracing.sampler.parent",
+		Usage:   "set otel sampler parent attribute attached to each span.",
+		Value:   "sampler.parent",
```
wass3rw3rk, line 66: i couldn't find info on what values this should be/can be; should this have a default value?

tracing/flags.go
```diff
@@ -0,0 +1,80 @@
...
+		Value:   cli.NewStringSlice("deployment.environment=CLOUD_ENVIRONMENT"),
+	},
+
+	// Sampler Flags
+
+	&cli.StringFlag{
+		EnvVars: []string{"VELA_OTEL_TRACING_SAMPLER_TRACESTATE"},
+		Name:    "tracing.sampler.tracestate",
+		Usage:   "set otel sampler trace state attached to each span as an attribute.",
+		Value:   "sampler.tracestate",
```
wass3rw3rk, line 60: i couldn't find info on what values this should be/can be; should this have a default value?

tracing/flags.go
```diff
@@ -0,0 +1,80 @@
...
+	&cli.StringFlag{
+		EnvVars: []string{"VELA_OTEL_TRACING_SAMPLER_TYPE"},
+		Name:    "tracing.sampler.type",
+		Usage:   "set otel sampler type attribute attached to each span.",
+		Value:   "sampler.type",
+	},
+	&cli.Float64Flag{
+		EnvVars: []string{"VELA_OTEL_TRACING_SAMPLER_RATELIMIT_PER_SECOND"},
+		Name:    "tracing.sampler.persecond",
+		Usage:   "set otel tracing head-sampler rate-limiting to N per second. see: https://opentelemetry.io/docs/concepts/sampling/",
```
wass3rw3rk, line 77: i like the reference to official docs ❤️

docker-compose.yml
```diff
@@ -44,6 +44,13 @@ services:
       VELA_ENABLE_SECURE_COOKIE: 'false'
       VELA_REPO_ALLOWLIST: '*'
       VELA_SCHEDULE_ALLOWLIST: '*'
+      VELA_OTEL_EXPORTER_OTLP_ENDPOINT: http://jaeger:4318
```
wass3rw3rk, line 47: i don't think `VELA_OTEL_EXPORTER_OTLP_ENDPOINT` is used?

tracing/tracer.go
```diff
@@ -0,0 +1,82 @@
...
+			return nil, err
+		}
+
+		certs := x509.NewCertPool()
+		certs.AppendCertsFromPEM(pem)
+
+		withTLS = otlptracehttp.WithTLSClientConfig(
+			&tls.Config{
+				RootCAs:    certs,
+				MinVersion: tls.VersionTLS12,
```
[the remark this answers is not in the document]
plyr4: yeah thats right its for collector servers that are hosted elsewhere, i'll tweak and make it configurable, good call.

docker-compose.yml
plyr4: should be `OTEL_EXPORTER_OTLP_ENDPOINT` for making it easy to use with jaeger all-in-one.

i think we can get away with setting the traces endpoint, let me double check if this is still needed

tracing/tracer.go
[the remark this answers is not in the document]
plyr4: https://github.com/go-vela/server/pull/899/commits/7b27836044eca6e563a85b3490394d0d7fd7b205

tracing/flags.go
plyr4: #0525a143

tracing/flags.go
plyr4: removed this in favor of `VELA_OTEL_EXPORTER_OTLP_ENDPOINT`, intentionally has `VELA_` such that it does NOT get picked up by the library, instead this URL is _parsed_, _validated_ then used in `otlp.WithEndpoint`

docker-compose.yml
plyr4: went with `VELA_OTEL_EXPORTER_OTLP_ENDPOINT` with better parsing/validation before using `WithEndpoint`. this makes everyone happy!

ecrupper (dismissed):

wass3rw3rk (approved):
lgtm

ecrupper (approved):

PR #900, opened 2023-07-06
fix(api): support schedule in filter

wass3rw3rk:
adds "schedule" as an allowed event type for endpoints that accept event filter in query:
* /api/v1/repos/{org}/builds
* /api/v1/repos/{org}/{repo}/builds

also, adds missing query param filters to docs for `/api/v1/repos/{org}/builds`.


timhuynh94 (approved):
👍👌

plyr4 (approved):

PR #916, opened 2023-07-25
fix(compiler): convert local templates to an engine field for nested templates

ecrupper:
When leveraging `vela validate pipeline --template --template-file exampleOne:one.yml,exampleTwo:two.yml` in a nested template sense (exampleTwo is called within exampleOne), the current implementation of local template retrieval can't handle that. 

By adding local templates to the engine itself, we can access the provided info at any time during compilation, allowing proper local retrieval deep within a template chain.


wass3rw3rk (dismissed):

plyr4 (approved):

wass3rw3rk (approved):

KellyMerrick (approved):
lgtm

PR #924, opened 2023-08-04
fix: getting visbility to match on vela and SCM

claire1618:
Corresponds with issue "when enabling a private repo in scm, set it to private in vela as well" #784: 
[https://github.com/go-vela/community/issues/784]


plyr4 (approved):
nice

cognifloyd (approved):

PR #926, opened 2023-08-07
enhance(renovate): keep buildkite/yaml static

wass3rw3rk:


wass3rw3rk:
actually, moving this to go-vela/renovate-config

PR #934, opened 2023-08-22
enhance: adding a branch field to scheduled builds

claire1618:


ecrupper (changes requested):

api/schedule/create.go
```diff
@@ -136,6 +136,7 @@ func CreateSchedule(c *gin.Context) {
 	s.SetCreatedAt(time.Now().UTC().Unix())
 	s.SetUpdatedAt(time.Now().UTC().Unix())
 	s.SetUpdatedBy(u.GetName())
+	s.SetBranch(input.GetBranch())
```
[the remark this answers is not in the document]
claire1618: Good idea, I updated the code to reflect that

KellyMerrick (approved):
lgtm

api/schedule/update.go
```diff
@@ -127,6 +127,7 @@ func UpdateSchedule(c *gin.Context) {
 
 	// set the updated by field using claims
 	s.SetUpdatedBy(u.GetName())
+	s.SetBranch(input.GetBranch())
```
ecrupper, line 130: We'll want to follow the same pattern as above in order to make sure an update payload like this: 
```json
{
    "entry": "0 0 * * *"
}
```
Doesn't result in the branch being set to an empty string.

```go
if input.GetBranch() != "" {
    s.SetBranch(input.GetBranch)
}
```

ecrupper (changes requested):

ecrupper (approved):

api/schedule/update.go
claire1618: I updated the code to reflect that

plyr4 (approved):

PR #943, opened 2023-08-28
Increase Starlark execution limit, abstract limit resolver

colindean:
5,000 was too few to enable the example added to the testdata to work, so was 6,000. I chose 7,500 arbitrarily after a test at 10,000 and both worked.

In the long term, this should probably be configurable so as not to require recompilation. For now, this kicks the can down the road while allowing this build matrix use case to exist.

### DIFF
--- a/compiler/template/starlark/render.go
+++ b/compiler/template/starlark/render.go
@@ -38,7 +38,7 @@ func Render(tmpl string, name string, tName string, environment raw.StringSliceM
 	// arbitrarily limiting the steps of the thread to 5000 to help prevent infinite loops
 	// may need to further investigate spawning a separate POSIX process if user input is problematic
 	// see https://github.com/google/starlark-go/issues/160#issuecomment-466794230 for further details
-	thread.SetMaxExecutionSteps(5000)
+	thread.SetMaxExecutionSteps(GetStarlarkExecutionStepLimit())
 
 	predeclared := starlark.StringDict{"struct": starlark.NewBuiltin("struct", starlarkstruct.Make)}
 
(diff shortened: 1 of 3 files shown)


colindean:
N.b. I didn't put a want.yml with this because it was very time-consuming to copy-paste and fix stuff in an effective manual generation of a very large config.

plyr4:
sweet, thanks for the fix @colindean. 
one note, what do you think about making this value configurable at runtime via flags? 
(picking a random flag as an example: https://github.com/go-vela/server/blob/main/cmd/vela-server/main.go#L107-L112)
that way if 7500 doesnt work we can just apply a new config and redeploy rather than rebuild the source 

also regarding timing, we're planning on slotting this fix into our v0.21 release, which should be right around the corner. we're just wrapping up a couple more things.

ecrupper:
> sweet, thanks for the fix @colindean. one note, what do you think about making this value configurable at runtime via flags? (picking a random flag as an example: https://github.com/go-vela/server/blob/main/cmd/vela-server/main.go#L107-L112) that way if 7500 doesnt work we can just apply a new config and redeploy rather than rebuild the source

@plyr4 I jumped the gun on this and coded up that solution + added some more tests. Link is above on this page^

colindean:
I'd be happy with it either way. Something still hardcoded but at least abstracted is what I could fit into my timebox yesterday ;-)

colindean:
Sure. 

Although I wouldn't mind having the commits in the code if @ecrupper wants to rebase his work on top of mine LOL.

ecrupper:
> Sure.
> 
> Although I wouldn't mind having the commits in the code if @ecrupper wants to rebase his work on top of mine LOL.

@colindean PR is updated with your commits too 👍

colindean:
Aww yisss

PR #947, opened 2023-08-29
enhance(starlark): make execution step limit configurable

ecrupper:
Ref: https://github.com/go-vela/server/pull/943

Since we don't have a good idea on how to potentially calculate a limit, it's a good idea to make this value configurable on start up, rather than incrementing it whenever we run into problems.

Default is set to 7500, though I'm open to other ideas on that 👍 .


colindean (commented):
You might want to have a "Starlark execution environment" config struct where this limit and inevitably other stuff gets shoved. It would perhaps include the environment name, e.g.

```go
thread := &starlark.Thread{Name: "templated-base"}
```

Also, there's still a hardcoded step limit in the `Render` method on around line 41-42.

plyr4 (dismissed):

wass3r (dismissed):

wass3rw3rk (approved):

wass3r:
@colindean what is the use case for controlling the thread name? or was that just arbitrarily chosen as an example for making it more extensible?

colindean:
Arbitrarily chosen. Wouldn't be surprised if there's other things in there that end up needing to be tuned

colindean:
For that config could also include things for controlling the inevitable subprocess, like a timeout, memory limits, etc

KellyMerrick (approved):
lgtm

PR #948, opened 2023-08-29
enhance(queue): adjust pop to allow channels as argument

ecrupper:
This is some groundwork for allowing platform admins to live update routes for workers.

Changing the `Pop` method to accept an optional channels argument gives the worker the ability to leverage the DB record of its routes when listening to the queue rather than what was configured on start up.

There is a question on whether or not this change makes start-up routes obsolete. Admins could use the API to supply routes rather than put them in the container environment. However, since the server leverages `QUEUE_ROUTES` to validate routes prior to publishing, removal of that flag is not an option. The deprecation would be better explained in docs rather than in code.

### DIFF
--- a/queue/redis/pop.go
+++ b/queue/redis/pop.go
@@ -15,13 +15,23 @@ import (
 )
 
 // Pop grabs an item from the specified channel off the queue.
-func (c *client) Pop(ctx context.Context) (*types.Item, error) {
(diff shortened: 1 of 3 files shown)


wass3rw3rk (approved):

KellyMerrick (approved):
lgtm

PR #952, opened 2023-09-01
enhance(integration-test): leverage a custom implementation of DeepEqual

ecrupper:
What are some thoughts on this? It will help eliminate those pesky time skew failing workflows. Not sure this is the best implementation of the concept either, so any advice would be appreciated too.

### DIFF
--- a/database/integration_test.go
+++ b/database/integration_test.go
@@ -1760,7 +1760,7 @@ func testWorkers(t *testing.T, db Interface, resources *Resources) {
 
 	// create the workers
 	for _, worker := range resources.Workers {
-		_, err := db.CreateWorker(worker)
+		_, err := db.CreateWorker(context.TODO(), worker)
 		if err != nil {
 			t.Errorf("unable to create worker %d: %v", worker.GetID(), err)
 		}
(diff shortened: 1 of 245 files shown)


wass3rw3rk:
love the idea - can we leverage https://pkg.go.dev/github.com/google/go-cmp/cmp somehow for this? we already import it.

wass3rw3rk:
I see the edits, but I was thinking of it a bit differently. Can we just use cmp.Equal at the top level and define a custom Comparer to handle UpdatedAt to replace your custom function and switch statement?

database/integration_test.go
```diff
@@ -2317,3 +2318,37 @@ func newResources() *Resources {
 		Workers:     []*library.Worker{workerOne, workerTwo},
 	}
 }
+
+// CmpOptApproxUpdatedAt is a custom comparator for cmp.Equal
+// to reduce flakiness in tests when comparing structs with UpdatedAt field.
+func CmpOptApproxUpdatedAt() cmp.Option {
+	// Custom Comparer
```
wass3rw3rk, line 2325:
```suggestion
	// Custom Comparer for *int64 fields typically used to store unix timestamps.
	// Will consider time difference of 5s to be equal for sake of tests.
```

wass3rw3rk (approved):

cognifloyd (approved):

PR #953, opened 2023-09-01
bug(repo sync): ignore inactive repos

KellyMerrick:
This closes https://github.com/go-vela/community/issues/852.

Add bool to track if webhook has been deleted from SCM:
https://github.com/go-vela/server/blob/f89e9299cf7b38ca9bf097e288769df406958ce4/scm/github/repo.go#L263-L274

And then update status to inactive in the database:
https://github.com/go-vela/server/blob/f89e9299cf7b38ca9bf097e288769df406958ce4/api/scm/sync_org.go#L140-L164

### DIFF
--- a/api/scm/sync.go
+++ b/api/scm/sync.go
@@ -125,7 +125,7 @@ func SyncRepo(c *gin.Context) {
 		}
 
 		// update webhook
-		err = scm.FromContext(c).Update(u, r, lastHook.GetWebhookID())
+		_, err = scm.FromContext(c).Update(u, r, lastHook.GetWebhookID())
 		if err != nil {
 			retErr := fmt.Errorf("unable to update repo webhook for %s: %w", r.GetFullName(), err)
 
--- a/api/scm/sync.go
+++ b/api/scm/sync.go
@@ -111,8 +111,9 @@ func SyncRepo(c *gin.Context) {
 		return
 	}
 
-	// if we have webhook validation, update the repo hook in the SCM
-	if c.Value("webhookvalidation").(bool) {
+	// if we have webhook validation and the repo is active in the database,
+	// update the repo hook in the SCM
+	if c.Value("webhookvalidation").(bool) && r.GetActive() {
 		// grab last hook from repo to fetch the webhook ID
 		lastHook, err := database.FromContext(c).LastHookForRepo(ctx, r)
 		if err != nil {
--- a/api/scm/sync.go
+++ b/api/scm/sync.go
@@ -125,13 +125,27 @@ func SyncRepo(c *gin.Context) {
 		}
 
 		// update webhook
-		_, err = scm.FromContext(c).Update(u, r, lastHook.GetWebhookID())
+		webhookExists, err := scm.FromContext(c).Update(u, r, lastHook.GetWebhookID())
 		if err != nil {
 			retErr := fmt.Errorf("unable to update repo webhook for %s: %w", r.GetFullName(), err)
 
 			util.HandleError(c, http.StatusInternalServerError, retErr)
 
-			return
+			// if webhook has been manually deleted from GitHub,
+			// set to inactive in database
+			if !webhookExists {
+
+				r.SetActive(false)
+
+				_, err := database.FromContext(c).UpdateRepo(ctx, r)
+				if err != nil {
+					retErr := fmt.Errorf("unable to update repo for org %s: %w", o, err)
+
+					util.HandleError(c, http.StatusInternalServerError, retErr)
+
+					return
+				}
+			}
 		}
 	}
 
--- a/api/scm/sync.go
+++ b/api/scm/sync.go
@@ -127,9 +127,6 @@ func SyncRepo(c *gin.Context) {
 		// update webhook
 		webhookExists, err := scm.FromContext(c).Update(u, r, lastHook.GetWebhookID())
 		if err != nil {
-			retErr := fmt.Errorf("unable to update repo webhook for %s: %w", r.GetFullName(), err)
-
-			util.HandleError(c, http.StatusInternalServerError, retErr)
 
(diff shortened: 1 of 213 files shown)


api/scm/sync.go
```diff
@@ -124,7 +125,7 @@ func SyncRepo(c *gin.Context) {
 		}
 
 		// update webhook
-		err = scm.FromContext(c).Update(u, r, lastHook.GetWebhookID())
+		_, err = scm.FromContext(c).Update(u, r, lastHook.GetWebhookID())
```
ecrupper, line 128: I say we do the same think at the repo level as we do the org level (i.e. change the repo to inactive if the webhook isn't found when attempting to update)

scm/github/repo.go
```diff
@@ -258,13 +258,20 @@ func (c *client) Update(u *library.User, r *library.Repo, hookID int64) error {
...
+		isWebhookDel = true
+	}
 
 	if err != nil {
-		return err
+		return isWebhookDel, err
 	}
 
-	return nil
+	return isWebhookDel, nil
```
ecrupper, line 274: Thoughts on shuffling this a bit to
```go
	if err != nil {
		return resp.StatusCode == http.StatusNotFound, err
	}

	return true, nil
```

That way the affirmative `true` is when the webhook does exist, and `false` is when the webhook doesn't exist. Or we could return the code itself and let the API do the logic? Not really sure the cleanest way...

ecrupper (changes requested):

scm/github/repo.go
KellyMerrick: I did implement the inverse for now, for mvp.

api/scm/sync.go
```diff
@@ -124,13 +125,27 @@ func SyncRepo(c *gin.Context) {
 		}
 
 		// update webhook
-		err = scm.FromContext(c).Update(u, r, lastHook.GetWebhookID())
+		webhookExists, err := scm.FromContext(c).Update(u, r, lastHook.GetWebhookID())
 		if err != nil {
 			retErr := fmt.Errorf("unable to update repo webhook for %s: %w", r.GetFullName(), err)
 
 			util.HandleError(c, http.StatusInternalServerError, retErr)
```
ecrupper, line 132: I don't think we want these lines in the error block unless the error is something other than `Not Found`

api/scm/sync_org.go
```diff
@@ -137,12 +138,28 @@ func SyncReposForOrg(c *gin.Context) {
 			}
 
 			// update webhook
-			err = scm.FromContext(c).Update(u, repo, lastHook.GetWebhookID())
+			webhookExists, err := scm.FromContext(c).Update(u, repo, lastHook.GetWebhookID())
 			if err != nil {
 				retErr := fmt.Errorf("unable to update repo webhook for %s: %w", repo.GetFullName(), err)
 
 				util.HandleError(c, http.StatusInternalServerError, retErr)
```
ecrupper, line 145: Same idea as above

api/scm/sync_org.go
```diff
@@ -137,12 +138,28 @@ func SyncReposForOrg(c *gin.Context) {
...
 
+				// if webhook has been manually deleted from GitHub,
+				// set to inactive in database
+				if !webhookExists {
+
+					repo.SetActive(false)
+
+					_, err := database.FromContext(c).UpdateRepo(ctx, repo)
```
ecrupper, line 153: We'll want a `continue` after the successful execution of this update so we can continue updating repos if one gets tripped up by this `Not Found` business

ecrupper (changes requested):

scm/github/repo.go
```diff
@@ -258,13 +258,20 @@ func (c *client) Update(u *library.User, r *library.Repo, hookID int64) error {
...
+	}
 
 	if err != nil {
-		return err
+		return webhookExists, err
 	}
 
-	return nil
+	return webhookExists, nil
 }
```
wass3rw3rk, line 275: suggestion: we can just collapse this to `return webhookExists, err` and skip the error check.

api/scm/sync.go
KellyMerrick: addressed

api/scm/sync_org.go
KellyMerrick: addressed

api/scm/sync_org.go
KellyMerrick: addressed

scm/github/repo.go
KellyMerrick: addressed

ecrupper (dismissed):

wass3rw3rk (approved):

ecrupper (dismissed):

scm/github/repo.go
```diff
@@ -258,13 +258,11 @@ func (c *client) Update(u *library.User, r *library.Repo, hookID int64) error {
...
-	_, _, err := client.Repositories.EditHook(ctx, r.GetOrg(), r.GetName(), hookID, hook)
+	_, resp, err := client.Repositories.EditHook(ctx, r.GetOrg(), r.GetName(), hookID, hook)
 
-	if err != nil {
-		return err
-	}
-
-	return nil
+	// track if webhook exists in GitHub; a missing webhook
+	// indicates the webhook has been manually deleted from GitHu
```
wass3r, line 264: RIP "b" on GitHu(b)

wass3r (dismissed):

scm/github/repo.go
ecrupper: It was manually deleted

wass3r (approved):

scm/github/repo.go
KellyMerrick: bfixed it

ecrupper (approved):

PR #955, opened 2023-09-05
fix(webhook): repo transfer handling with renaming - support newer GitHub versions

ecrupper:
With newer GitHub versions allowing users to transfer _and_ rename in one fell swoop, there happened to be a bug in the payloads that are sent our way.

GitHub offered this `transfer` + `rename` option with the following events under the hood (example involves the transfer of `ExampleOrg/name` to `EastonCrupper/new-name`):

```json
{
  "action": "transferred",
  "changes": {
    "owner": {
      "from": {
        "user": {
          "login": "EastonCrupper",
          ....
        }
      }
    }
  },
  "repository": {
    "id": 352163,
    "name": "new-name",
.......
```

followed by

```json
{
  "action": "renamed",
  "changes": {
    "repository": {
      "name": {
        "from": "name"
      }
    }
  },
  "repository": {
    "id": 352163,
    "name": "new-name",
    "full_name": "EastonCrupper/new-name",
    "private": false,
    "owner": {
      "login": "EastonCrupper",
.......
```

The issue: **we have no way to find the old repository's name during the transfer event.** One would assume the first payload would contain info about the organization change and _only_ the organization change, and, in turn, the rename payload would contain info about the name change. This is unfortunately not the case and is surely a bug with GitHub.

This issue persists in the newest versions of GitHub, too.

Therefore, this change suggests we find our DB record of the repo in a different way: by using the `webhook_id` conveniently stored in all hooks, which is a unique identifier for repos. 

If I could do this over again, I would remove `repo.PreviousName` and replace it with `repo.WebhookID`, and rework the redelivery mechanism to find the webhook ID from the repo rather than the hook. Something to ponder....

### DIFF
--- a/database/hook/interface.go
+++ b/database/hook/interface.go
@@ -38,7 +38,7 @@ type HookInterface interface {
 	DeleteHook(context.Context, *library.Hook) error
 	// GetHook defines a function that gets a hook by ID.
 	GetHook(context.Context, int64) (*library.Hook, error)
-	// GetHookByWebhook defines a function that gets any hook with a matching webhook_id.
+	// GetHookByWebhookID defines a function that gets any hook with a matching webhook_id.
 	GetHookByWebhookID(context.Context, int64) (*library.Hook, error)
 	// GetHookForRepo defines a function that gets a hook by repo ID and number.
 	GetHookForRepo(context.Context, *library.Repo, int) (*library.Hook, error)
--- a/database/hook/interface.go
+++ b/database/hook/interface.go
@@ -39,7 +39,7 @@ type HookInterface interface {
 	// GetHook defines a function that gets a hook by ID.
 	GetHook(context.Context, int64) (*library.Hook, error)
 	// GetHookByWebhook defines a function that gets any hook with a matching webhook_id.
-	GetHookByWebhook(context.Context, int64) (*library.Hook, error)
+	GetHookByWebhookID(context.Context, int64) (*library.Hook, error)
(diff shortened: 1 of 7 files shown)


database/integration_test.go
```diff
@@ -494,7 +494,7 @@ func testHooks(t *testing.T, db Interface, resources *Resources) {
 	if err != nil {
 		t.Errorf("unable to list hooks for repo %d: %v", resources.Repos[0].GetID(), err)
 	}
-	if int(count) != len(resources.Hooks) {
+	if int(count) != len(resources.Hooks)-1 {
```
wass3rw3rk, line 497: do you mind adding a quick comment here on why it's `-1`? i know the reason, but it took me a minute :D

wass3rw3rk (changes requested):

database/hook/interface.go
```diff
@@ -38,6 +38,8 @@ type HookInterface interface {
 	DeleteHook(context.Context, *library.Hook) error
 	// GetHook defines a function that gets a hook by ID.
 	GetHook(context.Context, int64) (*library.Hook, error)
+	// GetHookByWebhook defines a function that gets any hook with a matching webhook_id.
```
wass3rw3rk, line 41:
```suggestion
	// GetHookByWebhookID defines a function that gets any hook with a matching webhook_id.
```

wass3rw3rk (changes requested):

wass3rw3rk (approved):

PR #957, opened 2023-09-06
feat(queue)!: register endpoint to use QueueInfo type

timhuynh94:
Changes:
- give it the QUEUE_PUBLIC_KEY
- embed public key and queue address into the Gin context
- in api/queue/queue.go pass along (1) pub key (2) queue address in a QueueInfo struct

### DIFF
--- a/api/queue/queue.go
+++ b/api/queue/queue.go
@@ -33,9 +33,9 @@ import (
 //     schema:
 //       "$ref": "#/definitions/Error"
 
-// Registration represents the API handler to
-// provide queue credentials after worker onboarded with registration token.
-func Registration(c *gin.Context) {
+// Register represents the API handler to
+// retrieve queue credentials as part of worker onboarding.
+func Register(c *gin.Context) {
 	cl := claims.Retrieve(c)
 
 	logrus.WithFields(logrus.Fields{
--- a/api/queue/queue.go
+++ b/api/queue/queue.go
@@ -44,14 +44,14 @@ func Registration(c *gin.Context) {
 	}).Info("requesting queue credentials with registration token")
 
 	// extract the public key that was packed into gin context
-	k := c.MustGet("public-key").(*string)
+	k := c.MustGet("public-key").(string)
 
 	// extract the queue-address that was packed into gin context
-	a := c.MustGet("queue-address").(*string)
+	a := c.MustGet("queue-address").(string)
 
 	wr := library.QueueRegistration{
-		QueuePublicKey: k,
-		QueueAddress:   a,
+		QueuePublicKey: &k,
+		QueueAddress:   &a,
 	}
 
 	c.JSON(http.StatusOK, wr)
--- a/api/queue/queue.go
+++ b/api/queue/queue.go
@@ -25,7 +25,7 @@ import (
 //   - ApiKeyAuth: []
 // responses:
 //   '200':
-//     description: Successfully generated registration token
+//     description: Successfully retrieved queue credentials
 //     schema:
 //       "$ref": "#/definitions/Token"
 //   '401':
--- a/api/queue/queue.go
+++ b/api/queue/queue.go
@@ -26,15 +26,15 @@ import (
 //   '200':
 //     description: Successfully retrieved queue credentials
 //     schema:
-//       "$ref": "#/definitions/QueueRegistration"
+//       "$ref": "#/definitions/QueueInfo"
 //   '401':
 //     description: Unauthorized
 //     schema:
 //       "$ref": "#/definitions/Error"
 
-// Register represents the API handler to
+// Info represents the API handler to
 // retrieve queue credentials as part of worker onboarding.
-func Register(c *gin.Context) {
+func Info(c *gin.Context) {
 	cl := claims.Retrieve(c)
 
 	logrus.WithFields(logrus.Fields{
@@ -47,7 +47,7 @@ func Register(c *gin.Context) {
 	// extract the queue-address that was packed into gin context
 	a := c.MustGet("queue-address").(string)
 
-	wr := library.QueueRegistration{
+	wr := library.QueueInfo{
 		QueuePublicKey: &k,
 		QueueAddress:   &a,
 	}
--- a/api/queue/queue.go
+++ b/api/queue/queue.go
@@ -33,9 +33,9 @@ import (
 //     schema:
 //       "$ref": "#/definitions/Error"
 
-// QueueRegistration represents the API handler to
+// Registration represents the API handler to
 // provide queue credentials after worker onboarded with registration token.
-func QueueRegistration(c *gin.Context) {
+func Registration(c *gin.Context) {
 	cl := claims.Retrieve(c)
 	// retrieve user from context
 
--- a/api/queue/queue.go
+++ b/api/queue/queue.go
@@ -37,7 +37,6 @@ import (
 // provide queue credentials after worker onboarded with registration token.
 func Registration(c *gin.Context) {
 	cl := claims.Retrieve(c)
-	// retrieve user from context
 
 	logrus.WithFields(logrus.Fields{
 		"user": cl.Subject,
--- a/api/queue/queue.go
+++ b/api/queue/queue.go
@@ -14,19 +14,13 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// swagger:operation POST /api/v1/admin/queue-credentials admin QueueRegistration
+// swagger:operation POST /api/v1/admin/queue-registration admin QueueRegistration
 //
 // Get queue credentials
 //
 // ---
 // produces:
 // - application/json
-// parameters:
-// - in: path
-//   name: worker
-//   description: Hostname of the worker
-//   required: true
-//   type: string
 // security:
 //   - ApiKeyAuth: []
 // responses:
@@ -76,6 +70,6 @@ func QueueRegistration(c *gin.Context) {
 		QueuePublicKey: &pk,
 		QueueAddress:   &qa,
 	}
-	logrus.Infof("current creds %s: ", wr)
+
 	c.JSON(http.StatusOK, wr)
 }
--- a/api/queue/queue.go
+++ b/api/queue/queue.go
@@ -44,26 +44,14 @@ func QueueRegistration(c *gin.Context) {
 	}).Info("requesting queue credentials with registration token")
 
 	// extract the public key that was packed into gin context
-	k := c.MustGet("public-key")
+	k := c.MustGet("public-key").(*string)
 
 	// extract the queue-address that was packed into gin context
-	a := c.MustGet("queue-address")
-
-	pk, ok := k.(string)
-	if !ok {
-		c.JSON(http.StatusInternalServerError, "public key in the context is the wrong type")
-		return
-	}
-
-	qa, ok := a.(string)
-	if !ok {
-		c.JSON(http.StatusInternalServerError, "queue address in the context is the wrong type")
-		return
-	}
+	a := c.MustGet("queue-address").(*string)
 
 	wr := library.QueueRegistration{
-		QueuePublicKey: &pk,
-		QueueAddress:   &qa,
+		QueuePublicKey: k,
+		QueueAddress:   a,
 	}
 
 	c.JSON(http.StatusOK, wr)
--- a/api/queue/queue.go
+++ b/api/queue/queue.go
@@ -27,7 +27,7 @@ import (
 //   '200':
 //     description: Successfully retrieved queue credentials
 //     schema:
-//       "$ref": "#/definitions/Token"
+//       "$ref": "#/definitions/QueueRegistration"
 //   '401':
 //     description: Unauthorized
 //     schema:
--- a/api/queue/queue.go
+++ b/api/queue/queue.go
@@ -13,7 +13,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// swagger:operation POST /api/v1/queue/register queue Registration
+// swagger:operation POST /api/v1/queue/info queue Info
 //
 // Get queue credentials
 //
--- a/api/queue/queue.go
+++ b/api/queue/queue.go
@@ -5,16 +5,16 @@
 package queue
 
 import (
+	"github.com/go-vela/server/router/middleware/claims"
 	"net/http"
 
-	"github.com/go-vela/server/router/middleware/user"
 	"github.com/go-vela/types/library"
 
 	"github.com/gin-gonic/gin"
 	"github.com/sirupsen/logrus"
 )
 
-// swagger:operation POST /api/v1/admin/queue-registration admin QueueRegistration
+// swagger:operation POST /api/v1/queue/queue-registration admin QueueRegistration
 //
 // Get queue credentials
 //
@@ -36,23 +36,18 @@ import (
 // QueueRegistration represents the API handler to
 // provide queue credentials after worker onboarded with registration token.
 func QueueRegistration(c *gin.Context) {
+	cl := claims.Retrieve(c)
 	// retrieve user from context
-	u := user.Retrieve(c)
 
-	logrus.Infof("Platform admin %s: fetching queue creds", u.GetName())
+	logrus.WithFields(logrus.Fields{
+		"user": cl.Subject,
+	}).Info("requesting queue credentials with registration token")
 
 	// extract the public key that was packed into gin context
-	k, ok := c.Get("public-key")
-	if !ok {
-		c.JSON(http.StatusInternalServerError, "no public-key in the context")
-		return
-	}
+	k := c.MustGet("public-key")
+
 	// extract the queue-address that was packed into gin context
-	a, ok := c.Get("queue-address")
-	if !ok {
-		c.JSON(http.StatusInternalServerError, "no queue-address in the context")
-		return
-	}
+	a := c.MustGet("queue-address")
 
 	pk, ok := k.(string)
 	if !ok {
--- a/api/queue/queue.go
+++ b/api/queue/queue.go
@@ -5,12 +5,11 @@
 package queue
 
 import (
-	"github.com/go-vela/server/router/middleware/claims"
 	"net/http"
 
-	"github.com/go-vela/types/library"
-
 	"github.com/gin-gonic/gin"
+	"github.com/go-vela/server/router/middleware/claims"
+	"github.com/go-vela/types/library"
 	"github.com/sirupsen/logrus"
 )
 
--- a/api/queue/queue.go
+++ b/api/queue/queue.go
@@ -14,7 +14,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// swagger:operation POST /api/v1/queue/queue-registration admin QueueRegistration
+// swagger:operation POST /api/v1/queue/register queue Registration
 //
 // Get queue credentials
 //
--- a/api/queue/queue.go
+++ b/api/queue/queue.go
@@ -2,7 +2,7 @@
 //
 // Use of this source code is governed by the LICENSE file in this repository.
 
-package admin
+package queue
 
 import (
 	"net/http"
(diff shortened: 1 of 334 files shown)


router/admin.go
```diff
@@ -59,6 +60,6 @@ func AdminHandlers(base *gin.RouterGroup) {
 		_admin.PUT("/user", admin.UpdateUser)
 
 		// Admin worker endpoint
-		_admin.POST("/workers/:worker/register-token", admin.RegisterToken)
+		_admin.POST("/workers/:worker/register", admin.WorkerRegistration)
```
ecrupper, line 63: I like the change, but this will have downstream impact on the SDK. Just marking it here so we don't forget.

ecrupper (dismissed):

wass3rw3rk (changes requested):
this probably should be marked as a breaking change?

also.. and maybe this can be a separate PR, but we might want to consider adding validation in /server/queue/setup.go for the private/public key queue flag values?

router/middleware/signing.go
```diff
@@ -16,3 +16,21 @@ func QueueSigningPrivateKey(key string) gin.HandlerFunc {
 		c.Next()
 	}
 }
+
+// QueueSigningPublicKey is a middleware function that attaches the public key used
+// to opened signed items that are pushed to the queue.
```
cognifloyd, line 21:
```suggestion
// to open signed items that are pushed to the queue.
```

api/queue/queue.go
```diff
@@ -0,0 +1,75 @@
...
+	"net/http"
+
+	"github.com/go-vela/server/router/middleware/user"
+	"github.com/go-vela/types/library"
+
+	"github.com/gin-gonic/gin"
+	"github.com/sirupsen/logrus"
+)
+
+// swagger:operation POST /api/v1/admin/queue-registration admin QueueRegistration
```
ecrupper, line 17: Change route to be in queue path

api/queue/queue.go
```diff
@@ -0,0 +1,75 @@
...
+//     schema:
+//       "$ref": "#/definitions/Error"
+
+// QueueRegistration represents the API handler to
+// provide queue credentials after worker onboarded with registration token.
+func QueueRegistration(c *gin.Context) {
+	// retrieve user from context
+	u := user.Retrieve(c)
+
+	logrus.Infof("Platform admin %s: fetching queue creds", u.GetName())
```
ecrupper, line 42: Let's leverage the normal logging format that we use:
```go
logrus.WithFields(logrus.Fields{
		"user":  cl.Subject,
	}).Info("requesting queue credentials with registration token")
```

Note: you'll have to use claims rather than user

api/queue/queue.go
```diff
@@ -0,0 +1,75 @@
...
+	if !ok {
+		c.JSON(http.StatusInternalServerError, "no public-key in the context")
+		return
+	}
+	// extract the queue-address that was packed into gin context
+	a, ok := c.Get("queue-address")
+	if !ok {
+		c.JSON(http.StatusInternalServerError, "no queue-address in the context")
+		return
+	}
```
ecrupper, line 55: I think we can use `c.MustGet` here since we validate the existence of these values in `queue/setup`.

api/queue/queue.go
```diff
@@ -0,0 +1,75 @@
...
+	if !ok {
+		c.JSON(http.StatusInternalServerError, "public key in the context is the wrong type")
+		return
+	}
+
+	qa, ok := a.(string)
+	if !ok {
+		c.JSON(http.StatusInternalServerError, "queue address in the context is the wrong type")
+		return
+	}
```
ecrupper, line 67: I don't think the type check is necessary. The start-up validation should be enough, especially with a length check too.

mock/server/server.go
```diff
@@ -137,5 +137,8 @@ func FakeHandler() http.Handler {
 	e.POST("/authenticate/token", getAuthenticateFromToken)
 	e.GET("/validate-token", validateToken)
 
+	// mock endpoint for queue credentials
+	e.POST("/api/v1/queue/queue-registration", getQueueCreds)
```
ecrupper, line 141:
```suggestion
	e.POST("/api/v1/queue/register", getQueueCreds)
```

Avoids a stutter in the path. Just a suggestion though.

mock/server/worker.go
```diff
@@ -199,3 +206,13 @@ func registerToken(c *gin.Context) {
 
 	c.JSON(http.StatusCreated, body)
 }
+
+// getQueueCreds returns mock JSON for a http POST.
+func getQueueCreds(c *gin.Context) {
```
ecrupper, line 211: This mock should have a fail option for testing. Example: https://github.com/go-vela/server/blob/7ee3ae39552c14a2eb7777195c23a15cd2bf4059/mock/server/worker.go#L181-L201

queue/setup.go
```diff
@@ -92,6 +92,10 @@ func (s *Setup) Validate() error {
 		return fmt.Errorf("no queue routes provided")
 	}
 
+	if len(s.PublicKey) == 0 {
```
ecrupper, line 95: I would add the 32 length check that we see in the pop function: https://github.com/go-vela/server/blob/7ee3ae39552c14a2eb7777195c23a15cd2bf4059/queue/redis/pop.go#L50. Then I would also say that line can be removed once the logic is moved here

ecrupper (changes requested):
Noted a few things.

I also think we could validate the presence of the private key in `validate.go` in `cmd/vela-server`. This way the worker won't be validating against that but we can still ensure the server is started up with proper values.

mock/server/server.go
timhuynh94: How about register-queue? So that we can distinguish between token and queue?

queue/setup.go
timhuynh94: The s.PublicKey here is the encoded string where c.config.PublicKey is a decoded string. In setup.go, we check for whether the encoded public key is provided upon startup or not.

mock/server/worker.go
```diff
@@ -199,3 +206,22 @@ func registerToken(c *gin.Context) {
...
+	q := c.Param("queue")
```
ecrupper, line 212: I don't think this is possible since the URL path doesn't have a queue parameter

router/queue.go
```diff
@@ -0,0 +1,23 @@
...
+
+// QueueHandlers is a function that extends the provided base router group
+// with the API handlers for queue registration functionality.
+//
+// POST   /api/v1/queue/queue-registration.
+func QueueHandlers(base *gin.RouterGroup) {
+	// Workers endpoints
+	_queue := base.Group("/queue")
+	{
+		_queue.POST("/register", perm.MustWorkerRegisterToken(), queue.Registration)
```
ecrupper, line 21: I think this should be a `GET` right? Since we're requesting static data.

ecrupper (changes requested):
Couple things

api/queue/queue.go
```diff
@@ -0,0 +1,58 @@
...
+	// extract the public key that was packed into gin context
+	k := c.MustGet("public-key").(*string)
+
+	// extract the queue-address that was packed into gin context
+	a := c.MustGet("queue-address").(*string)
+
+	wr := library.QueueRegistration{
+		QueuePublicKey: k,
+		QueueAddress:   a,
+	}
```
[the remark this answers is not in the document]
timhuynh94: I just realized this as well! Thanks for pointing it out.

mock/server/worker.go
timhuynh94: Yea it's not possible! I tried! You have any suggestions here?

ecrupper: I would mock a failed attempt by having a user pass in no token or something and then expect an error

timhuynh94: then would be in sdk-go that is using the mock. getQueueCreds has a mustRegToken perm attached.

router/admin.go
```diff
@@ -23,7 +23,8 @@ import (
 // PUT    /api/v1/admin/secret
 // PUT    /api/v1/admin/service
 // PUT    /api/v1/admin/step
-// PUT    /api/v1/admin/user.
+// PUT    /api/v1/admin/user
+// POST   /api/v1/admin/workers/:worker/register-token.
```
plyr4, line 27:
```suggestion
// POST   /api/v1/admin/workers/:worker/register.
```

router/queue.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"github.com/gin-gonic/gin"
+	"github.com/go-vela/server/api/queue"
+	"github.com/go-vela/server/router/middleware/perm"
+)
+
+// QueueHandlers is a function that extends the provided base router group
+// with the API handlers for queue registration functionality.
+//
+// POST   /api/v1/queue/queue-registration.
```
plyr4, line 16:
```suggestion
// POST   /api/v1/queue/register.
```

plyr4 (changes requested):

api/queue/queue.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/go-vela/server/router/middleware/claims"
+	"net/http"
+
+	"github.com/go-vela/types/library"
+
+	"github.com/gin-gonic/gin"
+	"github.com/sirupsen/logrus"
+)
+
+// swagger:operation POST /api/v1/queue/queue-registration admin QueueRegistration
```
plyr4, line 17:
```suggestion
// swagger:operation POST /api/v1/queue/register queue Registration
```

api/queue/queue.go
```diff
@@ -0,0 +1,58 @@
...
+// Get queue credentials
+//
+// ---
+// produces:
+// - application/json
+// security:
+//   - ApiKeyAuth: []
+// responses:
+//   '200':
+//     description: Successfully generated registration token
```
plyr4, line 28:
```suggestion
//     description: Successfully retrieved queue credentials
```

api/queue/queue.go
```diff
@@ -0,0 +1,58 @@
...
+//     schema:
+//       "$ref": "#/definitions/Token"
```
plyr4, line 30: should we point this ref to the QueueRegistration struct?

api/queue/queue.go
```diff
@@ -0,0 +1,58 @@
...
+//   '401':
+//     description: Unauthorized
+//     schema:
+//       "$ref": "#/definitions/Error"
+
+// Registration represents the API handler to
+// provide queue credentials after worker onboarded with registration token.
+func Registration(c *gin.Context) {
+	cl := claims.Retrieve(c)
+	// retrieve user from context
```
plyr4, line 40:
```suggestion
```

api/queue/queue.go
```diff
@@ -0,0 +1,58 @@
...
+//     schema:
+//       "$ref": "#/definitions/Token"
+//   '401':
+//     description: Unauthorized
+//     schema:
+//       "$ref": "#/definitions/Error"
+
+// Registration represents the API handler to
+// provide queue credentials after worker onboarded with registration token.
+func Registration(c *gin.Context) {
```
plyr4, line 38: personal preference:
```suggestion
// Register represents the API handler to
// retrieve queue credentials as part of worker onboarding.
func Register(c *gin.Context) {
```

api/queue/queue.go
plyr4: also personal preference, related to my other comment:
```suggestion
// swagger:operation POST /api/v1/queue/queue-registration queue Register
```

plyr4 (changes requested):

api/queue/queue.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"github.com/go-vela/server/router/middleware/claims"
+	"net/http"
+
+	"github.com/go-vela/types/library"
+
+	"github.com/gin-gonic/gin"
+	"github.com/sirupsen/logrus"
+)
```
wass3rw3rk, line 15: suggestion: sort imports to match how we do it elsewhere.

mock/server/worker.go
```diff
@@ -70,11 +70,18 @@ const (
 	}`
 
 	// RegisterTokenResp represents a JSON return for an admin requesting a registration token.
-	//
-	//nolint:gosec // not actual credentials
+	// not actual credentials
 	RegisterTokenResp = `{
```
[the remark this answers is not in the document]
wass3rw3rk: suggestion: revert this change? the previous directive (`//nolint:gosec // not actual credentials`) should take care of the warning here

router/queue.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/go-vela/server/api/queue"
+	"github.com/go-vela/server/router/middleware/perm"
+)
+
+// QueueHandlers is a function that extends the provided base router group
+// with the API handlers for queue registration functionality.
+//
+// POST   /api/v1/queue/register.
+func QueueHandlers(base *gin.RouterGroup) {
+	// Workers endpoints
```
wass3rw3rk, line 18:
```suggestion
	// Queue endpoints
```

queue/redis/opts.go
```diff
@@ -96,6 +96,10 @@ func WithPrivateKey(key string) ClientOpt {
 		c.config.PrivateKey = new([64]byte)
 		copy(c.config.PrivateKey[:], decoded)
 
+		if len(*c.config.PrivateKey) != 64 {
+			return errors.New("no valid signing private key provided")
```
wass3rw3rk, line 100:
```suggestion
			return errors.New("no valid queue signing private key provided")
```

queue/redis/opts.go
```diff
@@ -132,6 +136,10 @@ func WithPublicKey(key string) ClientOpt {
 		c.config.PublicKey = new([32]byte)
 		copy(c.config.PublicKey[:], decoded)
 
+		if len(*c.config.PublicKey) != 32 {
+			return errors.New("no valid signing public key provided")
```
wass3rw3rk, line 140:
```suggestion
			return errors.New("no valid queue public key provided")
```

mock/server/worker.go
```diff
@@ -70,11 +70,18 @@ const (
...
-	//
-	//nolint:gosec // not actual credentials
+	// not actual credentials.
 	RegisterTokenResp = `{
 		"token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJ3b3JrZXIiLCJpYXQiOjE1MTYyMzkwMjIsInRva2VuX3R5cGUiOiJXb3JrZXJSZWdpc3RlciJ9.gEzKaZB-sDd_gFCVF5uGo2mcf3iy9CrXDTLPZ6PTsTc"
 	}`
+
+	// QueueRegistrationResp represents a JSON return for an admin requesting a queue registration info.
+	//
+	// not actual credentials.
```
wass3rw3rk, line 80: note: surprised linter didn't bark here - if it does, add `//nolint:gosec // not actual credentials`

queue/redis/opts.go
```diff
@@ -132,6 +136,10 @@ func WithPublicKey(key string) ClientOpt {
...
+		}
+
 		if c.config.PublicKey == nil {
 			return errors.New("unable to copy decoded queue signing public key, copied key is nil")
```
wass3rw3rk, line 144:
```suggestion
			return errors.New("unable to copy decoded queue public key, copied key is nil")
```
to match with previous suggested change. otherwise it reads like public key is used for queue signing.

queue/setup.go
```diff
@@ -92,6 +92,10 @@ func (s *Setup) Validate() error {
...
+		return fmt.Errorf("no public key was provided")
```
wass3rw3rk, line 96:
```suggestion
		return fmt.Errorf("no queue public key was provided")
```

wass3rw3rk (changes requested):

mock/server/worker.go
timhuynh94: odd that my linter was yelling at me for having `/nolint:gosec // not actual credentials` here. But i'll them back in

mock/server/worker.go
```diff
@@ -75,6 +75,14 @@ const (
 	RegisterTokenResp = `{
 		"token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJ3b3JrZXIiLCJpYXQiOjE1MTYyMzkwMjIsInRva2VuX3R5cGUiOiJXb3JrZXJSZWdpc3RlciJ9.gEzKaZB-sDd_gFCVF5uGo2mcf3iy9CrXDTLPZ6PTsTc"
 	}`
+
+	// QueueRegistrationResp represents a JSON return for an admin requesting a queue registration info.
+	//
+	//nolint:gosec // not actual credentials
```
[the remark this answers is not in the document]
wass3rw3rk: yea, that was just a note by me saying that if the linter yells, we might need that line. to be honest, i kind of expected that we would :tableflip:

wass3rw3rk (changes requested):
Thanks for the fixes ❤️ 

We should add the public key to the `environment` section in the `docker-compose.yml` in this vicinity: https://github.com/go-vela/server/blob/v21_registration/docker-compose.yml#L31. otherwise, you will get validation error with `make up`. we will need to do this in all repos with a `docker-compose.yml`.

Also, once the `worker` PR is done for this effort, we will want to remove https://github.com/go-vela/server/blob/388e91e332261851603fb0812f4433f5ddc151a9/docker-compose.yml#L72-L73 everywhere as well.

api/queue/queue.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	"github.com/go-vela/server/router/middleware/claims"
+	"github.com/go-vela/types/library"
+	"github.com/sirupsen/logrus"
+)
+
+// swagger:operation POST /api/v1/queue/register queue Registration
```
plyr4, line 16: nice, thanks for the rename. 

```suggestion
// swagger:operation POST /api/v1/queue/register queue Register
```

wass3rw3rk (approved):

plyr4 (approved):

ecrupper (approved):

docker-compose.yml
```diff
@@ -55,36 +61,43 @@ services:
...
   # queue and executing them based off their configuration.
   #
   # https://go-vela.github.io/docs/administration/worker/
   worker:
+# comment the 5 lines below to use local server built image
+#    build:
+#      context: ../worker
+#      dockerfile: Dockerfile
+#    container_name: worker
+#    image: worker:local
```
wass3r, line 77: i think this is leftover from local testing, i say we remove it

wass3r (changes requested):

docker-compose.yml
timhuynh94: Should we leave it in here so that folks can uncomment and run locally?

wass3r: there a similar spot in server section too

wass3r: I'm not a fan because it's beyond the scope of the PR. It also makes some assumptions about local folder structure.  If we want to enhance the file for local dev, I'd do it in separate PR in all the repos that need it.

timhuynh94: Alright! I'll keep that to a different PR for now to keep this effort going!

wass3r (dismissed):

docker-compose.yml
```diff
@@ -69,22 +73,22 @@ services:
...
       WORKER_ADDR: 'http://worker:8080'
-      WORKER_CHECK_IN: 5m
+      WORKER_CHECK_IN: 2m
     restart: always
     ports:
-      - '8081:8080'
+      - "8081:8080"
     volumes:
-      - '/var/run/docker.sock:/var/run/docker.sock'
+      - "/var/run/docker.sock:/var/run/docker.sock"
```
plyr4, line 91: should we revert these? the other variables use single quotes

cognifloyd (approved):

plyr4 (approved):

PR #961, opened 2023-09-07
enhance(api/deploy): use default branch if ref is not supplied

ecrupper:
Taking the injected default duty away from CLI [here](https://github.com/go-vela/cli/blob/010e3166b256e365315ceb91e30bb70403a458f5/command/deployment/add.go#L47). Will be a follow up PR after this.


wass3rw3rk (approved):
❤️

timhuynh94 (approved):

PR #962, opened 2023-09-07
patch(sync): address sync bug with update webhook call

ecrupper:
Preparing for patch `v0.20.2`. This will include these changes: https://github.com/go-vela/server/pull/953


wass3rw3rk (approved):

plyr4 (approved):
nice

PR #968, opened 2023-09-18
fix(schedule): use repo default branch when schedule branch is empty

ecrupper:
This should only be an issue for schedules created prior to `v0.21.0`, as they will have an empty value for `branch`. Will spare the users the headache of having to update the schedule for it to run properly.

I could have made this part of the migration script I suppose, so if anyone thinks that should be the preferred route, feel free to comment 👍


ecrupper:
Closing in favor of using a migration script for this and not introducing unnecessary code

PR #971, opened 2023-09-22
enhance(logging): add flag to allow logging in ecs format, map custom fields to ecs fields

KellyMerrick:
Add flag to allow logging in [Elasticsearch Common Schema (ECS)](https://www.elastic.co/guide/en/ecs-logging/go-logrus/current/intro.html) format.  Default will remain `logrus.JSONFormatter`.

Custom Vela fields will be nested under `labels.vela`, which allows separation from other non-Vela custom labels that may exist in one's logging stack.

https://github.com/go-vela/server/blob/a000fd02ba65eaae3c52283667a5c1ea1bc39b07/cmd/vela-server/main.go#L51-L56

https://github.com/go-vela/server/blob/a000fd02ba65eaae3c52283667a5c1ea1bc39b07/cmd/vela-server/server.go#L29-L39

As noted below, a custom formatter, borrowing from ecslogrus, was created to handle Vela custom fields.

https://github.com/go-vela/server/blob/a000fd02ba65eaae3c52283667a5c1ea1bc39b07/router/middleware/logger.go#L23-L25

https://github.com/go-vela/server/blob/a000fd02ba65eaae3c52283667a5c1ea1bc39b07/router/middleware/logger.go#L145-L192

default log:

```
{"build":"a build","ip":"123.4.5.6","latency":2303458,"level":"info","method":"GET","msg":"200 2.303458ms 123.4.5.6 GET /some/path","path":"/some/path","status":200,"time":"2023-09-26T10:03:27-05:00","user":"a user","user-agent":"user-agent field","version":"version field"}
```

ecs formatted log:

```
{"@timestamp":"2023-09-26T10:04:07.218-0500","client.ip":"123.4.5.6","ecs.version":"1.6.0","event.duration":2928917,"http.request.method":"GET","http.response.status_code":200,"labels.vela":{"build":"a build","user":"a user"},"log.level":"info","message":"200 2.928917ms 123.4.5.6 GET /some/path","url.path":"/some/path","user_agent.name":"user-agent field","user_agent.version":"version field"}
```


router/middleware/logger_test.go
```diff
@@ -217,3 +224,61 @@ func TestMiddleware_Logger_Sanitize(t *testing.T) {
...
+		t.Errorf("Logger returned %v, want %v", gotLabels, wantLabels)
+	}
+
+	if !reflect.DeepEqual(wantMethod, gotMethod) {
+		t.Errorf("Logger returned %v, want %v", gotMethod, wantMethod)
+	}
+
+	if !reflect.DeepEqual(wantMethod, gotMethod) {
+		t.Errorf("Logger returned %v, want %v", gotMethod, wantMethod)
+	}
```
plyr4, line 278: should one of these duplicates be removed?

router/middleware/logger_test.go
```diff
@@ -217,3 +224,61 @@ func TestMiddleware_Logger_Sanitize(t *testing.T) {
...
+	gotLabels := string(formatter.DataKey)
+	gotMethod := entry.Data["method"]
+	gotOrg := entry.Data["org"]
+
+	if err != nil {
+		fmt.Println("err:", err.Error())
+	}
+
+	if got == nil {
+		t.Errorf("Logger returned nothing, want a log")
```
plyr4, line 265: func name typo
```suggestion
		t.Errorf("Format returned nothing, want a log")
```

router/middleware/logger_test.go
```diff
@@ -217,3 +224,61 @@ func TestMiddleware_Logger_Sanitize(t *testing.T) {
...
+		"version":    "v1.0.0",
+		"org":        "foo",
+	}
+
+	logger := logrus.NewEntry(logrus.StandardLogger())
+	entry := logger.WithFields(fields)
+
+	got, err := formatter.Format(entry)
+
+	wantLabels := "labels.vela"
```
plyr4, line 251: not necessary, but you could clean this up by moving this variable to the top and passing it into the `formatter`'s `DataKey`

router/middleware/logger_test.go
```diff
@@ -217,3 +224,61 @@ func TestMiddleware_Logger_Sanitize(t *testing.T) {
...
+	wantMethod := http.MethodGet
+	wantOrg := "foo"
+
+	// run test
+	gotLabels := string(formatter.DataKey)
+	gotMethod := entry.Data["method"]
+	gotOrg := entry.Data["org"]
+
+	if err != nil {
+		fmt.Println("err:", err.Error())
```
plyr4, line 261: should we propagate the error up, rather than just print?
```suggestion
		t.Errorf("Format returned err: %v", err)
```

router/middleware/logger_test.go
```diff
@@ -217,3 +224,61 @@ func TestMiddleware_Logger_Sanitize(t *testing.T) {
...
+	got, err := formatter.Format(entry)
+
+	wantLabels := "labels.vela"
+	wantMethod := http.MethodGet
+	wantOrg := "foo"
+
+	// run test
+	gotLabels := string(formatter.DataKey)
+	gotMethod := entry.Data["method"]
+	gotOrg := entry.Data["org"]
```
plyr4, line 258: these look like they are checking the "old" field keys? i believe we should be checking the contents of `got` from _after_ the `Format` is called
```
	got, err := formatter.Format(entry)
```

router/middleware/logger_test.go
KellyMerrick: addressed

router/middleware/logger_test.go
KellyMerrick: addressed

router/middleware/logger_test.go
KellyMerrick: addressed

router/middleware/logger_test.go
KellyMerrick: addressed

router/middleware/logger.go
```diff
@@ -20,6 +20,33 @@ import (
...
+	ecsVersion = "1.6.0"
+)
+
+var (
+	ecsFieldMap = logrus.FieldMap{
+		logrus.FieldKeyTime:  "@timestamp",
+		logrus.FieldKeyMsg:   "message",
+		logrus.FieldKeyLevel: "log.level",
+	}
+)
```
wass3rw3rk, line 38: instead of having these global variables, let's move them inside the .Format method. We don't need them anywhere else.

router/middleware/logger.go
```diff
@@ -20,6 +20,33 @@ import (
...
+
+// Formatter holds ECS parameter information for logging.
+type Formatter struct {
```
wass3rw3rk, line 41: I would propose a more specific name. Maybe `ECSFormatter`? If/When we add another formatter we will run into issues otherwise.

router/middleware/logger.go
```diff
@@ -114,3 +141,52 @@ func sanitize(body interface{}) interface{} {
 
 	return body
 }
+
+// Format formats e as ECS-compliant JSON,
```
wass3rw3rk, line 145:
```suggestion
// Format formats logrus.Entry as ECS-compliant JSON,
```

wass3rw3rk (changes requested):

docker-compose.yml
```diff
@@ -37,6 +37,7 @@ services:
       VELA_ADDR: 'http://localhost:8080'
       VELA_WEBUI_ADDR: 'http://localhost:8888'
       VELA_LOG_LEVEL: trace
+      VELA_LOG_FORMATTER: json
```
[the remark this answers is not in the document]
KellyMerrick: addressed

router/middleware/logger.go
KellyMerrick: addressed

router/middleware/logger.go
KellyMerrick: addressed. i actually initially had it this way, but changed it. lesson learned, go with your gut.

wass3rw3rk (approved):

timhuynh94 (approved):

PR #972, opened 2023-09-26
feat: endpoint for validating oauth token source

plyr4:
the following PR refactors oauth token validation into its own function and exposes the functionality through an API endpoint. 

the purpose is to allow users to pass back tokens embedded by vela pipelines to validate that they were indeed created by Vela. it is meant to also improve the security around third-party integrated tooling.

- the `AuthenticateToken` has been refactored to use the new func, but it is functionally identical.
- also fixes a couple test typos

closes https://github.com/go-vela/community/issues/854


wass3rw3rk (approved):

timhuynh94 (approved):

api/auth/validate_oauth.go
```diff
@@ -0,0 +1,79 @@
...
+//     schema:
+//       "$ref": "#/definitions/Token"
+//   '401':
+//     description: Unable to validate
+//     schema:
+//       "$ref": "#/definitions/Error"
+//   '503':
+//     description: Service unavailable
+//     schema:
+//       "$ref": "#/definitions/Error"
```
wass3rw3rk, line 42: i don't think we return 503 on this endpoint?

plyr4: oh good point, copy pasta. ill remove

wass3rw3rk (approved):

timhuynh94 (approved):

PR #983, opened 2023-10-06
enhance(compiler): environment and template adjustments for local compilation

ecrupper:
This is the first of three changes that will enhance the `vela exec pipeline` command. These changes fall under the `Server` section below.

### The Issue

`vela exec pipeline` does not work with templates. It also does not work with custom environments if the user is trying to emulate any of the default environment variables, such as `VELA_BUILD_COMMIT`.

### Server Changes

1. Execute the local environment sourcing _after_ the default env override. This will ensure that user-set values that would be found in the default environment will be properly interpreted by the compiler.
2. Adjustments to `getTemplate` that allows the compiler to read both from local file tree as well as remote using a GitHub token.

### Worker Changes

1. Remove calls to merge environment with default environment found in [executor/local/step.go](https://github.com/go-vela/worker/blob/be492bd1e78717a54717c183983845be3234c236/executor/local/step.go#L64-L70). This call is unnecessary and will only overwrite user-supplied environment values found in the default env.

### CLI Changes

1. Support for `--template`, `--template-file`, `--compiler.github.token`, and `--compiler.github.url` flags to the `exec pipeline` command. This will allow users the option to leverage templates in local pipeline execution, both local and remote.
2. Support for `--env-file` and `--env-file-path` to allow users to source environment variables from a file.

### DIFF
--- a/compiler/native/environment.go
+++ b/compiler/native/environment.go
@@ -90,18 +90,6 @@ func (c *client) EnvironmentStep(s *yaml.Step, stageEnv raw.StringSliceMap) (*ya
 	// gather set of default environment variables
 	defaultEnv := environment(c.build, c.metadata, c.repo, c.user)
 
-	// check if the compiler is setup for a local pipeline
-	// and the step isn't setup to run in a detached state
-	if c.local && !s.Detach {
-		// capture all environment variables from the local environment
-		for _, e := range os.Environ() {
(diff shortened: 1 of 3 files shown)


plyr4 (approved):
makes sense, nice

wass3rw3rk (approved):

PR #989, opened 2023-10-17
feat(api/user): add endpoint for viewing builds sent by current user

ecrupper:
This would allow for the creation of a "Your Builds" page in the UI, which I think would be really convenient for our users.

Pretty slim related issue: https://github.com/go-vela/community/issues/551

This query would be very taxing without introducing an index on `sender` for the builds table, which is the approach I took here.

### DIFF
--- a/database/build/interface.go
+++ b/database/build/interface.go
@@ -36,6 +36,8 @@ type BuildInterface interface {
 	CountBuildsForOrg(context.Context, string, map[string]interface{}) (int64, error)
 	// CountBuildsForRepo defines a function that gets the count of builds by repo ID.
 	CountBuildsForRepo(context.Context, *library.Repo, map[string]interface{}) (int64, error)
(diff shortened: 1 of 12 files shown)


database/build/interface.go
```diff
@@ -36,6 +36,8 @@ type BuildInterface interface {
 	CountBuildsForOrg(context.Context, string, map[string]interface{}) (int64, error)
 	// CountBuildsForRepo defines a function that gets the count of builds by repo ID.
 	CountBuildsForRepo(context.Context, *library.Repo, map[string]interface{}) (int64, error)
+	// CountBuildsForSender defines a function that gets the count  of builds by sender.
```
cognifloyd, line 39:
```suggestion
	// CountBuildsForSender defines a function that gets the count of builds by sender.
```

cognifloyd (approved):

ecrupper:
Closing this. I think we'll prefer dashboards over a page like this for the time being. May revisit at some point.

PR #991, opened 2023-10-23
feat(queue): add ping function for redis

timhuynh94:
Closes [#859](https://github.com/go-vela/community/issues/859)


plyr4 (approved):

ecrupper:
I believe we [ping the queue](https://github.com/go-vela/server/blob/main/queue/redis/redis.go#L144-L166) on start up when creating a new Redis connection. I imagine any logic based on the success/failure of pinging the queue could be branched off of whether or not the `New(...)` method was successful?

queue/redis/ping_test.go
```diff
@@ -0,0 +1,61 @@
...
+		WithAddress(fmt.Sprintf("redis://%s", _redis.Addr())),
+		WithChannels("foo"),
+		WithCluster(false),
+		WithTimeout(5*time.Second),
+	)
+	_redis.SetError("not aiv")
+	// run tests
+	err = badRedis.Ping(context.Background())
+	if err == nil {
+		t.Errorf("Ping shoudl have returned err: %v", err)
```
cognifloyd, line 59:
```suggestion
		t.Errorf("Ping should have returned err: %v", err)
```

cognifloyd (approved):

timhuynh94:
> I believe we [ping the queue](https://github.com/go-vela/server/blob/main/queue/redis/redis.go#L144-L166) on start up when creating a new Redis connection. I imagine any logic based on the success/failure of pinging the queue could be branched off of whether or not the `New(...)` method was successful?

The issue with `New(...)` method is that it only gets pinged once during setup and never pinged again unless `New(...)` method is called again. We could also entirely with the `New(...)` method logic by creating new queue client everytime the worker checks in with server.

plyr4 (approved):

PR #993, opened 2023-10-26
enhance(api/webhook): update hook status to skipped when build is skipped

ecrupper:
Oftentimes, users will have their pipeline steps purged completely, and the hook's successful status in addition to the lack of any messaging leads to confusion. End result of these changes would be something like this: 

Screenshot 2023-10-26 at 1 11 04 PM

The above provides a clear path forward for users.


claire1618 (dismissed):

timhuynh94 (dismissed):

timhuynh94:
Looks like test failed!

timhuynh94 (dismissed):

api/build/skip.go
```diff
@@ -13,25 +13,25 @@ import (
 func SkipEmptyBuild(p *pipeline.Build) string {
 	if len(p.Stages) == 1 {
 		if p.Stages[0].Name == "init" {
-			return "skipping build since only init stage found"
+			return "skipping build since only init stage found. It is likely no rulesets matched for the webhook payload."
```
wass3r, line 16: Super minor thing: I think we follow Go convention here (even though it never ends up in a proper `error`) for lower casing error message and no periods. So it makes it a little awkward when you add a proper sentence and punctuation. Should we upper case the beginning or lower case everything and pick something like `-` to separate?

wass3r (approved):

claire1618 (approved):

PR #998, opened 2023-11-02
fix(schedule): honor allow list for previously created schedules

ecrupper:
This change makes sure that any repo taken off of the allowlist does not continue to trigger existing schedules


claire1618 (approved):

wass3rw3rk (approved):
nice. the arity of that function is getting a little out of hand, probably time for a struct 🫣

PR #1011, opened 2023-11-22
fix(repo): repair update name

timhuynh94:
This change is to fix the Repair button under Settings. Previously, when clicking Repair button on an unarchived renamed repo and push to renamed repo, builds will not be triggered based on webhook signatures failures.
Closes [#862](https://github.com/go-vela/community/issues/862)


claire1618 (dismissed):

plyr4 (approved):

api/repo/repair.go
```diff
@@ -116,6 +116,34 @@ func RepairRepo(c *gin.Context) {
 		}
 	}
 
+	// get repo information from the source
+	sourceRepo, err := scm.FromContext(c).GetRepo(ctx, u, r)
```
wass3rw3rk, line 120: just to clarify here.. this will follow GH's redirect for a renamed repo and return the new info? just asking since you are passing in current repo to fetch?

wass3rw3rk (approved):

api/repo/repair.go
timhuynh94: Yes that is correct! It will follow GH's redirect for a renamed repo and uses the new info to update the current one in the DB

api/repo/repair.go
```diff
@@ -116,6 +116,35 @@ func RepairRepo(c *gin.Context) {
...
+	if err != nil {
+		retErr := fmt.Errorf("unable to retrieve repo info for %s from source: %w", sourceRepo.GetFullName(), err)
+
+		util.HandleError(c, http.StatusBadRequest, retErr)
+
+		return
+	}
+	// if repo has a name change, then update DB with new name
+	// if repo has a org change, update org as well
+	if sourceRepo.GetName() != r.GetName() || sourceRepo.GetOrg() != r.GetOrg() {
```
ecrupper, line 130: Discussed in a side conversation but noting it here: we'll want to handle situations like this similarly to how we [currently handle](https://github.com/go-vela/server/blob/3d67002de8f87154fd52e79e562a90f1aacb6894/api/webhook/post.go#L814) `renamed` / `transferred` repository events, specifically adjusting the repository's resources (builds and secrets) to accommodate the name change.

ecrupper (changes requested):
Looks like a great way to further utilize the repair API. I noted a few extra steps that will be necessary for this to work as intended. Out of scope for this PR, but I wonder if the `sync` endpoint could also do something similar in these odd cases.

api/repo/repair.go
```diff
@@ -116,6 +116,35 @@ func RepairRepo(c *gin.Context) {
...
+		r.SetPreviousName(r.GetName())
```
wass3rw3rk, line 131: not sure that's it's crucial, but do we only want to do this if repo name specifically changed? as it stands, it will overwrite previous name if only org changed?

api/repo/repair.go
wass3rw3rk: good call - is this being implemented as part of this PR or a follow-up PR?

api/repo/repair.go
```diff
@@ -116,6 +116,35 @@ func RepairRepo(c *gin.Context) {
...
+		r.SetName(sourceRepo.GetName())
+		r.SetFullName(sourceRepo.GetFullName())
+		r.SetLink(sourceRepo.GetLink())
+		r.SetClone(sourceRepo.GetClone())
+		r.SetOrg(sourceRepo.GetOrg())
+		// send API call to update the repo
+		_, err := database.FromContext(c).UpdateRepo(ctx, r)
+		if err != nil {
+			retErr := fmt.Errorf("unable to rename repo %s to %s: %w", sourceRepo.GetFullName(), r.GetFullName(), err)
```
wass3rw3rk, line 140: wouldn't this error message always show the same full name? because of `r.SetFullName(sourceRepo.GetFullName())` a few lines above?

wass3rw3rk (changes requested):

api/repo/repair.go
wass3rw3rk: good call - is this being implemented as part of this PR or a follow-up PR?

api/repo/repair.go
```diff
@@ -116,6 +119,41 @@ func RepairRepo(c *gin.Context) {
...
+	if err != nil {
+		retErr := fmt.Errorf("unable to retrieve repo info for %s from source: %w", sourceRepo.GetFullName(), err)
+
+		util.HandleError(c, http.StatusBadRequest, retErr)
+
+		return
+	}
+
+	// if repo has a name change, then update DB with new name
+	// if repo has a org change, update org as well
```
KellyMerrick, line 133:
```suggestion
	// if repo has an org change, update org as well
```

KellyMerrick (commented):
Super minor request :disappear:

wass3rw3rk (approved):
i would change the module import order in api/repo/repair.go to be closer to what it was before, but we don't currently enforce this and it's not consistent across repos either. otherwise, lgtm.

PR #1012, opened 2023-11-24
fix(auto_cancel): support canceling pull_request:opened and abstract determination logic 

ecrupper:
Closes https://github.com/go-vela/community/issues/884

I also took this moment as an opportunity to abstract the logic of determining when and what to auto-cancel:

- `isCancelable` will determine whether or not a pending/running build (target) should be canceled by the superseding build (current)
- `ShouldAutoCancel` will determine whether the build being processed is eligible to auto-cancel builds

Lots of ifs, ands, and switches, so feel free to chime in on if you think any particular bit of code needs further elaboration.

Abstraction did make testing easier, hence the new tests 😄


plyr4 (approved):
nice

KellyMerrick (approved):
lgtm

PR #1015, opened 2023-11-28
enhance(api/scm)!: return affected repos in API response and do not assume 404 on GetRepo error

ecrupper:
Particularly helpful when syncing repositories is knowing if anything actually changed with the database object. This change has the API response return any affected repo in its response. Otherwise, if nothing changed, it will return a `204 - No Content`. 

Also fixed an issue where any error resulting from the SCM call to `GetRepo` was interpreted as a `404` error. That is not always true.


wass3rw3rk (commented):
does that mean the repo sync endpoint could return 204 as well? do we need to lift the response codes the github api can return to our api spec comments in this section: https://github.com/go-vela/server/pull/1015/files#diff-ccd33b844d96fe96ea644179c0ac423c805dc392a9e24117a60d2b269721eb9aL40

timhuynh94 (approved):

claire1618 (approved):

PR #1016, opened 2023-11-28
feat(webhook)!: support build approval based on repository settings

ecrupper:
This is the server-side implementation for approving builds, specifically from fork repos during PR builds.

A couple small refactors that made this possible:

- `PublishToQueue` now takes the route as a parameter rather than determining it. This was done in order to avoid unmarshaling the pipeline object when approved. We're able to do this by preemptively setting the `build.Host` to the specified route rather than waiting for a worker to pick it up. 
- `RepoAccess` accepts a user name rather than an entire user object since that's all we need + adds ability to check for access of `sender` without creating a fake user object and populating the name
- `executors.Establish()` has an additional check on build status since host is now being set as the route while the build is pending / pending approval.

The main idea of the implementation is to separate the publishing of the build executable from the publishing to the queue. If the build is clear to run, the `PostWebhook` handler will take care of both. However, if it is not clear to run, then the new `ApproveBuild` handler will handle queue publishing while the `PostWebhook` handler just generates and publishes the executable.

Lastly, the status report to the SCM for builds pending approval has been adjusted to describe the situation, making the experience more understandable for users having their builds gated.


KellyMerrick (dismissed):
lgtm

KellyMerrick (approved):
lgtm

timhuynh94 (approved):

PR #1018, opened 2023-12-01
test(integration): refactor to run db integration test as matrix

wass3r:
closes https://github.com/go-vela/server/pull/967

split postgres and sqlite testing so that we can matrix test postgres versions to adhere to n-1 compatibility.

note: not sure what renovate will do here, so might have to adjustments once we find out. potentially a little more overhead if renovate doesn't update for us anymore, though maybe there's a way to configure it to still handle this scenario.


KellyMerrick (approved):
lgtm

PR #1019, opened 2023-12-01
chore: minor tweaks for local dev

wass3r:
couple of minor tweaks and fixes for annoyances i intend to carry over to the other core repos if accepted:

1. after running `make up` there's a 95%+ chance i will attempt to access the stack before it is ready (mainly due to postgres service taking a while on my machine). this is fixed by adding a health check for postgres and making sure the server doesn't start until it's healthy. could maybe tweak interval to check more frequently. lmk.
1. changes `docker-compose` -> `docker compose`. ref: https://www.docker.com/blog/announcing-compose-v2-general-availability/ . i don't want to alias something that is deprecated, so i typically apply this edit locally before i run `make ...`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -227,7 +227,7 @@ bump-deps-full: check
 pull:
 	@echo
 	@echo "### Pulling images for docker-compose stack"
-	@docker-compose pull
+	@docker compose pull
 
 # The `compose-up` target is intended to build and create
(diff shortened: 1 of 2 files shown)


ecrupper (approved):

KellyMerrick (approved):
lgtm

PR #1023, opened 2023-12-11
feat(repo)!: support allow_events struct

ecrupper:
Keeping all the legacy fields (`AllowPush`, `AllowPull`, `AllowTag`, `AllowDeploy`, `AllowComment`) in order to preserve data for now with the intent of dropping the columns in a later release.

Otherwise, all logic in the core path has been migrated to use the new `AllowEvents` struct in this PR. I also threw in the optional `VELA_DEFAULT_REPO_EVENTS_MASK`, which allows admins to further customize those default events if they so choose.


api/webhook/post.go
```diff
@@ -256,15 +256,16 @@ func PostWebhook(c *gin.Context) {
 	}
 
 	// verify the build has a valid event and the repo allows that event type
-	if (b.GetEvent() == constants.EventPush && !repo.GetAllowPush()) ||
-		(b.GetEvent() == constants.EventPull && !repo.GetAllowPull()) ||
-		(b.GetEvent() == constants.EventComment && !repo.GetAllowComment()) ||
-		(b.GetEvent() == constants.EventTag && !repo.GetAllowTag()) ||
-		(b.GetEvent() == constants.EventDeploy && !repo.GetAllowDeploy()) {
-		retErr := fmt.Errorf("%s: %s does not have %s events enabled", baseErr, repo.GetFullName(), b.GetEvent())
+	if !repo.EventAllowed(b.GetEvent(), b.GetEventAction()) {
```
wass3rw3rk, line 259: much nicer :D

api/repo/update.go
```diff
@@ -170,6 +173,14 @@ func UpdateRepo(c *gin.Context) {
 		r.SetActive(input.GetActive())
 	}
 
+	// set allow events based on input if given
+	if !reflect.DeepEqual(input.GetAllowEvents(), new(library.Events)) {
+		r.SetAllowEvents(input.GetAllowEvents())
```
wass3rw3rk, line 178: wouldn't we want to compare against r.GetAllowEvents?

ecrupper: Hm, I think honestly we just want to compare

```go
if input.AllowEvents != nil
```

Because this line is really just seeing if they supplied something. I don't like using reflect either for something like that

wass3rw3rk: sure, that works. i also cringed a little at seeing reflect here :D

api/repo/update.go
```diff
@@ -212,6 +222,12 @@ func UpdateRepo(c *gin.Context) {
 		r.SetAllowPull(true)
 		r.SetAllowPush(true)
 	}
+	// -- END DEPRECATED SECTION --
+
+	// set default events if no events are enabled
+	if r.GetAllowEvents().ToDatabase() == 0 {
```
wass3rw3rk, line 228: nit: i know this is in types, but i kind of wish `ToDatabase` was something like `ToBitmask` or something that signals why it's returning an int here.

wass3rw3rk (approved):

claire1618 (approved):

PR #1028, opened 2023-12-19
feat(dashboards)!: server-side implementation of dashboards

ecrupper:
Executing on a concept where users can create pages wherein they can look at a collection of repositories at a high level. Below is a simple mock up of the eventual UI:

dashboard_mock

This back-end struct design accounts for a few key features: share-ability, admin access, and customizability.

**Share-ability**: using a uniquely generated UUID that will serve as the `link` to the dashboard along with user-specific `user.Dashboards` we will be able to allow open navigation to any dashboard via a link + easy navigation through the UI via "My Dashboards", much like many group-board applications such as Draw IO, Miro, etc.

**Admin Access**: since the dashboards will be public and discoverable, it's important to limit update/delete access. This can be done via the `Dashboard.Admins` field, which holds a slice of user_ids capable of editing the dashboard. There is also auditing in place in the form of `created_at`, `created_by`, `updated_at`, `updated_by`.

**Customizability**: choosing to store the dashboard repo collection as a JSON field allows users to customize their "subscription" to that repo, using the `Events` and `Branches` field. While it would have been possible to reference repositories in a slice field of repo_ids, that way makes customization to this degree much more challenging.


api/dashboard/update.go
```diff
@@ -0,0 +1,156 @@
...
+	// determine if claims user is in the admin set
+	for _, a := range admins {
+		if strings.EqualFold(fmt.Sprintf("%d", u), a) {
+			admin = true
+			break
+		}
+	}
+
+	return admin
+}
```
wass3rw3rk, line 156: should we use `slices.Contains`?

database/dashboard/create.go
```diff
@@ -0,0 +1,37 @@
...
+
+// CreateUser creates a new user in the database.
+func (e *engine) CreateDashboard(ctx context.Context, d *api.Dashboard) (*api.Dashboard, error) {
+	e.logger.WithFields(logrus.Fields{
+		"dashboard": d.GetName(),
+	}).Tracef("creating dashboard %s in the database", d.GetName())
+
+	// cast the library type to database type
+	//
+	// https://pkg.go.dev/github.com/go-vela/types/database#UserFromLibrary
```
wass3rw3rk, line 22: wrong link

database/dashboard/create.go
```diff
@@ -0,0 +1,37 @@
...
+	dashboard := FromAPI(d)
+
+	// validate the necessary fields are populated
+	//
+	// https://pkg.go.dev/github.com/go-vela/types/database#User.Validate
```
wass3rw3rk, line 27: wrong link

database/dashboard/create.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"context"
+
+	"github.com/sirupsen/logrus"
+
+	api "github.com/go-vela/server/api/types"
+	"github.com/go-vela/server/constants"
+)
+
+// CreateUser creates a new user in the database.
```
wass3rw3rk, line 14: wrong comment

wass3rw3rk (changes requested):

database/build/list_dashboard.go
```diff
@@ -0,0 +1,60 @@
...
+	api "github.com/go-vela/server/api/types"
+	"github.com/go-vela/types/constants"
+	"github.com/go-vela/types/database"
+	"github.com/go-vela/types/library"
+)
+
+// ListBuildsForDashboardRepo gets a list of builds by repo ID from the database.
+//
+//nolint:lll // ignore long line length due to variable names
+func (e *engine) ListBuildsForDashboardRepo(ctx context.Context, r *api.Repo, branches, events []string) ([]*library.Build, error) {
```
plyr4, line 19: is there an advantage to having this version of ListBuildsForRepo?

api/dashboard/create.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	c.JSON(http.StatusCreated, d)
+}
+
+// validateAdminSet takes a slice of user names and converts it into a slice of matching
+// user ids in order to preserve data integrity in case of name change.
+func validateAdminSet(c context.Context, caller *types.User, users []string) ([]string, error) {
+	// add user creating the dashboard to admin list
+	admins := []string{fmt.Sprintf("%d", caller.GetID())}
```
plyr4, line 146: should we move this create-specific logic to outside of this func?

api/dashboard/create.go
```diff
@@ -0,0 +1,186 @@
...
+		if admin == caller.GetName() {
+			continue
+		}
+
+		u, err := database.FromContext(c).GetUserForName(c, admin)
+		if err != nil || !u.GetActive() {
+			return nil, fmt.Errorf("unable to create dashboard: %s is not an active user", admin)
+		}
+
+		admins = append(admins, fmt.Sprintf("%d", u.GetID()))
```
plyr4, line 159: wondering if we should check for duplicate user id's somewhere in here

database/build/list_dashboard.go
ecrupper: yeah, the ListBuildsForRepo uses the [map filter](https://github.com/go-vela/server/blob/04f930869c93b4a6bef0f1233f9f59d081ec0b8b/database/build/list_repo.go#L49) Gorm function, which to my knowledge doesn't cover the `IN` keyword, but I could be wrong. Gorm docs are not the most intuitive. Second, I didn't want to have the created constraints baked in... as well as the pagination

wass3rw3rk: tangential: `lll` has been [turned off](https://github.com/go-vela/server/blob/04f930869c93b4a6bef0f1233f9f59d081ec0b8b/.golangci.yml#L137), so you shouldn't need that `//nolint` directive

api/dashboard/list_user.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/go-vela/server/api/types"
+	"github.com/go-vela/server/database"
+	"github.com/go-vela/server/router/middleware/user"
+	"github.com/go-vela/server/util"
+)
+
+// swagger:operation GET /api/v1/user/dashboards dashboards ListUserDashboards
+//
+// Get all dashboards for the claims users in the configured backend
```
wass3rw3rk, line 22: nit: 
```suggestion
// Get all dashboards for the current user in the configured backend
```

wass3rw3rk: i know we use "claims user" elsewhere so i would propose updating there too. this reads better, imo.

database/dashboard/dashboard.go
```diff
@@ -0,0 +1,249 @@
...
+		UpdatedAt sql.NullInt64  `sql:"updated_at"`
+		UpdatedBy sql.NullString `sql:"updated_by"`
+		Admins    pq.StringArray `sql:"admins" gorm:"type:varchar(5000)"`
+		Repos     DashReposJSON
+	}
+
+	DashReposJSON []*api.DashboardRepo
+)
+
+// New creates and returns a Vela service for integrating with build executables in the database.
```
wass3rw3rk, line 75:
```suggestion
// New creates and returns a Vela service for integrating with dashboards in the database.
```

database/dashboard/dashboard.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	// engine represents the build executable functionality that implements the BuildExecutableService interface.
+	engine struct {
+		// engine configuration settings used in build executable functions
+		config *config
+
+		ctx context.Context
+
+		// gorm.io/gorm database client used in build executable functions
```
wass3rw3rk, line 49:
```suggestion
		// gorm.io/gorm database client used in dashboard functions
```

database/dashboard/dashboard.go
```diff
@@ -0,0 +1,249 @@
...
+		//
+		// https://pkg.go.dev/gorm.io/gorm#DB
+		client *gorm.DB
+
+		// sirupsen/logrus logger used in build executable functions
```
wass3rw3rk, line 54:
```suggestion
		// sirupsen/logrus logger used in dashboard functions
```

database/dashboard/dashboard.go
```diff
@@ -0,0 +1,249 @@
...
+	config struct {
+		// specifies to skip creating tables and indexes for the BuildExecutable engine
+		SkipCreation bool
+		// specifies the driver for proper popping query
+		Driver string
+	}
+
+	// engine represents the build executable functionality that implements the BuildExecutableService interface.
+	engine struct {
+		// engine configuration settings used in build executable functions
```
wass3rw3rk, line 44:
```suggestion
		// engine configuration settings used in dashboard functions
```

database/dashboard/dashboard.go
```diff
@@ -0,0 +1,249 @@
...
+	// User type has an empty Name field provided.
+	ErrEmptyDashName = errors.New("empty dashboard name provided")
+
+	// ErrExceededAdminLimit defines the error type when a
+	// User type has Admins field provided that exceeds the database limit.
+	ErrExceededAdminLimit = errors.New("exceeded admins limit")
+)
+
+type (
+	// config represents the settings required to create the engine that implements the BuildExecutableService interface.
```
wass3rw3rk, line 34:
```suggestion
	// config represents the settings required to create the engine that implements the DashboardService interface.
```

database/dashboard/dashboard.go
```diff
@@ -0,0 +1,249 @@
...
+	config struct {
+		// specifies to skip creating tables and indexes for the BuildExecutable engine
```
wass3rw3rk, line 36:
```suggestion
		// specifies to skip creating tables and indexes for the Dashboard engine
```

database/dashboard/dashboard.go
```diff
@@ -0,0 +1,249 @@
...
+		SkipCreation bool
+		// specifies the driver for proper popping query
+		Driver string
+	}
+
+	// engine represents the build executable functionality that implements the BuildExecutableService interface.
```
wass3rw3rk, line 42:
```suggestion
	// engine represents the build executable functionality that implements the DashboardService interface.
```

database/dashboard/dashboard.go
```diff
@@ -0,0 +1,249 @@
...
+		// https://pkg.go.dev/gorm.io/gorm#DB
+		client *gorm.DB
+
+		// sirupsen/logrus logger used in build executable functions
+		//
+		// https://pkg.go.dev/github.com/sirupsen/logrus#Entry
+		logger *logrus.Entry
+	}
+
+	// Dashboard is the database representation of a user.
```
wass3rw3rk, line 60:
```suggestion
	// Dashboard is the database representation of a dashboard.
```

database/dashboard/dashboard.go
```diff
@@ -0,0 +1,249 @@
...
+
+	// apply all provided configuration options
+	for _, opt := range opts {
+		err := opt(e)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// check if we should skip creating build executable database objects
```
wass3rw3rk, line 95:
```suggestion
	// check if we should skip creating dashboard database objects
```

database/dashboard/dashboard.go
```diff
@@ -0,0 +1,249 @@
...
+	if e.config.SkipCreation {
+		e.logger.Warning("skipping creation of build executables table and indexes in the database")
```
wass3rw3rk, line 97:
```suggestion
		e.logger.Warning("skipping creation of dashboard table and indexes in the database")
```

database/dashboard/dashboard.go
```diff
@@ -0,0 +1,249 @@
...
+
+		return e, nil
+	}
+
+	// create the build executables table
```
wass3rw3rk, line 102:
```suggestion
	// create the dashboard table
```

wass3rw3rk (changes requested):

api/dashboard/create.go
plyr4: just to right before it gets called, wherever that is, its just hard to tell from a glance that validating the list of admins also adds an admin to the list. 
purely an opinion tho

ecrupper: Oh I misunderstood. Yeah I agree.. but I'll just rename the function instead of move the logic outside

database/dashboard/dashboard.go
wass3rw3rk: looks like this didn't get updated?

database/build/list_dashboard.go
plyr4: ah okay, cool.

mock/server/user.go
```diff
@@ -34,15 +35,17 @@ const (
...
   },
   {
     "id": 1,
     "name": "OctoKitty",
     "token": null,
     "favorites": ["github/octocat"],
     "active": true,
-    "admin": false
+    "admin": false,
+	"dashboards": []
```
plyr4, line 48:
```suggestion
"dashboards": []
```

mock/server/user.go
```diff
@@ -23,7 +23,8 @@ const (
   "token": null,
```
plyr4, line 23: are we going to need dashboard mocks?

timhuynh94 (approved):

router/router.go
```diff
@@ -125,7 +125,7 @@ func Load(options ...gin.HandlerFunc) *gin.Engine {
 		ScmHandlers(baseAPI)
 
 		// Search endpoints
-		SearchHandlers(baseAPI)
```
cognifloyd: Was removing this an accident or on purpose?
It looks like `SearchHandlers` is now unused on main. So, if `/api/v1/dashboards` is actually supposed to replace `/api/v1/search/builds`, then that function needs to also be removed. Otherwise, we probably need to re-add the `SearchHandlers` call here.

PR #1029, opened 2023-12-21
enhance(api/workers): add filters to list workers

ecrupper:
I think it would be helpful to add these query parameters for listing workers. I was using `vela get workers` CLI command the other day and trying to do some bash scripting to filter out stale workers and it made me want to do this 😅


timhuynh94 (approved):

claire1618 (dismissed):

database/worker/list_test.go
```diff
@@ -67,20 +79,20 @@ func TestWorker_Engine_ListWorkers(t *testing.T) {
...
-			want:     []*library.Worker{_workerOne, _workerTwo},
+			want:     []*library.Worker{_workerOne, _workerTwo, _workerThree},
 		},
 	}
 
 	// run tests
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			got, err := test.database.ListWorkers(context.TODO())
+			got, err := test.database.ListWorkers(context.TODO(), "all", newer+1, 0)
```
wass3r, line 95: wouldn't you also want a test that will select a subset of workers?

api/worker/list.go
```diff
@@ -20,6 +22,24 @@ import (
...
 // - application/json
+// parameters:
+// - in: query
+//   name: active
+//   description: Filter workers based on active status
+//   type: string
+//   enum:
+//   - true
+//   - false
+//   - all
```
wass3rw3rk, line 33: looks like default is "true", we should add it here

wass3rw3rk (dismissed):

timhuynh94 (approved):

api/worker/list.go
```diff
@@ -20,6 +22,21 @@ import (
 // ---
 // produces:
 // - application/json
+// parameters:
+// - in: query
+//   name: active
+//   description: Filter workers based on active status
+//   type: string
+//   default: "true"
```
wass3r, line 30: any reason you took out the enum? it still only supports three possible values, right?

ecrupper: No it supports anything really... if it doesn't parse as boolean, the filter just isn't applied

wass3r: sure, i mean "valid" values :P

wass3r: ok, what threw me off is the usage of "all" as a value in tests and the type being "string" instead of "boolean" like we do here: https://github.com/go-vela/server/blob/a7cd07fe43cd7e5e0e79f7e1ddb6ca055d518d8a/api/metrics.go#L159-L162.

wass3r: so.. `type: boolean` and take out `default: "true"` - since there is no default? if it's not supplied or a non-boolean-like string is supplied, it just doesn't apply a filter. that, or i need more coffee.

ecrupper: Sounds good. Although it's hard to communicate the `all` / `both` option via the swagger docs

wass3r (approved):

claire1618 (approved):

PR #1037, opened 2024-01-14
fix(api/list-org-repos): ensure active flag is boolean

wass3r:
closes https://github.com/go-vela/community/issues/853

when GORM is given the proper type, it will use the appropriate query depending on which driver is in use.

tested in local setup with both postgres and sqlite. no functional differences.

as a side note, it's not intuitive to start up the local (or any) stack with sqlite. compilation requires CGO and when using alpine requires you to make sure you are on alpine 3.18 instead of the latest (see issue #1164 on github.com/mattn/go-sqlite3). latest gcc and musl in the newest alpine (3.19.x) apparently do not work with that library. alternatively, and probably preferably, it looks like you can pass `CGO_CFLAGS="-D_LARGEFILE64_SOURCE"` in addition to `CGO_ENABLED=1`.

### DIFF
--- a/api/repo/list_org.go
+++ b/api/repo/list_org.go
@@ -122,11 +122,16 @@ func ListReposForOrg(c *gin.Context) {
 	// capture the sort_by query parameter if present
 	sortBy := util.QueryParameter(c, "sort_by", "name")
 
+	// prep filters
+	filters := make(map[string]interface{})
+
 	// capture the query parameters if present:
 	//
 	// * active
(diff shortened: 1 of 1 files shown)


plyr4 (approved):

ecrupper (approved):
Nice. You'll have to share how you got the Sqlite stack running 😄

wass3r:
@ecrupper yes, maybe in form of documentation 😬

PR #1038, opened 2024-01-15
feat: adding a delete event

claire1618:
Similar to the **pull_request** and **push** events, it would be beneficial if Vela could recognize the delete event. So the users would have the option to have Vela set off builds when a branch or tag are deleted in their repo.

The code added here takes the new changes I made in the types repo and takes in a delete webhook event to trigger a build.


api/webhook/post.go
```diff
@@ -367,12 +367,27 @@ func PostWebhook(c *gin.Context) {
...
+		_, commit, err := scm.FromContext(c).GetBranch(ctx, u, r, r.GetBranch())
+		if err != nil {
+			retErr := fmt.Errorf("failed to get commit for repo %s on %s branch: %w", r.GetFullName(), r.GetBranch(), err)
+			util.HandleError(c, http.StatusInternalServerError, retErr)
+		}
+
+		b.SetCommit(commit)
+		b.SetRef(commit)
+	}
+
```
ecrupper, line 383: We don't need this block anymore since we are getting the commit from the `before` value

scm/github/webhook.go
```diff
@@ -68,14 +68,19 @@ func (c *client) ProcessWebhook(ctx context.Context, request *http.Request) (*ty
 	// process the event from the webhook
 	switch event := event.(type) {
 	case *github.PushEvent:
+		c.Logger.Tracef("push")
```
ecrupper, line 71: I think we can remove these trace logs

scm/github/testdata/hooks/delete.json
```diff
@@ -0,0 +1,129 @@
+{
```
ecrupper, line 1: I don't think we need this file, right? We probably need a push payload that is of the delete variety though... with a test 😄

compiler/template/native/testdata/step/loop_map/want.yml
```diff
@@ -5,15 +5,15 @@ steps:
...
+      event: [ push, pull_request, delete ]
 
   - name: sample_test_latest
     commands:
       - go test ./...
     image: golang:latest
     pull: true
     ruleset:
-      event: [ push, pull_request ]   
+      event: [ push, pull_request, delete ]
```
ecrupper, line 16: Is there any reason why all these test pipelines are including `delete`?

ecrupper (changes requested):
A few things... good work

compiler/template/native/testdata/step/loop_map/want.yml
claire1618: im not sure, I didn't add that and I'm a bit unsure as to why all these files are included in my PR

ecrupper (approved):
Nice

PR #1040, opened 2024-01-17
chore: add context to deployment functions

rfigueroa:
see: https://github.com/go-vela/community/issues/832

### DIFF
--- a/api/build/restart.go
+++ b/api/build/restart.go
@@ -350,7 +350,7 @@ func RestartBuild(c *gin.Context) {
 
 		d.SetBuilds(build)
 
-		_, err = database.FromContext(c).UpdateDeployment(d)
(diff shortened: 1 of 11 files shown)


rfigueroa:
@ecrupper https://go-vela.github.io/docs/community/contributing_guidelines/#development-workflow says scope is optional, but look like the the regex expect the scope to be mandatory

ecrupper:
> @ecrupper https://go-vela.github.io/docs/community/contributing_guidelines/#development-workflow says scope is optional, but look like the the regex expect the scope to be mandatory

https://github.com/go-vela/types/pull/347 Yup my bad. Will update all the core repos here now 👍

plyr4 (approved):

PR #1045, opened 2024-01-23
enhance(compiler): allow users to override $HOME in environment

ecrupper:
This came up with some recent versions of `elasticsearch`, where the `$HOME` path was not something we tend to expect. For example: 

```yaml
  - name: assumed home
    image: elasticsearch:8.5.2
    user: elasticsearch
    commands:
      - echo hi
``` 

This pipeline fails with `/bin/sh: line 2: /home/elasticsearch/.netrc: No such file or directory`

This makes sense, as the real value of `$HOME` is different:

```sh
$ docker run -it elasticsearch:8.5.2 sh
sh-5.0$ echo $HOME
/usr/share/elasticsearch
```

However, we assume that the user's `$HOME` directory will be at the path `/home/<user>`. This change allows users to also define the expected `$HOME` for the user. With this change, the below now works:

```yaml
steps:
  - name: proper home
    image: elasticsearch:8.5.2
    user: elasticsearch
    environment:
      HOME: /usr/share/elasticsearch
    commands:
      - echo hello
```

Let me know what you think!


timhuynh94 (dismissed):

compiler/native/script.go
```diff
@@ -98,16 +98,21 @@ func generateScriptPosix(commands []string) string {
...
-machine $VELA_NETRC_MACHINE
-login $VELA_NETRC_USERNAME
-password $VELA_NETRC_PASSWORD
+set +e
+touch $HOME/.netrc 2>/dev/null
+if [ -w "$HOME" ]; then
+  cat <<EOF > $HOME/.netrc
+machine ${VELA_NETRC_MACHINE:-}
+login ${VELA_NETRC_USERNAME:-}
+password ${VELA_NETRC_PASSWORD:-}
```
wass3rw3rk, line 107: why the `:-`?

compiler/native/script.go
```diff
@@ -98,16 +98,21 @@ func generateScriptPosix(commands []string) string {
 // setupScript is a helper script this is added to the build to ensure
 // a minimum set of environment variables are set correctly.
 const setupScript = `
-cat <<EOF > $HOME/.netrc
-machine $VELA_NETRC_MACHINE
-login $VELA_NETRC_USERNAME
-password $VELA_NETRC_PASSWORD
+set +e
+touch $HOME/.netrc 2>/dev/null
```
wass3rw3rk, line 102: do we need this? we are already only writing to it only if we can right after this line.

compiler/native/script.go
```diff
@@ -98,16 +98,21 @@ func generateScriptPosix(commands []string) string {
 // setupScript is a helper script this is added to the build to ensure
 // a minimum set of environment variables are set correctly.
 const setupScript = `
-cat <<EOF > $HOME/.netrc
-machine $VELA_NETRC_MACHINE
-login $VELA_NETRC_USERNAME
-password $VELA_NETRC_PASSWORD
+set +e
```
wass3rw3rk, line 101: "+e" is essentially default behavior, you are unsetting it here - is that the intent?

compiler/native/script.go
```diff
@@ -98,16 +98,21 @@ func generateScriptPosix(commands []string) string {
...
+password ${VELA_NETRC_PASSWORD:-}
 EOF
-chmod 0600 $HOME/.netrc
+  chmod 0600 $HOME/.netrc 2>/dev/null
+fi
 unset VELA_NETRC_MACHINE
 unset VELA_NETRC_USERNAME
 unset VELA_NETRC_PASSWORD
 unset VELA_BUILD_SCRIPT
+set -e
```
wass3rw3rk, line 115: i think this would have some potentially unintended consequences unless people take control and set these modes in their own script/commands

wass3rw3rk (changes requested):

compiler/native/script.go
ecrupper: Meh, just a substitution check that I don't think is necessary given we always set these. I'll remove it

compiler/native/script.go
ecrupper: Hm yeah this was an attempt to create the path if it didn't exist and we had access... now I'm seeing `touch` is only for the file itself, not the path.

compiler/native/script.go
ecrupper: I am confused. I thought `set -e` was default and this is just setting it back?

compiler/native/script.go
wass3rw3rk: i failed to realize we `set -e`. so the comments related to (un)setting this flag are to be ignored :disappear:

compiler/native/script.go
wass3rw3rk: it's not (usually), but we do set it manually, which i failed to realize :D

compiler/native/script.go
wass3rw3rk: even if it's not set it should result in empty value either way afaik, but sounds good

wass3rw3rk (dismissed):

wass3r (approved):

timhuynh94 (approved):

PR #1046, opened 2024-01-25
fix(schedule): better handling of commit status limit for schedule build

timhuynh94:
Ref [issue 841](https://github.com/go-vela/community/issues/841)

In GitHub you can update the commit status for a sha/context up to 1,000 times, see https://docs.github.com/en/enterprise-server@3.7/rest/commits/statuses#create-a-commit-status. The likelihood for this limit to be hit is higher with scheduled builds resulting in log errors like:

{"level":"error","msg":"unable to set commit status for build <org>/<repo>/9001: POST https://<domain>/api/v3/repos/<org>/<repo>/statuses/8968ddf24f66786c669ebd26fe83e3d4cf03a19c: 422 Validation Failed [{Resource:Status Field: Code:custom Message:This SHA and context has reached the maximum number of statuses.}]","time":"2023-07-31T19:00:13Z"}

The likelihood for this scenario to occur depend on schedule frequency and repository activity (or lack thereof).

After some discussion about potential solution, we decide not to update status for scheduled at all.

Solution:
Don't send a status update to SCM for scheduled builds including pending status at the start. Most folks check Vela and they will see historical runs.


api/webhook/post.go
```diff
@@ -766,10 +766,12 @@ func PostWebhook(c *gin.Context) {
...
-	// send API call to set the status on the commit
-	err = scm.FromContext(c).Status(ctx, u, b, repo.GetOrg(), repo.GetName())
-	if err != nil {
-		logrus.Errorf("unable to set commit status for %s/%d: %v", repo.GetFullName(), b.GetNumber(), err)
+	// send API call to set the status on the commit except for scheduled build
+	if b.GetEvent() != constants.EventSchedule {
+		err = scm.FromContext(c).Status(ctx, u, b, repo.GetOrg(), repo.GetName())
+		if err != nil {
+			logrus.Errorf("unable to set commit status for %s/%d: %v", repo.GetFullName(), b.GetNumber(), err)
+		}
```
ecrupper, line 774: We won't be receiving a webhook for schedule events

ecrupper (changes requested):
QQ: does CreateBuild actually work? Otherwise, one comment on post.go. LGTM

timhuynh94:
> QQ: does CreateBuild actually work? Otherwise, one comment on post.go. LGTM

yes! We don't want to set the schedule build status to pending to begin with if we're not updating it!

ecrupper (approved):

wass3rw3rk (approved):

PR #1052, opened 2024-02-01
fix(compiler): remove template param from CompileLite

ecrupper:
Context: with the update to the pipeline [validation function](https://github.com/go-vela/server/pull/1036), leveraging `CompileLite(_data, false, false)` with templates present in the pipeline became an impossibility since the resulting pipeline would contain a step without an image. 

Two approaches: revert that enhancement to validation or "fail forward" so to speak and fetch templates if they exist / are replaced locally. 

A downstream benefit of this is the removal of the arguably confusing `--template` boolean flag in the CLI.

I think one could argue that the benefit of an "ignore templates" option is the opportunity to validate pipeline changes without worrying about pulling in templates. I think this is a fair point, but I also think combining [our YAML schema](https://github.com/go-vela/types/blob/main/cmd/schema/main.go) with any sort of standard YAML validator is perfectly suitable. 

Let me know if you disagree!

### DIFF
--- a/api/pipeline/compile.go
+++ b/api/pipeline/compile.go
@@ -95,7 +95,7 @@ func CompilePipeline(c *gin.Context) {
 	compiler := compiler.FromContext(c).Duplicate().WithCommit(p.GetCommit()).WithMetadata(m).WithRepo(r).WithUser(u)
 
 	// compile the pipeline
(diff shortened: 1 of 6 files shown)


claire1618 (approved):

wass3rw3rk (approved):
i like it (and the removal of the `--template` flag)

PR #1054, opened 2024-02-02
fix: changing ProcessWebhook so that the deployment ref is only set once

claire1618:
Deleting the second instance of deployment.SetRef() due to errors it was causing when in dev when a deployment was created in prod


ecrupper (approved):
nice lgtm

wass3rw3rk (approved):

PR #1062, opened 2024-02-15
enhance(approve build): add default setting flag and implement first-time option

ecrupper:
One important thing to note in this PR is that the contributors list [will not always be up to date](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repository-contributors). It can take a few hours for a new contributor to be reflected on the list we pull from.


jbrockopp (approved):
LGTM

wass3rw3rk (approved):

PR #1063, opened 2024-02-16
refactor: create re-usable `CompileAndPublish` function

ecrupper:
[Context](https://github.com/go-vela/community/issues/874)

There's an opportunity to re-use the same several-hundred lines of code that are present in `CreateBuild`, `RestartBuild`, `processSchedule`, and `PostWebhook`. The differences between each were surprisingly minimal, but that's not to say this wasn't a very tedious process.

Very much open to suggestions on overall style, naming, etc.

I think the value of this refactor can be summarized by a few points:

* changes to the build creation process will now automatically be applied to the 4 separate entry points, eliminating the occasional disparity between them.
* easier to squash bugs related to the process
* simpler to follow for developers new to the code base (I hope 😅 )


scm/github/webhook.go
```diff
@@ -437,8 +437,8 @@ func (c *client) processIssueCommentEvent(h *library.Hook, payload *github.Issue
 		fmt.Sprintf("https://%s/%s/settings/hooks", h.GetHost(), payload.GetRepo().GetFullName()),
 	)
 
-	// skip if the comment action is deleted
-	if strings.EqualFold(payload.GetAction(), "deleted") {
+	// skip if the comment action is deleted or not part of a pull request
+	if strings.EqualFold(payload.GetAction(), "deleted") || !payload.GetIssue().IsPullRequest() {
```
wass3r, line 441: this tweak would make this PR a breaking change, right?

api/webhook/post.go
```diff
@@ -273,399 +272,52 @@ func PostWebhook(c *gin.Context) {
...
-	b, err = database.FromContext(c).GetBuildForRepo(ctx, repo, b.GetNumber())
-	if err != nil {
-		retErr := fmt.Errorf("%s: failed to get new build %s/%d: %w", baseErr, repo.GetFullName(), b.GetNumber(), err)
-		util.HandleError(c, http.StatusInternalServerError, retErr)
-
-		h.SetStatus(constants.StatusFailure)
-		h.SetError(retErr.Error())
+	if !pushed {
+		// if the build was not pushed to the queue, it was skipped
+		h.SetStatus(constants.StatusSkipped)
```
wass3r, line 316: you would reach this when `err` is nil and `pushed` is false which looks to be possible in two instances:
1. build got skipped, or
2. unable to set route for build (https://github.com/go-vela/server/pull/1063/files#diff-0e4342c1b8362db4ef095686b38347fc9fada69877f4432cdcd26a99b9c379feR434)

i'm guessing we probably want to update 2. to actually return an error?

wass3r (changes requested):

scm/github/webhook.go
ecrupper: I was looking at this... this actually just makes the return nicer. If you make a comment on an issue, this is the current return: 
```json
{
  "error":  "unable to process webhook: failed to create pipeline for EastonCrupper/myvela: empty pipeline commit provided"
}
```

Now it will be `no build to process`. Looks like ever since pipelines, the `comment` event got a little wonky outside PR contexts.

api/build/create.go
```diff
@@ -109,297 +102,46 @@ func CreateBuild(c *gin.Context) {
...
-
-			return
-		}
+	// create config
+	config := CompileAndPublishConfig{
+		Build:    input,
+		Repo:     r,
+		Metadata: m,
+		BaseErr:  "unable to restart build",
+		Source:   "restart",
```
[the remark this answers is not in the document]
ecrupper: Yes, thank you!

api/webhook/post.go
ecrupper: Ah, yep, you're right. 2 should be an error like it is today

scm/github/webhook.go
wass3r: makes sense. i just quick scanned the docs and we do only talk about comments in the context of PRs as far as I can tell, so i think we're good. thanks

wass3r (approved):
lgtm

PR #1064, opened 2024-02-21
fix: abstract build component status updates for build approval and cancelation

ecrupper:
During cancel we go through all the steps and services and update their statuses to `canceled`. A similar process is needed for `gatekeepBuild` for builds that need approval. So I abstracted the process into its own function to avoid duplication. 

We have a routine cleanup job that leverages the `clean` endpoint, which searches for stale pending/running builds, and it was "cleaning" the build components of `pending approval` status because they were still set to `pending`, which is done during `PlanBuild`.


wass3rw3rk (approved):

KellyMerrick (approved):
lgtm

PR #1065, opened 2024-02-22
fix(scm/webhook): change fork determination logic

ecrupper:
A PR build on a repo that is a fork of another repo is interpreted as "from a fork", even if the head and base are both from the same repo.

Fixed that and added tests.


wass3rw3rk (approved):

KellyMerrick (approved):
lgtm

PR #1071, opened 2024-02-28
feat: add types, fix errors, add conditional status report

ecrupper:


wass3rw3rk:
if this doesn't need review atm, can we change to draft PR?

wass3rw3rk (changes requested):
one suggestion, otherwise looks fine to me.

PR #1075, opened 2024-03-04
fix(build-approval): correct approved_by and disallow self-approval

ecrupper:
Make sure the approver is not the owner of the repo.

Also make sure the approver is not the sender of the build, as that would be counter-intuitive to the fact that we have a `no-write` option in place.


wass3rw3rk (approved):

KellyMerrick (approved):
lgtm

PR #1076, opened 2024-03-05
enhance(compiler): add ruledata option to CompileLite

ecrupper:
I'm visualizing a UI / CLI experience where users can use ruledata filters to watch their pipelines purge themselves in an interactive way:

`vela validate pipeline --event push` would return the pipeline that they can expect to run for a `push` event. 

We could expand the `pipeline` tab in the UI to offer a similar process.


timhuynh94 (approved):
LGTM

plyr4 (approved):

PR #1083, opened 2024-03-14
fix(vault): fix for integer64 fields

ecrupper:
Vault was turning our `int64` inputs into `json.Number` outputs, which caused all the type checking to fail and set those values to 0.

Also added a test to confirm existence of fields for library.Secret in api.Secret data

### DIFF
--- a/secret/vault/vault.go
+++ b/secret/vault/vault.go
@@ -3,6 +3,7 @@
 package vault
 
 import (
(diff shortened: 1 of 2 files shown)


wass3rw3rk (approved):
great test additions too 🥇

timhuynh94 (approved):

PR #1086, opened 2024-03-15
fix(vault): parse pre-v0.23 vault secrets

ecrupper:
If a vault secret has not been updated since before `v0.23`, it will not have `allow_events` or `allow_substitution`. We can infer these values by converting event slice into the allow events mask and by setting substitution setting to the same setting as the command setting.

I didn't include any update process on the Vela side. Fortunately the UI submits the entire secret form on update, so users will slowly update their vault secrets on their own accord. 

We could also consider creating a migration script of sorts to prevent this code, but we're also supporting v1 and v2 of Vault in code as well. Not saying that's an excuse, but perhaps revisiting our backwards compatibility in general could be a good future issue 🤷

### DIFF
--- a/secret/vault/vault.go
+++ b/secret/vault/vault.go
@@ -174,6 +174,7 @@ func secretFromVault(vault *api.Secret) *library.Secret {
 		events, ok := data["events"]
 		if ok {
 			allowEventsMask := int64(0)
+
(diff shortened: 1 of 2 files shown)


jbrockopp (approved):
LGTM

plyr4 (approved):

PR #1087, opened 2024-03-19
fix(deployments): set deployment created event action

plyr4:
adding support for `event: [deployment:created]`
depends on: https://github.com/go-vela/types/pull/365

reference: https://go-vela.github.io/docs/usage/repo_settings/#webhook-events


go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/adhocore/gronx v1.8.1
 	github.com/alicebob/miniredis/v2 v2.32.1
 	github.com/aws/aws-sdk-go v1.51.0
-	github.com/buildkite/yaml v0.0.0-20181016232759-0caa5f0796e3
+	github.com/buildkite/yaml v0.0.0-20230306222819-0e4e032d4835
```
ecrupper, line 13: There a reason for this change? I noticed that types still uses the `0caa` version

plyr4: hm no, but i fixed some merge conflicts in go modules. probably an auto merge mishap. let me check :ty:

plyr4: okay fixed, didnt mean to pull those in with go mod tidy. thank youuuuuu

wass3rw3rk (approved):

PR #1088, opened 2024-03-20
fix(api/webhook): build approval for fork-no-write

jbrockopp:
This change resolves an issue with the new build approval feature introduced in `v0.23`:

https://github.com/go-vela/community/blob/main/releases/v0.23.md

The issue is builds triggered from a fork by a user with `admin` perms were being forced to wait for an approval.

This is unexpected behavior because the repos in question have a build approval policy of `fork-no-write`:

https://github.com/go-vela/types/blob/v0.23.2/constants/repo.go#L26-L28

After successfully reproducing the issue locally, I was able to determine the code causing this:

https://github.com/go-vela/server/blob/1809638e7e729857c4ae9fb441796644ddb9aa1a/api/webhook/post.go#L773-L786

This code is setup to check if the build sender has access to the repo:

* if they don't, then we call the function to gatekeep the build until an approval is provided
* if they do, then we `fallthrough` to the `first-time` approval policy `case` statement

After this change, we'll no longer fallthrough to the `first-time` approval policy `case` statement.


ecrupper (approved):
thanks!

JordanSussman (approved):

PR #1090, opened 2024-03-21
feat: step status reporting + fix pull request context overwrite bug

ecrupper:
draft.. need to add testing. will update this description later. want this up for context on types PR

### DIFF
--- a/api/build/create.go
+++ b/api/build/create.go
@@ -320,7 +320,7 @@ func CreateBuild(c *gin.Context) {
 	input.SetPipelineID(pipeline.GetID())
 
 	// create the objects from the pipeline in the database
-	err = PlanBuild(ctx, database.FromContext(c), p, input, r)
+	err = PlanBuild(ctx, database.FromContext(c), scm.FromContext(c), p, input, r)
 	if err != nil {
 		util.HandleError(c, http.StatusInternalServerError, err)
 
--- a/compiler/native/validate.go
+++ b/compiler/native/validate.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/hashicorp/go-multierror"
 
+	"github.com/go-vela/types/constants"
 	"github.com/go-vela/types/yaml"
 )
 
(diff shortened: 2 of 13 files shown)


api/build/compile_publish.go
```diff
@@ -320,12 +320,11 @@ func CompileAndPublish(
...
 			}
 
-			return false,
-				nil,
+			return nil,
 				&types.Item{
 					Build: b,
 				},
-				nil
+				fmt.Errorf(skip)
```
plyr4, line 327: should we use `errors.New(skip)` ?

im surprised golangci lets you use a format func without a format string

ecrupper: Cool with me

api/build/create.go
```diff
@@ -129,6 +131,13 @@ func CreateBuild(c *gin.Context) {
 		queue.FromContext(c),
 	)
 
+	// check if build was skipped
+	if strings.EqualFold(item.Build.GetStatus(), constants.StatusSkipped) {
+		c.JSON(http.StatusOK, err.Error())
```
plyr4, line 136: i know err is not nil here, but it scares me to not check it lol

compiler/native/validate.go
```diff
@@ -136,12 +141,25 @@ func validateSteps(s yaml.StepSlice) error {
 			continue
 		}
 
+		if s, ok := reportMap[step.ReportAs]; ok {
+			return fmt.Errorf("report_as to %s for step %s is already targeted by step %s", step.ReportAs, step.Name, s)
```
plyr4, line 145: nice

api/build/create.go
ecrupper: ha, will throw that in there

plyr4 (approved):

plyr4:
> Question I had: should a skipped step be a success?
in my opinion i dont think so, that might make required status checks auto pass when it never ran

KellyMerrick (approved):
lgtm

JordanSussman:
@ecrupper do you think it makes sense to open up another community issue for expanding this functionality to work with stages?

ecrupper:
> @ecrupper do you think it makes sense to open up another community issue for expanding this functionality to work with stages?

Definitely. Also somewhat related would be https://github.com/go-vela/community/issues/910 with the idea of a "status" for a stage.

PR #1092, opened 2024-03-21
fix: swagger typo in CreateWorker

plyr4:
simple typo fix 
v0.23 api-spec.json causes an error in swagger

```
Semantic error at paths./api/v1/workers.post.responses.201.schema.$ref
$ref paths must begin with `#/`
```


ecrupper (approved):

jbrockopp (approved):
LGTM

PR #997, opened 2023-11-01
refactor(types)!: move worker type to server and nest API object

ecrupper:
Proposal ref: https://github.com/go-vela/community/pull/639

@jbrockopp saw this refactor as a good opportunity to migrate off of the `go-vela/types` repo a while back.

Starting with worker since it is a very low user-requested resource but is still leveraged often enough by the Vela application that we would know if something wasn't quite right very early.

I was also itching to start with worker, as the [admin page](https://github.com/go-vela/ui/pull/707) would benefit greatly from clickable build links sourced from the `ListWorkers` API response.

As a visual:

OLD
```json
{
    "id": 1,
    "hostname": "worker",
    "address": "http://worker:8080/",
    "routes": [
      "vela"
    ],
    "active": true,
    "status": "busy",
    "last_status_update_at": 1698852759,
    "running_builds": [
      "1"
    ],
    "last_build_started_at": 1698852759,
    "last_build_finished_at": 0,
    "last_checked_in": 1698852748,
    "build_limit": 1
  }
```

NEW
```json
{
    "id": 1,
    "hostname": "worker",
    "address": "http://worker:8080/",
    "routes": [
      "vela"
    ],
    "active": true,
    "status": "busy",
    "last_status_update_at": 1698852759,
    "running_builds": [
      {
        "id": 1,
        "repo_id": 1,
        "pipeline_id": 1,
        "number": 1,
        "parent": 0,
        "event": "push",
        "event_action": "",
        "status": "running",
        "error": "",
        "enqueued": 1698852759,
        "created": 1698852759,
        "started": 1698852759,
        "finished": 0,
        "deploy": "",
        "deploy_payload": {},
        "clone": "https://github.com/Octocat/myvela.git",
        "source": "https://github.com/Octocat/myvela/commit/123abc",
        "title": "push received from https://github.com/Octocat/myvela",
        "message": "Update .vela.yml",
        "commit": "123abc",
        "sender": "Octocat",
        "author": "Octocat",
        "email": "octocat@gmail.com",
        "link": "http://localhost:8888/Octocat/myvela/1",
        "branch": "main",
        "ref": "refs/heads/main",
        "base_ref": "",
        "head_ref": "",
        "host": "worker",
        "runtime": "docker",
        "distribution": "linux"
      }
    ],
    "last_build_started_at": 1698852759,
    "last_build_finished_at": 0,
    "last_checked_in": 1698852748,
    "build_limit": 1
  }
```

PR #1095, opened 2024-03-26
refactor(repo)!: nested repository with migration from types

ecrupper:
Continuation of:
- https://github.com/go-vela/server/pull/997

Context:
https://github.com/go-vela/community/pull/639


ecrupper:
Putting in draft right now. Functionally it's there, but I'd like to work out more comments, address some tests, and search for more optimizations

timhuynh94 (dismissed):
LGTM

api/admin/repo.go
```diff
@@ -7,11 +7,10 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/go-vela/server/api/types"
```
plyr4, line 10: as we move everything to this pattern, any thoughts on giving _every_ types import the `api` alias?

ecrupper: I was operating on the following pattern:
- `go-vela/server/api` —> types
- `anything else` —> api

But I'm open to using `api` always if people would prefer that

plyr4: cool that works for me, not blocking either way

database/repo/get.go
```diff
@@ -5,21 +5,21 @@ package repo
...
 	e.logger.Tracef("getting repo %d from the database", id)
 
 	// variable to store query results
-	r := new(database.Repo)
+	r := new(Repo)
 
 	// send query to the database and store result in variable
 	err := e.client.
 		Table(constants.TableRepo).
+		Preload("Owner").
```
plyr4, line 22: i wonder if this should be a constant somewhere, like the table name references

plyr4:
```suggestion
		Preload(constants.PreloadsOwner).
```

database/repo/repo.go
```diff
@@ -38,6 +75,33 @@ type (
...
+		Visibility   sql.NullString `sql:"visibility"`
+		Private      sql.NullBool   `sql:"private"`
+		Trusted      sql.NullBool   `sql:"trusted"`
+		Active       sql.NullBool   `sql:"active"`
+		AllowEvents  sql.NullInt64  `sql:"allow_events"`
+		PipelineType sql.NullString `sql:"pipeline_type"`
+		PreviousName sql.NullString `sql:"previous_name"`
+		ApproveBuild sql.NullString `sql:"approve_build"`
+
+		Owner database.User `gorm:"foreignKey:UserID"`
```
plyr4, line 103: interesting that the foreign key isnt snake-case

scm/github/changeset.go
```diff
@@ -8,12 +8,13 @@ import (
...
 	"github.com/sirupsen/logrus"
 
+	api "github.com/go-vela/server/api/types"
 	"github.com/go-vela/types/library"
 	"github.com/google/go-github/v61/github"
 )
 
 // Changeset captures the list of files changed for a commit.
-func (c *client) Changeset(ctx context.Context, u *library.User, r *library.Repo, sha string) ([]string, error) {
+func (c *client) Changeset(ctx context.Context, u *library.User, r *api.Repo, sha string) ([]string, error) {
```
wass3rw3rk, line 17: question: could we get rid off `u *library.User` since that is likely coming from `r.GetOwner()` now? i know there's some other place in /scm/*, so i won't repeat the question in those spots.

wass3rw3rk (changes requested):
looks good overall.. a few minor things.

any thoughts around moving some of the "internal" types that live in `/internal` to `/internal/types/` or another location?

api/webhook/post.go
```diff
@@ -308,7 +309,7 @@ func PostWebhook(c *gin.Context) {
 	)
 
 	// capture the build, repo, and user from the items
```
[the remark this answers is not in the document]
wass3rw3rk: for context, the following here was changed to exclude user, but it wouldn't let me select both lines 😢

api/types/string.go
```diff
@@ -0,0 +1,77 @@
...
+	case uint32:
+		return strconv.FormatUint(uint64(v), 10)
+	case uint64:
+		return strconv.FormatUint(v, 10)
+	case []interface{}:
+		return unmarshalSlice(v)
+	default:
+		return unmarshalMap(v)
+	}
+}
```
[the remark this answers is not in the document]
ecrupper: The type switching prevents this unfortunately 😞

database/repo/get.go
ecrupper: Going to implement this idea when `go-vela/types/constants` is migrated

plyr4 (approved):

timhuynh94 (dismissed):

scm/service.go
wass3rw3rk, line 1: does it make sense to drop library.User from any method that takes User and Repo?

ecrupper: Some yes, some no. I will merge this and follow up with a PR that focuses on just that

ecrupper: Some yes some no. Let me address that

ecrupper: addressed in latest commit. There were surprisingly few that needed updating

plyr4 (approved):

wass3rw3rk (approved):

PR #1096, opened 2024-03-27
enhance(templates): inject template name as variable

ecrupper:
Example pipeline:

`.vela.yml`
```yaml
version: "1"

metadata:
  render_inline: true

templates:
  - name: example
    source: example.yml
    type: file
```

`example.yml`
```yaml
metadata:
  template: true

stages:
  wait:
    steps:
      - name: run
        image: alpine
        commands:
          - sleep 20
          - printenv

  test:
    needs: [ wait ]
    steps:
      - name: run
        image: alpine
        commands:
          - echo "Hello World"
```

Due to the fact that the steps and stages get modified to include the template name, this configuration will not work due to `wait` stage being transformed to `example_wait`. With this change, that part can be modified to

`needs: [{{ .VELA_TEMPLATE_NAME }}_wait]`

This is beneficial to developing widely used templates without instructing all adopters to use the same template name.

### DIFF
--- a/compiler/native/compile.go
+++ b/compiler/native/compile.go
@@ -191,6 +191,14 @@ func (c *client) compileInline(p *yaml.Build, depth int) (*yaml.Build, error) {
 			format = constants.PipelineTypeGo
 		}
 
+		// initialize variable map if not parsed from config
+		if len(template.Variables) == 0 {
+			template.Variables = make(map[string]interface{})
+		}
+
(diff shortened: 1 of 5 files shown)


plyr4 (approved):

KellyMerrick (approved):
lgtm

timhuynh94 (approved):

PR #1097, opened 2024-03-29
fix(scm): skipped map to failure for step status + remove deploy

ecrupper:
Addressing [this comment](https://github.com/go-vela/server/pull/1090#issuecomment-2026056483) from the [original PR](https://github.com/go-vela/server/pull/1090) to have steps that are skipped report as failures. 

Also ignoring deployment events for step status reporting. We report the build status to the deployment rather than the commit, and given that there are not multiple deployment statuses, it makes sense to just ignore. This also prevents status limits to the same commit if a repo rarely makes changes but deploys often.

### DIFF
--- a/api/build/compile_publish.go
+++ b/api/build/compile_publish.go
@@ -4,6 +4,7 @@ package build
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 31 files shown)


timhuynh94 (approved):

KellyMerrick (approved):
lgtm

PR #1100, opened 2024-04-05
fix(api/pipeline): correct type for path query param + add testing

ecrupper:
Immediate follow up to https://github.com/go-vela/server/pull/1076, as splitting a query value on a comma is very finicky given URL encoding and fortunately Gin and `"github.com/google/go-querystring/query"` (sdk-go) both handle multi key queries as a slice input.


plyr4 (approved):
good call

wass3rw3rk (approved):

PR #1105, opened 2024-04-11
fix: handle errors from CompileAndPublish

plyr4:
fixes a nil panic when you supply an invalid worker route, introduced in https://github.com/go-vela/server/pull/1063

### DIFF
--- a/api/webhook/post.go
+++ b/api/webhook/post.go
@@ -309,6 +309,11 @@ func PostWebhook(c *gin.Context) {
 		queue.FromContext(c),
 	)
 
+	// error handling done in CompileAndPublish
+	if err != nil {
+		return
+	}
+
 	// capture the build and repo from the items
 	b, repo = item.Build, item.Repo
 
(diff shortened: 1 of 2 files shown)


api/webhook/post.go
```diff
@@ -309,6 +309,11 @@ func PostWebhook(c *gin.Context) {
 		queue.FromContext(c),
 	)
 
+	// error handling done in CompileAndPublish
+	if err != nil {
+		return
+	}
```
wass3rw3rk, line 315: it should be one of these responses, right? https://github.com/go-vela/server/blob/cbc016c7b710a04d7c5eb6e0a02e4914fa761ceb/api/webhook/post.go#L48-L68

plyr4: yeah the other change makes it return writes a 400 error inside of `CompileAndPublish` (admittedly, not my favorite to handle errors inside the func)

https://github.com/go-vela/server/pull/1105/files#diff-0e4342c1b8362db4ef095686b38347fc9fada69877f4432cdcd26a99b9c379feR429-R430
unless im misunderstanding your question

wass3rw3rk: yea, my fault.. agree - not great. every error in CompileAndPublish also returns API response. seems like it would be good to return the error (and code) and let us handle the API response in one spot at this level.

wass3rw3rk (approved):

ecrupper (approved):

PR #1106, opened 2024-04-12
refactor(user)!: migrate from types repo

ecrupper:
Continuing efforts from:
- https://github.com/go-vela/server/pull/1095
- https://github.com/go-vela/server/pull/997

Also removing deprecated `hash` field. Will need to be added to migration docs, hence the breaking change marker.


api/build/skip.go
```diff
@@ -9,7 +9,7 @@ import (
 // SkipEmptyBuild checks if the build should be skipped due to it
 // not containing any steps besides init or clone.
 //
-//nolint:goconst // ignore init and clone constants
+
```
wass3rw3rk, line 12: minor: linter with fix _does some work_ but for stuff like this requires some manual intervention unfortunately. i'd remove line 11 + 12.

wass3rw3rk (dismissed):
super minor thing

timhuynh94 (dismissed):

api/build/skip.go
ecrupper: oof, that's too bad

timhuynh94 (approved):

PR #1107, opened 2024-04-12
fix: return code in CompileAndPublish rather than handleError

ecrupper:
It'll be easier to follow that `CompileAndPublish` errors are returned with a code to handle rather than handle them itself.

I also made all the responses match across create, restart, and PostWebhook:
- 200 = skipped 
- 201 = created
- 400 = bad request
- 401 = unauthorized owner
- 404 = missing resource
- 429 = build limit reached
- 500 = error processing or server unreachable

Open to suggestions on that, but the inconsistencies should be addressed at least.


api/build/create.go
```diff
@@ -133,15 +139,14 @@ func CreateBuild(c *gin.Context) {
 	)
 
 	// check if build was skipped
-	if err != nil && strings.EqualFold(item.Build.GetStatus(), constants.StatusSkipped) {
+	if err != nil && code == http.StatusOK {
```
plyr4, line 142: ❤️

api/build/create.go
```diff
@@ -133,15 +139,14 @@ func CreateBuild(c *gin.Context) {
...
 		c.JSON(http.StatusOK, err.Error())
 
 		return
 	}
 
-	// error handling done in CompileAndPublish
 	if err != nil {
-		return
+		util.HandleError(c, code, err)
```
plyr4, line 149: do we still need the return here? so we dont double write to the response

api/build/restart.go
```diff
@@ -129,9 +138,8 @@ func RestartBuild(c *gin.Context) {
 		queue.FromContext(c),
 	)
 
-	// error handling done in CompileAndPublish
 	if err != nil {
-		return
+		util.HandleError(c, code, err)
```
plyr4, line 142: same question, do we need to return here?

api/build/create.go
ecrupper: Yes

api/build/create.go
```diff
@@ -51,24 +49,32 @@ import (
 //   - ApiKeyAuth: []
 // responses:
 //   '200':
-//     description: Request processed but build was skipped
+//     description: Successfully received the webhook but build was skipped
```
wass3rw3rk, line 52: you can create builds without webhooks, so i don't know if it makes sense to single out the webhook scenario here (and elsewhere).

wass3rw3rk (changes requested):

plyr4 (approved):

wass3rw3rk (approved):

PR #1108, opened 2024-04-15
feat(schedule)!: show schedule errors and nest object

wsan3:
### Description
Whenever a user has a YAML bug in their pipeline, or any other build-breaking error that short-circuits pipeline execution, they can usually find these error in the Audit tab, where we populate the hook with the error message.

This is currently not available for schedules. Instead, the error is just logged.

### Desired View
Screenshot 2024-04-26 at 10 37 57 AM

### Notable Changes

- Moved `Schedule` out of `go-vela/types` and into `go-vela/server`
- Turned `Schedule` into a nested object containing `Repo` and `User` information
- Replaced `reflect.DeepEqual` with `cmp.Diff` in test files

### Related
https://github.com/go-vela/community/issues/872


compiler/native/validate.go
```diff
@@ -16,7 +16,7 @@ func (c *client) Validate(p *yaml.Build) error {
 	var result error
 	// check a version is provided
 	if len(p.Version) == 0 {
-		result = multierror.Append(result, fmt.Errorf("no \"version:\" YAML property provided"))
```
ecrupper, line 19: I think we want to keep this. Any reason you removed it?

ecrupper (commented):
Nice. One small thing I noticed

compiler/native/validate.go
wsan3: Good catch, I think I did a repo wide replacement of that string and accidently changed that.

cmd/vela-server/schedule.go
```diff
@@ -203,3 +212,20 @@ func processSchedule(ctx context.Context, s *library.Schedule, compiler compiler
 
 	return nil
 }
+
+func logError(ctx context.Context, database database.Interface, err error, schedule *api.Schedule) {
```
plyr4, line 216: the name of this function seems vague compared to what it actually does (log _and_ update the db record). maybe `handleError` would fit better? or move that db call outside of this function?

wsan3: Good suggestion, `handleError` is a more appropriate name. I'll use that.

cmd/vela-server/schedule.go
```diff
@@ -117,15 +116,25 @@ func processSchedules(ctx context.Context, start time.Time, compiler compiler.En
...
+
+			continue
+		}
+
+		schedule.SetError("")
+
+		// send API call to update schedule with the error message field cleared
+		_, err = database.UpdateSchedule(ctx, schedule, true)
+		if err != nil {
+			handleError(ctx, database, err, schedule)
```
wass3r, line 137: i don't think we want to keep updating schedules (read: make database calls) on a happy path to update error field. how about wrapping this in a conditional to only execute if the error field is non-empty on a given schedule?

database/integration_test.go
```diff
@@ -1278,19 +1278,33 @@ func testSchedules(t *testing.T, db Interface, resources *Resources) {
...
+		_, err := db.CreateRepo(context.TODO(), repo)
+		if err != nil {
+			t.Errorf("unable to create repo %d: %v", repo.GetID(), err)
+		}
+	}
 
 	// create the schedules
 	for _, schedule := range resources.Schedules {
-		_, err := db.CreateSchedule(ctx, schedule)
+		_, err := db.CreateSchedule(context.TODO(), schedule)
```
wass3r, line 1299: what is the reason for inlining the context?

database/schedule/opts.go
```diff
@@ -22,6 +22,16 @@ func WithClient(client *gorm.DB) EngineOpt {
 	}
 }
 
+// WithEncryptionKey sets the encryption key in the database engine for Repos.
+func WithEncryptionKey(key string) EngineOpt {
+	return func(e *engine) error {
+		// set the encryption key in the repo engine
+		e.config.EncryptionKey = key
```
wass3r, line 29:
```suggestion
// WithEncryptionKey sets the encryption key in the database engine for Schedules.
func WithEncryptionKey(key string) EngineOpt {
	return func(e *engine) error {
		// set the encryption key in the schedule engine
		e.config.EncryptionKey = key
```

looks like a previous oversight here as well: https://github.com/go-vela/server/blob/8ad123451469039e0125af1943e916c211415642/database/build/opts.go#L25-L28

wass3r (changes requested):
couple minor things. lgtm otherwise.

database/integration_test.go
wsan3: I noticed that the majority of the other test functions had the context inline and wanted to be consistent.

database/schedule/opts.go
wsan3: Good catches

cmd/vela-server/schedule.go
wsan3: Yeah that makes sense. I will update that.

PR #1109, opened 2024-04-16
fix(repo): sanitize owner information

ecrupper:
Even though we have the JSON tag of `-` to prevent marshaling of token / refresh token, we may want to further censor information.

This `Sanitize()` function is actually never used, so I slapped it on the `ToAPI()` call for repo owners and extended it to exclude `favorites` and `admin` because not only can favorites get huge, they also shouldn't really be public info.


KellyMerrick (approved):
lgtm

plyr4 (approved):

PR #1110, opened 2024-04-17
feat!: db driven platform settings

plyr4:
see: https://github.com/go-vela/community/pull/921

- creates a singleton database record that persists platform settings across all servers
- settings are initialized using the standard flag flow
- middleware refreshes settings on each request

introduces some new admin api endpoints
- `api/v1/admin/settings` GET PUT to retrieve and update
- `api/v1/admin/settings` DELETE to "restore" the settings record to the original values stored in the server environment 
- there is no POST because a user should never be responsible for creating settings records

**NOTE**: more fields need to be migrated, this is not an exhaustive implementation!


plyr4:
putting in draft while i build out the tooling (ui/cli) and add more testing

ecrupper (dismissed):

api/admin/settings.go
```diff
@@ -0,0 +1,144 @@
...
+			s.CloneImage = input.CloneImage
+		}
+
+		if input.TemplateDepth != nil {
+			s.TemplateDepth = input.TemplateDepth
+		}
+
+		if input.StarlarkExecLimit != nil {
+			s.StarlarkExecLimit = input.StarlarkExecLimit
+		}
```
ecrupper, line 114: Any reason we don't use setters here and do later in the function?

database/testutils/mock_args.go
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package testutils
```
ecrupper, line 3: Nice

database/settings/settings.go
```diff
@@ -0,0 +1,285 @@
...
+	Compiler struct {
+		CloneImage        sql.NullString `json:"clone_image" sql:"clone_image"`
+		TemplateDepth     sql.NullInt64  `json:"template_depth" sql:"template_depth"`
+		StarlarkExecLimit sql.NullInt64  `json:"starlark_exec_limit" sql:"starlark_exec_limit"`
+	}
+
+	// Queue is the database representation of queue settings.
+	Queue struct {
+		Routes pq.StringArray `json:"routes" sql:"routes" gorm:"type:varchar(1000)"`
+	}
```
ecrupper, line 80: Can these be moved to `database/types` for consistency?

cmd/vela-server/server.go
```diff
@@ -97,7 +100,51 @@ func server(c *cli.Context) error {
...
+			logrus.WithError(err).Error("unable to get platform settings")
+		}
+
+		// create initial settings record
+		s = new(settings.Platform)
+
+		// singleton record ID should always be 1
+		s.SetID(1)
+
+		s.SetCreatedAt(time.Now().UTC().Unix())
```
ecrupper, line 116: Should we also nullify `updated_by` when this happens?

ecrupper (changes requested):
A couple general questions but also a request to move the platform settings types into the types DB package. I like how you have it, but due to import cycles in other pkgs, the `types` package was born and I'd like the consistency.

cmd/vela-server/server.go
plyr4: you think empty string or something hardcoded and intentional like `vela-server` ?

database/settings/settings.go
plyr4: done

api/admin/settings.go
plyr4: laziness? lack of attention to detail? not enough coffee? :) fixed ty

ecrupper (approved):

cmd/vela-server/server.go
```diff
@@ -97,7 +100,60 @@ func server(c *cli.Context) error {
...
+
+	// update any internal settings
+	queue.SetSettings(s)
+	compiler.SetSettings(s)
+
 	router := router.Load(
+		// settings middleware must be first
+		// todo: figure out if deterministic order matters
+		//
+		// we arleady depend on establish running in order, so mayube this is fine
```
plyr4, line 155: remove

cmd/vela-server/server.go
```diff
@@ -97,7 +100,60 @@ func server(c *cli.Context) error {
...
 
+	jitter := wait.Jitter(5*time.Second, 0.5)
+
+	logrus.Infof("sleeping for %v before initializing settings", jitter)
+
+	// sleep for a duration of time before initializing settings
+	time.Sleep(jitter)
+
+	s, err := database.GetSettings(context.Background())
+	if c.Bool("vela-reinitialize-settings-on-startup") || s == nil || err != nil {
```
plyr4, line 111: this bool would need to update, not create a record

cmd/vela-server/server.go
```diff
@@ -97,7 +100,60 @@ func server(c *cli.Context) error {
...
+		s.SetQueue(queueSettings)
+
+		// set repos permitted to be added
+		s.SetRepoAllowlist(c.StringSlice("vela-repo-allowlist"))
+
+		// set repos permitted to use schedules
+		s.SetScheduleAllowlist(c.StringSlice("vela-schedule-allowlist"))
+
+		// create the settings record in the database
+		_, err = database.CreateSettings(context.Background(), s)
```
plyr4, line 141: check if re initializing and UPDATE not CREATE

database/secret/secret_test.go
```diff
@@ -251,29 +249,3 @@ func testEvents() *library.Events {
 		},
 	}
 }
-
```
plyr4, line 254: check for references of this elsewhere

database/settings/table.go
```diff
@@ -0,0 +1,60 @@
...
+	// CreatePostgresTable represents a query to create the Postgres settings table.
+	CreatePostgresTable = `
+CREATE TABLE
+IF NOT EXISTS
+settings (
+	id                		SERIAL PRIMARY KEY,
+	compiler          		JSON DEFAULT NULL,
+	queue         	  		JSON DEFAULT NULL,
+	repo_allowlist	  		VARCHAR(1000),
+	schedule_allowlist	    VARCHAR(1000),
```
plyr4, line 21: indent

router/admin.go
```diff
@@ -12,18 +12,21 @@ import (
...
+// PUT    	 /api/v1/admin/hook
+// PUT    	 /api/v1/admin/repo
+// PUT    	 /api/v1/admin/secret
+// PUT    	 /api/v1/admin/service
+// PUT    	 /api/v1/admin/step
+// PUT    	 /api/v1/admin/user
+// POST   	 /api/v1/admin/workers/:worker/register.
+// GET    	 /api/v1/admin/settings
+// PUT    	 /api/v1/admin/settings
+// DELETE    /api/v1/admin/settings.
```
plyr4, line 29: indent and fix this

router/middleware/settings.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/go-vela/server/database"
+	"github.com/go-vela/server/router/middleware/settings"
+)
+
+// Settings is a middleware function that fetches the latest settings and
+// attaches to the context of every http.Request.
+func Settings(d database.Interface) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		s, err := d.GetSettings(context.Background())
```
plyr4, line 19: can we optimize this? like, what if the API call doesnt NEED settings *or any of the other resources?* 


we are definitely over-querying the database

wass3rw3rk:
looks like you made notes for yourself for follow up? does that mean this is not ready yet?

api/types/settings/compiler.go
```diff
@@ -0,0 +1,112 @@
+// SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
wass3rw3rk: what now? lol

api/admin/settings.go
```diff
@@ -0,0 +1,221 @@
...
+
+		util.HandleError(c, http.StatusBadRequest, retErr)
+
+		return
+	}
+
+	if input.Compiler != nil {
+		if input.CloneImage != nil {
+			s.SetCloneImage(*input.CloneImage)
+		}
```
wass3r, line 110: i'd hate to bring in the docker lib, but some way to validate that the image you are updating exists would be nice, since it could cause a decent amount of breakage.

cmd/vela-server/server.go
```diff
@@ -97,7 +101,61 @@ func server(c *cli.Context) error {
...
 	}
 
+	jitter := wait.Jitter(5*time.Second, 2.0)
+
+	logrus.Infof("retrieving initial platform settings after %v delay", jitter)
+
+	time.Sleep(jitter)
+
+	s, err := database.GetSettings(context.Background())
+	if s == nil || err != nil {
```
wass3r, line 111: instead of `err != nil`, should we check specifically for not found like we do here and then do some of the logic further down: https://github.com/go-vela/server/blob/8a744e48d224197478bb48a468186783b04c44db/api/dashboard/list_user.go#L70

database/settings/table.go
```diff
@@ -0,0 +1,60 @@
...
+IF NOT EXISTS
+settings (
+	id						SERIAL PRIMARY KEY,
+	compiler				JSON DEFAULT NULL,
+	queue					JSON DEFAULT NULL,
+	repo_allowlist			VARCHAR(1000),
+	schedule_allowlist		VARCHAR(1000),
+	created_at				INTEGER,
+	updated_at				INTEGER,
+	updated_by				VARCHAR(250)
```
wass3r, line 24: minor: maybe it's github ui diff, but it looks like there's some indentation issues with the types

database/settings/table.go
```diff
@@ -0,0 +1,60 @@
...
+	CreatePostgresTable = `
+CREATE TABLE
+IF NOT EXISTS
+settings (
+	id						SERIAL PRIMARY KEY,
+	compiler				JSON DEFAULT NULL,
+	queue					JSON DEFAULT NULL,
+	repo_allowlist			VARCHAR(1000),
+	schedule_allowlist		VARCHAR(1000),
+	created_at				INTEGER,
```
wass3r, line 22: `created_by` is missing because it would/should always be `vela-server`?

database/types/settings.go
```diff
@@ -0,0 +1,218 @@
...
+		s.UpdatedAt.Valid = false
+	}
+
+	return s
+}
+
+// ToAPI converts the Settings type
+// to an API Settings type.
+func (s *Platform) ToAPI() *settings.Platform {
+	ss := new(settings.Platform)
```
wass3r, line 125: super minor, but i expected the variable to be `sp` (if we're using shortform)

queue/redis/opts.go
```diff
@@ -40,7 +40,7 @@ func WithChannels(channels ...string) ClientOpt {
 		}
 
 		// set the queue channels in the redis client
-		c.config.Channels = channels
+		c.SetRoutes(channels)
```
wass3r, line 43: do you think we should take the `routes` terminology all the way and update queue to use `routes` instead of `channels`. seems a little odd to see both here.

wass3r (changes requested):
couple of things. i didn't do a test run yet, but i'm excited for this addition 🥳

queue/redis/opts.go
plyr4: yeah i agree its confusing. they really should be called `keys` in the redis client _only_, not routes or channels. so we may as well go with all routes

database/settings/table.go
plyr4: correct, a user cant create the record ever

database/settings/table.go
plyr4: lol again?! ill fix it in vim, thanks

api/admin/settings.go
```diff
@@ -0,0 +1,221 @@
...
+//       items:
+//         "$ref": "#/definitions/Platform"
+//   '500':
+//     description: Unable to retrieve settings from the database
+//     schema:
+//       "$ref": "#/definitions/Error"
+
+// GetSettings represents the API handler to
+// captures settings stored in the database.
+func GetSettings(c *gin.Context) {
```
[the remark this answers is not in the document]
plyr4: nope, ty

api/admin/settings.go
```diff
@@ -0,0 +1,221 @@
...
+// - application/json
+// security:
+//   - ApiKeyAuth: []
+// responses:
+//   '200':
+//     description: Successfully retrieved settings from the database
+//     schema:
+//       type: array
+//       items:
+//         "$ref": "#/definitions/Platform"
```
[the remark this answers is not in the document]
plyr4: good catch ty

api/admin/settings.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	s.SetUpdatedBy(u.GetName())
+
+	// send API call to update the settings
+	s, err = database.FromContext(c).UpdateSettings(ctx, s)
+	if err != nil {
+		retErr := fmt.Errorf("unable to update settings: %w", err)
+
+		util.HandleError(c, http.StatusInternalServerError, retErr)
```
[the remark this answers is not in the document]
plyr4: good catch ty

api/admin/settings.go
```diff
@@ -0,0 +1,221 @@
...
+// - application/json
+// security:
+//   - ApiKeyAuth: []
+// responses:
+//   '200':
+//     description: Successfully restored default settings in the database
+//     schema:
+//       type: array
+//       items:
+//         "$ref": "#/definitions/Platform"
```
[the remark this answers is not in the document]
plyr4: good catch ty

api/types/settings/compiler.go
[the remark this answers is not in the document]
plyr4: golangci had an absolute meltdown yesterday, randomly. 50+ comments of nonsense

database/types/settings.go
plyr4: how about `ps` as shorthand for the moniker "platform settings"

cmd/vela-server/server.go
plyr4: yeah agreed, ill tweak it

database/types/settings.go
plyr4: went with `ps` for PlatformSettings to match the naming with `qs` and `cs` which denote QueueSettings and CompilerSettings.

queue/redis/opts.go
plyr4: done! `channels` are now `routes`

cmd/vela-server/server.go
plyr4: went with this:
```
	ps, err := database.GetSettings(context.Background())
	if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
		return err
	}

	// platform settings record does not exist
	if err != nil {
		logrus.Info("creating initial platform settings")
		......
```

wass3r (commented):
was running it locally and playing with the interval. i set it to one minute. however, any change to settings i would make (i was using clone image in this case) would always immediately apply.

plyr4:
> was running it locally and playing with the interval. i set it to one minute[1](#user-content-fn-1-686535cfb043439d3072d553cf6370a0). however, any change to settings i would make (i was using clone image in this case) would always immediately apply.
> ## Footnotes
> 
>     1. `{"level":"info","msg":"refreshing platform settings every 1m0s","time":"2024-05-10T13:22:27Z"}` [↩](#user-content-fnref-1-686535cfb043439d3072d553cf6370a0)

that makes sense, `UpdateSettings` retrieves and updates the shared pointer. refresh loop is useful for broadcasting changes across multiple server instances. perhaps i could add a note to the logs to make that more clear?

api/admin/settings.go
plyr4: sure i can look into that. it should be added to the server bootup as well to check the environment-provided value also, not just here, right?

wass3r:
> > was running it locally and playing with the interval. i set it to one minute[1](#user-content-fn-1-686535cfb043439d3072d553cf6370a0). however, any change to settings i would make (i was using clone image in this case) would always immediately apply.
> > ## Footnotes
> > ```
> > 1. `{"level":"info","msg":"refreshing platform settings every 1m0s","time":"2024-05-10T13:22:27Z"}` [↩](#user-content-fnref-1-686535cfb043439d3072d553cf6370a0)
> > ```
> 
> that makes sense, `UpdateSettings` retrieves and updates the shared pointer. refresh loop is useful for broadcasting changes across multiple server instances. perhaps i could add a note to the logs to make that more clear?

dang it. i swear i looked there and thought it just updated DB, not the in-memory settings 😲 i will run through it again. sorry.

i think we should make a note somewhere. what if folks want to use the interval as a safety-buffer. you know, to allow time for undoing an accidental change? that wouldn't be possible in this set up. arguably an edge case?

api/admin/settings.go
plyr4: taken care of, see https://github.com/go-vela/server/pull/1110/commits/197e42ff5919269bd98ec8fb21e148d028131cbc

plyr4:
> > > was running it locally and playing with the interval. i set it to one minute[1](#user-content-fn-1-686535cfb043439d3072d553cf6370a0). however, any change to settings i would make (i was using clone image in this case) would always immediately apply.
> > > ## Footnotes
> > > ```
> > > 1. `{"level":"info","msg":"refreshing platform settings every 1m0s","time":"2024-05-10T13:22:27Z"}` [↩](#user-content-fnref-1-686535cfb043439d3072d553cf6370a0)
> > > ```
> > 
> > 
> > that makes sense, `UpdateSettings` retrieves and updates the shared pointer. refresh loop is useful for broadcasting changes across multiple server instances. perhaps i could add a note to the logs to make that more clear?
> 
> dang it. i swear i looked there and thought it just updated DB, not the in-memory settings 😲 i will run through it again. sorry.
> 
> i think we should make a note somewhere. what if folks want to use the interval as a safety-buffer. you know, to allow time for undoing an accidental change? that wouldn't be possible in this set up. arguably an edge case?

seems like an edge case but i can totally understand the confusion. i mean it would be extremely easy to just use a different pointer and let the loop refresh it in the background, i suppose at least it would make things consistent/more straightforward. ill make the tweak

wass3r:
> > > > was running it locally and playing with the interval. i set it to one minute[1](#user-content-fn-1-686535cfb043439d3072d553cf6370a0). however, any change to settings i would make (i was using clone image in this case) would always immediately apply.
> 
> > > > ## Footnotes
> 
> > > > ```
> 
> > > > 1. `{"level":"info","msg":"refreshing platform settings every 1m0s","time":"2024-05-10T13:22:27Z"}` [↩](#user-content-fnref-1-686535cfb043439d3072d553cf6370a0)
> 
> > > > ```
> 
> > > 
> 
> > > 
> 
> > > that makes sense, `UpdateSettings` retrieves and updates the shared pointer. refresh loop is useful for broadcasting changes across multiple server instances. perhaps i could add a note to the logs to make that more clear?
> 
> > 
> 
> > dang it. i swear i looked there and thought it just updated DB, not the in-memory settings 😲 i will run through it again. sorry.
> 
> > 
> 
> > i think we should make a note somewhere. what if folks want to use the interval as a safety-buffer. you know, to allow time for undoing an accidental change? that wouldn't be possible in this set up. arguably an edge case?
> 
> 
> 
> seems like an edge case but i can totally understand the confusion. i mean it would be extremely easy to just use a different pointer and let the loop refresh it in the background, i suppose at least it would make things consistent/more straightforward. ill make the tweak

Yes, I guess the more impactful scenario is that the bigger the interval the longer you will see inconsistent behavior because only one server will have the update. Thanks!

api/admin/settings.go
plyr4: lol..... okay this actually turned into a bigger effort because of this test:
https://github.com/go-vela/server/blob/8a744e48d224197478bb48a468186783b04c44db/compiler/native/compile_test.go#L1852

that is failing because there is no clone image provided, NOT because the yaml has no steps or stages.
the steps validation injects clone+init and doesnt actually fail but the test _thinks_ its failing, its actually failing because theres no clone image, so when i provide a clone image to get a valid compiler the test starts failing. 
its a not-my-problem-at-all scope creep where i need to now fix steps validation....
i dont feel comfortable changing compiler yaml validation in this PR

plyr4: im thinking of how i can get around this but now i know this test is a false positive so im sorta stuck.

fix now? no, not touching compiler functionality in this PR.
skip the test? gross, we might not ever get around to actually fixing it.
remove docker image validation for the compiler? i want to keep it.

ill think about it over lunch

plyr4: fwiw this lets us move on, but it hides the underlying issue. an issue that isnt at-all related to this PR
https://github.com/go-vela/server/pull/1110/commits/a43546e83ddd47c1da398c5e40967eea3d4cde9f#diff-a25c05febe73513fec1340739f5f59b78fb50a646dccf716b5825cd5fe3d8652R1849-R1851

plyr4:
> > > > > was running it locally and playing with the interval. i set it to one minute[1](#user-content-fn-1-686535cfb043439d3072d553cf6370a0). however, any change to settings i would make (i was using clone image in this case) would always immediately apply.
> > 
> > 
> > > > > ## Footnotes
> > 
> > 
> > > > >
> > 
> > 
> > > > > 1. `{"level":"info","msg":"refreshing platform settings every 1m0s","time":"2024-05-10T13:22:27Z"}` [↩](#user-content-fnref-1-686535cfb043439d3072d553cf6370a0)
> > 
> > 
> > > > >
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > that makes sense, `UpdateSettings` retrieves and updates the shared pointer. refresh loop is useful for broadcasting changes across multiple server instances. perhaps i could add a note to the logs to make that more clear?
> > 
> > 
> > > 
> > 
> > 
> > > dang it. i swear i looked there and thought it just updated DB, not the in-memory settings 😲 i will run through it again. sorry.
> > 
> > 
> > > 
> > 
> > 
> > > i think we should make a note somewhere. what if folks want to use the interval as a safety-buffer. you know, to allow time for undoing an accidental change? that wouldn't be possible in this set up. arguably an edge case?
> > 
> > 
> > seems like an edge case but i can totally understand the confusion. i mean it would be extremely easy to just use a different pointer and let the loop refresh it in the background, i suppose at least it would make things consistent/more straightforward. ill make the tweak
> 
> Yes, I guess the more impactful scenario is that the bigger the interval the longer you will see inconsistent behavior because only one server will have the update (which makes a perhaps unreasonable assumptions that all servers start at the same time). Thanks! But maybe I'm making an elephant out of a mosquito here 😅

heres the update: https://github.com/go-vela/server/pull/1110/commits/4cad3e66a5fe1c6bd5d5b145a9dc295b624547ea

i'll let you make the call on keeping it or reverting it. i dont mind either way.

api/admin/settings.go
wass3r: ah, interesting. i like this added validation, but i was thinking more along the lines of https://pkg.go.dev/github.com/docker/docker/client#Client.DistributionInspect . i could have explained it better. thanks for adding this.

wass3r:
> > > > > > was running it locally and playing with the interval. i set it to one minute[1](#user-content-fn-1-686535cfb043439d3072d553cf6370a0). however, any change to settings i would make (i was using clone image in this case) would always immediately apply.
> > > 
> > > 
> > > > > > ## Footnotes
> > > 
> > > 
> > > > > >
> > > 
> > > 
> > > > > > 1. `{"level":"info","msg":"refreshing platform settings every 1m0s","time":"2024-05-10T13:22:27Z"}` [↩](#user-content-fnref-1-686535cfb043439d3072d553cf6370a0)
> > > 
> > > 
> > > > > >
> > > 
> > > 
> > > > >
> > > 
> > > 
> > > > >
> > > 
> > > 
> > > > > that makes sense, `UpdateSettings` retrieves and updates the shared pointer. refresh loop is useful for broadcasting changes across multiple server instances. perhaps i could add a note to the logs to make that more clear?
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > dang it. i swear i looked there and thought it just updated DB, not the in-memory settings 😲 i will run through it again. sorry.
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > i think we should make a note somewhere. what if folks want to use the interval as a safety-buffer. you know, to allow time for undoing an accidental change? that wouldn't be possible in this set up. arguably an edge case?
> > > 
> > > 
> > > seems like an edge case but i can totally understand the confusion. i mean it would be extremely easy to just use a different pointer and let the loop refresh it in the background, i suppose at least it would make things consistent/more straightforward. ill make the tweak
> > 
> > 
> > Yes, I guess the more impactful scenario is that the bigger the interval the longer you will see inconsistent behavior because only one server will have the update (which makes a perhaps unreasonable assumptions that all servers start at the same time). Thanks! But maybe I'm making an elephant out of a mosquito here 😅
> 
> heres the update: [4cad3e6](https://github.com/go-vela/server/commit/4cad3e66a5fe1c6bd5d5b145a9dc295b624547ea)
> 
> i'll let you make the call on keeping it or reverting it. i dont mind either way.

i like the change. i think that is expected behaviour in a single instance scenario and it gives you the option of close consistency in multi-server scenario. thank you!

api/admin/settings.go
plyr4: oh well instead of using DistributionInspect, i imported distribution and inspected it! lol 

to be fair i yoinked that entire helper function from go-vela/worker, and i kind of like that its tied to the general container spec and not docker

wass3r: true. i was thinking of actually pulling info from the registry (which is what [DistributionInspect](https://pkg.go.dev/github.com/docker/docker/client#Client.DistributionInspect) does vs a image ref format validation. we can add it later if we think that would be useful.

wass3r (approved):
nice work 😍

database/types/settings.go
```diff
@@ -0,0 +1,218 @@
...
+	if ps.UpdatedAt.Int64 < 0 {
+		return fmt.Errorf("updated_at must be greater than zero, got: %d", ps.UpdatedAt.Int64)
+	}
+
+	return nil
+}
+
+// FromAPI converts the API Settings type
+// to a database Settings type.
+func FromAPI(s *settings.Platform) *Platform {
```
ecrupper, line 199:
```suggestion
func SettingsFromAPI(s *settings.Platform) *Platform {
```
Sorry just caught this. Need to have the resource part of the name if they are all in the `types` package.

ecrupper (changes requested):

database/types/settings.go
plyr4: cool cool lemme tweak it

ecrupper: Will address in follow up PR

ecrupper (approved):

database/types/settings.go
plyr4: https://github.com/go-vela/server/pull/1126 🙇

PR #1128, opened 2024-05-13
chore: Correct the register worker token swagger path

tmapes:
Noticed in the [admin router](https://github.com/go-vela/server/blob/v0.23.4/router/admin.go#L61) that this endpoint is wired up  under `[...]/workers/:worker/register` but it's currently documented as `[...]/workers/:worker/register-token`


plyr4 (approved):
ah nice catch, thank you!

ecrupper (approved):

PR #1130, opened 2024-05-13
fix(settings): missing fields in api conversion and func renames

plyr4:
created_at was returning 0, only if the value was updated via API. 

also, noticed a variable type `newSettingps` so i made it the ol' shadow variable with an underscore


wass3rw3rk (approved):

api/types/settings/platform.go
```diff
@@ -230,19 +230,23 @@ func (ps *Platform) SetUpdatedBy(v string) {
...
-	ps.SetCompiler(newSettingps.GetCompiler())
-	ps.SetQueue(newSettingps.GetQueue())
-	ps.SetRepoAllowlist(newSettingps.GetRepoAllowlist())
-	ps.SetScheduleAllowlist(newSettingps.GetScheduleAllowlist())
+	ps.SetCompiler(_ps.GetCompiler())
+	ps.SetQueue(_ps.GetQueue())
+	ps.SetRepoAllowlist(_ps.GetRepoAllowlist())
+	ps.SetScheduleAllowlist(_ps.GetScheduleAllowlist())
+
+	ps.SetCreatedAt(_ps.GetCreatedAt())
```
wass3rw3rk, line 247: `created_at` shouldn't be updateable, right?

wass3rw3rk (changes requested):

api/types/settings/platform.go
plyr4: no but that raises a good point this function should really be called `DuplicateInPlace` or something like that. i see the confusion

plyr4: or... `Overwrite` or `FromSettings` ?

fwiw its used in these places:
https://github.com/go-vela/server/blob/main/cmd/vela-server/server.go#L255
https://github.com/go-vela/server/blob/main/api/admin/settings.go#L121-L122

plyr4: i made some tweaks and renamed `Update` to `FromSettings`

plyr4:
added a `settings.FromCLIContext` that fixes missing fields in `Restore`

ecrupper (approved):

wass3rw3rk (approved):

PR #1133, opened 2024-06-03
enhance: add scm user id to build obj

plyr4:
attaching the user's scm id to builds.

required for build identification via OIDC, see: #1120


ecrupper (dismissed):

api/webhook/post.go
```diff
@@ -207,6 +207,23 @@ func PostWebhook(c *gin.Context) {
...
+			util.HandleError(c, http.StatusBadRequest, retErr)
+
+			h.SetStatus(constants.StatusFailure)
+			h.SetError(retErr.Error())
+
+			return
+		}
+
+		b.SetSenderSCMID(senderID)
+	}
```
wass3rw3rk, line 225: in what circumstances will this get hit? or, when is there not a SCMID? Just trying to gauge overhead.

plyr4: this might happen when a git webhook payload doesnt contain a `sender` object and we fallback to setting sender using a field that contains username but no ID, for example `pusher` in this piece of code has no `id`

here: https://github.com/go-vela/server/blob/main/scm/github/webhook.go#L179

though, when can _that_ scenario occur? i dont know, but the code is there, so it implies that its possible for sender to be empty. in this case the sender_scm_id from `GetID()` above will return `0`, hence the `== '0'`.

ill add some comments real quick

wass3rw3rk: yup, sounds good. it didn't seem like it would get hit a lot. thanks for the clarification.

wass3rw3rk (approved):

ecrupper (approved):

wass3rw3rk (approved):

PR #1135, opened 2024-06-05
feat(db/log): use logrus as logger for db

wass3r:
db logs are not using logrus format currently. this change ensure database logs are formatted in same manner as other logs. some highlights:
- uses logrus as logger with same format as global logger (json, ecs, etc)
- adds knobs to control db logging:
  - `VELA_DATABASE_LOG_LEVEL` - "warn" is default, pretty noisy otherwise
  - `VELA_DATABASE_LOG_SLOW_THRESHOLD` - 200ms default, inherited from gorm default
  - `VELA_DATABASE_LOG_SHOW_SQL` - false is default
  - `VELA_DATABASE_LOG_SKIP_NOTFOUND` - true is default
- will create a copy of global logger and configure custom level if the log level is different than global logger


KellyMerrick (dismissed):
lgtm

ecrupper (approved):

KellyMerrick (approved):
lgtm

PR #1139, opened 2024-06-07
fix(docs/swagger): various fixes and improved consistency

wass3r:
well, well.. started with fixing some errors i noticed in the swagger annotations and quickly turned into a rabbit hole of fixing and trying to improve various things.

however, here are some highlights:
- correct flat out errors in swagger annotations
- ensure all response codes are covered
- simplify language (also meant removing some unnecessary language)
- improve consistency in language

Note: only on the admin endpoints did i correct some 404 returns that should have been 400, otherwise there are no actual code changes.

there is more to be done. i noticed some inconsistencies in some places where we return 404 when there was an error retrieving objects from the database when in most cases we return internal server error (500). might want to consider determining what kind of error and returning appropriate response.

### DIFF
--- a/api/admin/build.go
+++ b/api/admin/build.go
@@ -19,7 +19,7 @@ import (
 
 // swagger:operation GET /api/v1/admin/builds/queue admin AllBuildsQueue
 //
-// Get all of the running and pending builds in the database
+// Get running and pending builds
 //
 // ---
 // produces:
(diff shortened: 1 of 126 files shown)


ecrupper (approved):
Nice thanks for doing this

plyr4 (approved):

PR #1141, opened 2024-06-08
fix(deps): switch go-jose lib

wass3r:
closes https://github.com/go-vela/server/security/dependabot/17 although only used (accidentally?) in tests in our case. we do still have an indirect dependency on the good version of the library.


ecrupper (approved):

plyr4 (dismissed):

ecrupper (approved):

KellyMerrick (approved):
lgtm

PR #1145, opened 2024-06-17
fix: logging

wass3rw3rk:
PR is focused on various logging fixes and accomplishes the following:
1. Injects a logger into the context via existing logging middleware in `router/middleware/logger.go`
1. Adds more fields to the request logger (in the logging middleware), such as IDs for resources 
1. Other middleware builds up fields on the injected logger, for example: “build” middleware adds build number and id to the logger and updates the logger in the context - this is to avoid repetition in the handlers where we would have to manually add each. 
1. Request logger is indicator for final state of resource. for example, a 200 on a DELETE for a resource indicates successful deletion, it won’t be logged outside the request logger, but should have all the augmented metadata for a richer log entry. one exception is on creation for resources, since middleware can’t populate something that doesn’t exist, we populate the logger with appropriate fields once the resource has been created and emit a extra log, so you will see a request logger entry (201 status code) and a separate entry that a resource was created with associated ID and such.
1. Logs creation and modification of resources, including IP, path, and other metadata (see previous points)
1. Logs read access to sensitive resources, such as secrets
1. More consistency: always lower case, debug level when entering handler, info/error *after* a resource was created/changed (unsuccessfully), etc
1. Additional debug logging in main webhook handler
1. Adds hook middleware

Note: this PR also addresses some swagger annotations issues discovered (missed in https://github.com/go-vela/server/pull/1139)


api/admin/repo.go
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
-//nolint:dupl // ignore similar code
+//nolint:dupl // ignore similar code with service.go
```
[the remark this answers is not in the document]
wass3rw3rk: addressed

api/admin/build.go
```diff
@@ -103,10 +104,11 @@ func AllBuildsQueue(c *gin.Context) {
 
 // UpdateBuild represents the API handler to update a build.
 func UpdateBuild(c *gin.Context) {
-	logrus.Info("Admin: updating build in database")
-
 	// capture middleware values
 	ctx := c.Request.Context()
+	l := c.MustGet("logger").(*logrus.Entry)
+
+	l.Debug("platform admin: updating build")
```
plyr4, line 111: consistency nitpick, some funcs had the logging moved to the top of the function (`AllBuildsQueue`) and some were moved below middleware values

plyr4: fwiw i prefer it being at the very top if it doesnt require anything from the middleware

api/admin/user.go
```diff
@@ -78,5 +84,10 @@ func UpdateUser(c *gin.Context) {
 		return
 	}
 
-	c.JSON(http.StatusOK, u)
+	l.WithFields(logrus.Fields{
+		"target_user_id": tu.GetID(),
```
plyr4, line 88: i like the specificity

api/build/approve.go
```diff
@@ -113,9 +103,11 @@ func ApproveBuild(c *gin.Context) {
 	// update the build in the db
 	_, err := database.FromContext(c).UpdateBuild(ctx, b)
 	if err != nil {
-		logrus.Errorf("Failed to update build %d during publish to queue for %s: %v", b.GetNumber(), r.GetFullName(), err)
+		l.Errorf("failed to update build during publish to queue: %v", err)
```
plyr4, line 106: ❤️ no more upper case errors ❤️

api/build/clean.go
```diff
@@ -19,6 +19,16 @@ import (
 // without execution. This will kill all resources,
 // like steps and services, for the build.
 func CleanBuild(ctx context.Context, database database.Interface, b *types.Build, services []*library.Service, steps []*library.Step, e error) {
+	logger := logrus.WithFields(logrus.Fields{
```
plyr4, line 22: no gin/middleware for this one? 
```
	l := c.MustGet("logger").(*logrus.Entry)
```

plyr4: should we call it `l` at least to be consistent

api/build/compile_publish.go
```diff
@@ -49,7 +49,15 @@ func CompileAndPublish(
 	compiler compiler.Engine,
 	queue queue.Service,
 ) (*pipeline.Build, *models.Item, int, error) {
-	logrus.Debugf("generating queue items for build %s/%d", cfg.Build.GetRepo().GetFullName(), cfg.Build.GetNumber())
+	logger := logrus.WithFields(logrus.Fields{
```
plyr4, line 52: same question, why no `c.MustGet` and `l`

api/build/enqueue.go
```diff
@@ -16,28 +16,36 @@ import (
...
-	logrus.Infof("Converting queue item to json for build %d for %s", item.Build.GetNumber(), item.Build.GetRepo().GetFullName())
+	logger := logrus.WithFields(logrus.Fields{
+		"build":    item.Build.GetNumber(),
+		"build_id": item.Build.GetID(),
+		"org":      item.Build.GetRepo().GetOrg(),
+		"repo":     item.Build.GetRepo().GetName(),
+		"repo_id":  item.Build.GetRepo().GetID(),
+	})
+
+	logger.Debug("converting queue item to json")
```
plyr4, line 27: hard to fix all of these, but maybe some specificity here

```suggestion
	logger.Debugf("enqueuing build item to route %s", route)
```

api/build/enqueue.go
```diff
@@ -16,28 +16,36 @@ import (
...
+		logger.Errorf("failed to convert item to json: %v", err)
 
 		// error out the build
 		CleanBuild(ctx, db, item.Build, nil, nil, err)
 
 		return
 	}
 
-	logrus.Infof("Pushing item for build %d for %s to queue route %s", item.Build.GetNumber(), item.Build.GetRepo().GetFullName(), route)
+	logger.Debugf("pushing item for build to queue route %#q", route)
```
plyr4, line 39: whats the reasoning for the raw string literal`%#q` ? just curious

api/build/enqueue.go
plyr4: oh, i see, this log line is a bit lower. still, feels different from how the other funcs do it. 🤷

api/build/id_token.go
```diff
@@ -75,19 +75,12 @@ import (
...
-	// update engine logger with API metadata
-	//
-	// https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Entry.WithFields
-	logrus.WithFields(logrus.Fields{
-		"build":   b.GetNumber(),
-		"org":     b.GetRepo().GetOrg(),
-		"repo":    b.GetRepo().GetName(),
-		"subject": cl.Subject,
-	}).Infof("generating ID token for build %s/%d", b.GetRepo().GetFullName(), b.GetNumber())
+	l.Infof("generating ID token for build %s/%d", b.GetRepo().GetFullName(), b.GetNumber())
```
plyr4, line 83: btw not sure if it was intended but i really like anything token/user/auth related being `info`

api/dashboard/delete.go
```diff
@@ -56,16 +56,11 @@ import (
...
 	u := user.Retrieve(c)
 
-	// update engine logger with API metadata
-	//
-	// https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Entry.WithFields
-	logrus.WithFields(logrus.Fields{
-		"dashboard": d.GetID(),
-		"user":      u.GetName(),
-	}).Infof("deleting dashboard %s", d.GetID())
+	l.Debugf("deleting dashboard %s", d.GetID())
```
plyr4, line 63: wondering the thought around using fields vs not using fields. just curious

api/secret/create.go
```diff
@@ -259,5 +255,14 @@ func CreateSecret(c *gin.Context) {
 		return
 	}
 
+	l.WithFields(logrus.Fields{
+		"secret_id":   s.GetID(),
+		"secret_name": s.GetName(),
+		"secret_org":  s.GetOrg(),
+		"secret_repo": s.GetRepo(),
+		"secret_type": s.GetType(),
+		"secret_team": s.GetTeam(),
```
plyr4, line 264: do we need to check type and `delete` team like the other funcs? doesnt seem necessary but nice to have 🤷

api/webhook/post.go
```diff
@@ -82,12 +82,13 @@ var baseErr = "unable to process webhook"
...
 //nolint:funlen,gocyclo // ignore function length and cyclomatic complexity
 func PostWebhook(c *gin.Context) {
-	logrus.Info("webhook received")
-
 	// capture middleware values
 	m := c.MustGet("metadata").(*internal.Metadata)
+	l := c.MustGet("logger").(*logrus.Entry)
 	ctx := c.Request.Context()
 
+	l.Info("webhook received")
```
plyr4, line 90: considering the noise, maybe `Debug` level? or instead we add some fields so its helpful

api/webhook/post.go
```diff
@@ -405,6 +429,14 @@ func PostWebhook(c *gin.Context) {
 
 				return
 			}
+
+			l.WithFields(logrus.Fields{
+				"deployment":    d.GetNumber(),
```
plyr4, line 434: could do the same and set/reference `dr` from `UpdateDeployment` but not necessary

api/webhook/post.go
```diff
@@ -191,8 +192,16 @@ func PostWebhook(c *gin.Context) {
...
+			l.Errorf("unable to update webhook %s/%d: %v", r.GetFullName(), h.GetNumber(), err)
 		}
+
+		l.WithFields(logrus.Fields{
+			"hook":    h.GetNumber(),
+			"hook_id": h.GetID(),
+			"org":     r.GetOrg(),
+			"repo":    r.GetName(),
+			"repo_id": r.GetID(),
+		}).Info("hook updated in database")
```
plyr4, line 204: thoughts on getting rid of "in database" 
im feeding off your annotations fixes here https://github.com/go-vela/server/pull/1139
i really liked the removal of the redundant language for "in the configured backend" and "in the database"

api/webhook/post.go
```diff
@@ -381,6 +397,14 @@ func PostWebhook(c *gin.Context) {
...
 					return
 				}
+
+				l.WithFields(logrus.Fields{
+					"deployment":    dr.GetNumber(),
+					"deployment_id": dr.GetID(),
+					"org":           repo.GetOrg(),
+					"repo":          repo.GetName(),
+					"repo_id":       repo.GetID(),
+				}).Info("deployment created in database")
```
plyr4, line 407: same question, redundant language

api/dashboard/delete.go
plyr4: oh! some of the resource logger fields are set in the middleware `Establish`/`Retrieve` functions, i see that now. sweet!

api/build/id_token.go
wass3rw3rk: agree. i'll double check to see if that's consistent.

api/build/enqueue.go
wass3rw3rk: i was starting to play with introducing this pattern. it's a bit nicer than straight up `%s`. you can avoid things like `queue route route` (if the route is named `route`) or `queue route ` (if the var is empty). it would be `` queue route `route` `` and `` queue route `​` `` respectively. it also, of course, safely escapes the string. i would like to use it over `%s`. 

i reverted my changes for that (but missed that spot), to maybe introduce later. i will fix this instance up.

api/webhook/post.go
wass3rw3rk: should be addressed in https://github.com/go-vela/server/pull/1145/commits/53edfd650dbff8ab902ed3490e4fc55bd36ee151

api/webhook/post.go
wass3rw3rk: should be addressed in https://github.com/go-vela/server/pull/1145/commits/53edfd650dbff8ab902ed3490e4fc55bd36ee151

api/build/enqueue.go
wass3rw3rk: i reverted this

api/build/clean.go
wass3rw3rk: i might have to throw in a check to see if ctx is gin or not, then we could. i just checked one path and it was, but the signature made me think it was not always.

api/webhook/post.go
wass3r: yea, my thought was that if the update was successful, i don't need the new object. i think i used that approach consistently, but if not, i will update in those spots.

api/webhook/post.go
wass3r: i was going back and forth on this one initially, but i think the request logger is good enough for getting request volume on the endpoint. changed to debug in https://github.com/go-vela/server/pull/1145/commits/2c8afeee31314c8095f1782a3b31f31394464f2d.

api/admin/build.go
wass3r: found a couple of spots. fixed in https://github.com/go-vela/server/pull/1145/commits/2508d5f5fd067e9b3510914c27d4322e20ca3e2e

api/secret/create.go
wass3r: fixed in https://github.com/go-vela/server/pull/1145/commits/76997452b4a41417c00847b7b485fc1d0df3bda9

ecrupper (approved):

api/build/clean.go
wass3rw3rk: yea, this is not always gin.Context, so it won't have the logger in it in certain scenarios. opted to just rename to `l` for now. lmk, if you think it's worth doing a type check on context and pulling the logger out if we have it. https://github.com/go-vela/server/pull/1145/commits/924390dfb13b70146a4ddcb0fda4abc9bbe88f16

api/build/compile_publish.go
wass3rw3rk: yea, this is not always gin.Context, so it won't have the logger in it in certain scenarios. opted to just rename to `l` for now. lmk, if you think it's worth doing a type check on context and pulling the logger out if we have it. https://github.com/go-vela/server/pull/1145/commits/924390dfb13b70146a4ddcb0fda4abc9bbe88f16

api/build/id_token.go
wass3rw3rk: some changes in https://github.com/go-vela/server/pull/1145/commits/e1c4e37bb01afccec49b65490641faa9917f86ae - i'm actually a little torn on it. we get failed/success states from the request logger already, this will be fairly noisy. thoughts?

ecrupper (approved):

plyr4 (approved):
sweet, thanks for the followups

PR #1156, opened 2024-07-08
fix(oidc): add expected fields

JordanSussman:
I spent some time over this weekend experimenting with the new OIDC feature and noticed that the `/_services/token/.well-known/openid-configuration` endpoint doesn't align with AWS's expected format for openid. AWS's documentation specifies the required fields, which you can find at https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc.html#manage-oidc-provider-prerequisites. After updating the API response to include all of these fields, the issue was resolved. 

openid documents the discovery endpoint schema within https://openid.net/specs/openid-connect-discovery-1_0.html.


ecrupper (approved):
Thanks lgtm

PR #1159, opened 2024-07-14
fix(swagger/oidc): fix swagger and oidc docs

wass3r:
depends on https://github.com/go-vela/types/pull/385

includes the following changes:
- oidc and jwks endpoints don't need auth (fixed in swagger annotation)
- fix swagger doc creation
  - this must have been broken for a while, it doesn't actually generate types, only references with `x-go-package` fields. apparently it's a bug with the release binaries(?!). using the `go install` variant of installing go-swagger the spec is produced properly with the same command (see https://github.com/go-swagger/go-swagger/issues/3107)
- after the fixing the previous, i noticed that JWKSet was being documented as producing a Len and Clear field?! I changed up the type to match the response properly. maybe there's a better way yet.


wass3r:
I noticed it's still not quite right because now we have swagger model definitions in two places due to the types transition and it's still favoring go-vela/types :/

plyr4:
so we're looking to drop usage of `"github.com/lestrrat-go/jwx/v2/jwk"` for the `jwk.Set` ? 
it was kind of nice to not have to manage that type

wass3r:
@plyr4 without this change, JWKSet is being documented as the following via go-swagger:
```
    "JWKSet": {
      "type": "object",
      "title": "JWKSet is a wrapper of lestrrat-go/jwx/jwk.Set for API Swagger gen.",
      "properties": {
        "Clear": {
          "description": "Clear resets the list of keys associated with this set, emptying the\ninternal list of `jwk.Key`s, as well as clearing any other non-key\nfields",
          "type": "string"
        },
        "Len": {
          "description": "Len returns the number of keys in the set",
          "type": "integer",
          "format": "int64"
        }
      },
      "x-go-package": "github.com/go-vela/server/api/types"
    },
```

if we want to properly document what this produces in the swagger spec, we have to maintain something manually somewhere.

ideally, there is either a change in go-swagger or jwkset that would allow proper automatic documentation.

open to better ideas here, if there are any.

plyr4 (approved):
sure, if it still works for key discovery then it works for me!

ecrupper (approved):

PR #1160, opened 2024-07-15
enhance(oidc): provide support for custom issuer

ecrupper:
This change will allow platform admins to publish the OpenID configuration (and in turn the JWKS) to custom URLs. 

Of course, in order to have accurate information, these custom URLs will need to source their data from the standard Vela routes. But this allows selective exposure of OpenID data to external services such as GCP and AWS.


plyr4:
sweet, i think we also need it [here](https://github.com/go-vela/server/blob/main/compiler/native/environment.go#L318)?

plyr4:
also should we add url validation for `oidc-issuer`?
like we do for `server-addr`
```
addr, err := url.Parse(c.String("server-addr"))
if err != nil {
	return err
}
```

plyr4 (approved):

KellyMerrick (approved):
lgtm

PR #1161, opened 2024-07-17
fix(api): add ref to dashboard build partial

KellyMerrick:
Necessary so PR # can render in the UI for dashboard recent builds.

image


timhuynh94 (approved):

plyr4:
we need the mocks updated too 
https://github.com/go-vela/server/blob/main/mock/server/dashboard.go#L88

timhuynh94 (approved):

PR #1162, opened 2024-07-18
fix(webhook): use correct repo variable to populate client credentials

ecrupper:
`r` is what we get from parsing SCM webhook; `repo` is what we have in the database. Need the latter to gather proper credentials for contributor call.


plyr4 (approved):

KellyMerrick (approved):
lgtm

PR #1168, opened 2024-08-13
enhance: supply parent context to all db calls

plyr4:
required for grouping otel gorm spans into the parent trace, see: #899


ecrupper (approved):
ecrupper.WithContext(ctx).approves()

KellyMerrick (approved):
lgtm

PR #1171, opened 2024-08-26
fix: upgrade yaml library to go-yaml v3 over buildkite

ecrupper:
Follow up from https://github.com/go-vela/types/pull/386

And also fixing the tests that were broken from https://github.com/go-vela/types/pull/387


KellyMerrick (approved):
lgtm

PR #1172, opened 2024-08-26
enhance(oidc): add more claims

JordanSussman:
closes https://github.com/go-vela/community/issues/996

Note: I intended to include whether the build was triggered from a fork, but it seems we don’t currently store that information in the database. Therefore, I decided to omit that detail for now. If someone wants to add this feature in the future, they will also need to update the database to capture this data.


mock/server/authentication.go
```diff
@@ -40,6 +41,8 @@ const (
     "actor_scm_id",
     "commands",
     "image",
+	"image_name",
+	"image_tag",
```
wass3r, line 45: looks like these are not indented right?
image

wass3r (commented):
one minor thing, otherwise lgtm

ecrupper:
Is there an intention to follow up this PR with the API implementation [here](https://github.com/go-vela/server/blob/0894432bab40b5b6981007233aca0f7d093194b1/api/build/id_token.go#L76-L97)?

mock/server/authentication.go
JordanSussman: This should be resolved now.

JordanSussman:
> Is there an intention to follow up this PR with the API implementation [here](https://github.com/go-vela/server/blob/0894432bab40b5b6981007233aca0f7d093194b1/api/build/id_token.go#L76-L97)?

Can you expand on this a bit more? The code I added should be returned in the API payload due to https://github.com/go-vela/server/blob/0894432bab40b5b6981007233aca0f7d093194b1/api/build/id_token.go#L121.

ecrupper:
> > Is there an intention to follow up this PR with the API implementation [here](https://github.com/go-vela/server/blob/0894432bab40b5b6981007233aca0f7d093194b1/api/build/id_token.go#L76-L97)?
> 
> Can you expand on this a bit more? The code I added should be returned in the API payload due to
> 
> https://github.com/go-vela/server/blob/0894432bab40b5b6981007233aca0f7d093194b1/api/build/id_token.go#L121
> 
> .

Whoops, got confused and thought `image` was net new. Yep lgtm

ecrupper (approved):

wass3r (approved):

plyr4 (approved):

PR #1175, opened 2024-09-04
fix: webhook race condition on hook creation

jbrockopp:
xref: https://github.com/go-vela/community/issues/997

Currently, there is an inconvenient bug that our users experience when Vela processes a webhook:

```
{"error":"unable to create webhook OrgName/RepoName/BuildNumber: ERROR: duplicate key value violates unique constraint \"hooks_repo_id_number_key\" (SQLSTATE 23505)"}
```

This errors happens when we attempt to create a hook in the database that contains the same `number` and `repo_id`.

This attempts to resolve that issue by creating a retry loop with backoff that will repeat the following operations:

* capture the last hook for the repo
* set fields on the new hook object to be created in the database
* create the webhook object in the database

## Example

For the sake of discussion, I find examples helpful so let's say we have a repo where the last hook number is `10`.

Using the example provided in the ticket, let's say we have two `push` events fire after a PR is merged:

* a `push` to the target branch for the PR
* a `push` for the deletion event of the source branch from the PR

In this scenario, the Vela server would process each event concurrently and attempt to insert a `11` hook record but one of these webhooks will fail to be processed due to the conflict as displayed above. This resolves that issue because in that scenario, we'd fall back to looking up the last hook for the repo (`11`) and updating the number to `12`.


JordanSussman (approved):

plyr4 (approved):
sweet!

ecrupper (approved):

PR #1178, opened 2024-09-11
enhance: add a status call for failed compilation

ecrupper:
This will greatly ease the amount of sleuthing users have to do when they don't see a Vela status show up on their PR / commit. 

Further, the context URL will point to the hooks page, which should take users right to the error message.


KellyMerrick (approved):
lgtm

PR #1182, opened 2024-09-13
fix(dasboards): remove deleted dashboards from users list

KellyMerrick:
Dashboard ID's for deleted dashboards need to also be deleted from a user's dashboard list, as they would continue to count towards a user's total dashboard count otherwise.


ecrupper (commented):
Question: do you think it's a better idea to make the user updates _after_ the deletion of the dashboard? Say the DB call fails for whatever reason, it may be a better experience for the dashboard reference to still exist on the user object if the delete fails.

ecrupper (approved):

wass3rw3rk (approved):

PR #1192, opened 2024-09-20
refactor(db/get-repo): utilize db indexing

wass3rw3rk:
we index `full_name` (https://github.com/go-vela/server/blob/59553f5f606b275701132b0f5edfa2b725935b3e/database/repo/table.go#L36-L40), but are not leveraging it.


wass3rw3rk:
redacting, because we do index org and name :shame: https://github.com/go-vela/server/blob/59553f5f606b275701132b0f5edfa2b725935b3e/database/repo/index.go#L10-L15

PR #1197, opened 2024-09-30
refactor(deployment): use server API types for deployment

ecrupper:
Continuing the efforts to migrate from the types repo and also augment API responses.

This should be the last resource that we augment. The rest will stay slim (step, log, etc.)


wass3rw3rk (approved):

plyr4 (approved):

timhuynh94 (approved):

wass3rw3rk:
i would like to see a PR on go-vela/types to annotate the legacy Hook type as deprecated.

PR #1200, opened 2024-10-03
refactor(pipeline): use server API types for pipeline and migrate compiler types

ecrupper:
- nested API type for pipeline 
- preloading only for DB functions where the related resource is not already present in the function arguments
- moving all the compiler types over (`yaml`, `pipeline`, `raw`)


wass3rw3rk (approved):

KellyMerrick (approved):
lgtm

PR #1204, opened 2024-10-04
refactor: required changes from api build types migration

plyr4:
need this to let the worker build in https://github.com/go-vela/worker/pull/608

this is subject to change when pipeline and step are migrated, but until then we need it to build the worker

```bash
cd go-vela/worker

make build

executor/local/step.go:53:50: c.build.ToLibrary undefined (type *"github.com/go-vela/server/api/types".Build has no field or method ToLibrary)
```

### DIFF
--- a/api/types/build.go
+++ b/api/types/build.go
@@ -8,6 +8,8 @@ import (
 	"time"
 
 	"github.com/go-vela/types/constants"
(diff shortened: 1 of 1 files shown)


wass3rw3rk:
not needed anymore

PR #1207, opened 2024-10-15
fix(oidc): image parse sha support

plyr4:
tiny fix for https://github.com/go-vela/server/pull/1172 `invalid reference format` with sha tags

or should we use `reference.Parse` similar to the [worker code](https://github.com/go-vela/worker/blob/main/internal/image/image.go#L61-L75)


ecrupper (approved):
TODO may be to actually parse the SHA and make it a claim. Not all SHA digests are going to refer to `latest`

JordanSussman (approved):

PR #1210, opened 2024-10-18
refactor(steps/services): use server API types for steps and services

ecrupper:
Also added missing constants. All that remain after this is `log`, `build_executable`, and a few other stray types from outside the `library` and `database` packages.


timhuynh94 (approved):

wass3r (approved):

KellyMerrick (approved):
lgtm

PR #1214, opened 2024-10-24
feat: deployment config for expected params and targets

ecrupper:
Additional context: https://github.com/go-vela/community/issues/1002

This change will allow for the `deployment` block in Vela pipeline files:
```yaml
version: "1"

deployment:
    targets: [ dev, stage, prod ]
    parameters:
        foo:
            description: blah
            required: true
            type: string
```

These configurations can also be templated using the standard template format already existing today in Vela:

```yaml
version: "1"

templates:
    - name: my-deploy-config
       source: over-there.yml
       type: file

deployment:
    template:
        name: my-deploy-config
        vars:
            regions: [ north, east, west ]
```

Deployment configurations can also be requested using the new API endpoint:

`GET /api/v1/deployments/:org/:repo/config`

This will allow for the CLI and UI to validate deployments prior to their launch. However, because you can create deployments in other ways outside the scope of Vela, and Vela still will process the resulting webhook, this PR also validates the incoming deployment against the config and will stop the compilation if there's a mismatch.


KellyMerrick (approved):
lgtm

api/deployment/get_config.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/go-vela/server/database"
+	"github.com/go-vela/server/router/middleware/repo"
+	"github.com/go-vela/server/router/middleware/user"
+	"github.com/go-vela/server/scm"
+	"github.com/go-vela/server/util"
+)
+
+// swagger:operation GET /api/v1/deployments/{org}/{repo}/config deployments GetDeploymentConfig
+//
+// Get a deployment
```
plyr4, line 24:
```suggestion
// Get a deployment config
```

api/deployment/get_config.go
```diff
@@ -0,0 +1,124 @@
...
+func GetDeploymentConfig(c *gin.Context) {
+	// capture middleware values
+	l := c.MustGet("logger").(*logrus.Entry)
+	r := repo.Retrieve(c)
+	u := user.Retrieve(c)
+
+	ctx := c.Request.Context()
+
+	// capture ref from parameters - use default branch if not provided
+	ref := util.QueryParameter(c, "ref", r.GetBranch())
```
plyr4, line 74: should this query param be reflected in the api spec definition up above?

KellyMerrick (approved):
lgtm

plyr4 (approved):

PR #1215, opened 2024-10-25
enhance(ruleset): add stage status rule for independent stages

ecrupper:
Ref: https://github.com/go-vela/community/issues/910

Example pipeline:
```yaml
version: "1"

stages:
  failing:
    steps:
      - name: fail
        image: alpine:latest
        commands:
          - exit 1

      - name: run on failure
        image: alpine
        ruleset:
          status: failure
        commands:
          - echo fail

  indy:
    independent: true
    steps:
      - name: wait
        image: alpine:latest
        commands:
          - sleep 3

      - name: should not run
        image: alpine:latest
        ruleset:
          stage_status: failure
        commands:
          - echo "why am I running"
```

Before:
Screenshot 2024-10-25 at 10 06 57 AM

After:
Screenshot 2024-10-25 at 10 06 38 AM


wass3rw3rk:
i don't know if it's just me, but it seems counter intuitive to have `status: failure` that will *run* the step on build failure and then `stage_status: failure` which will *not run* the step.

wass3rw3rk:
to be clear, i'm not against the functionality. little hang up on the terminology and potential confusion with what we already have in place.

`status` implies build status, so `stage_status` seems fine to differentiate, though it's unclear if that means any stage's status, the current stage's status, a dependent stage's status, since there is a one to many relationship.

the other part is that a step with `status: failure` *runs* that step when there's a build failure, whereas a step within a stage with `stage_status: failure` *does not* run that step when there's a stage failure(?).

the independent stage is inverting the logic. at this time, i don't have a better alternative though 🤔

ecrupper:
I am going to close this PR in favor of pursuing a purely worker repo solution:

An independent stage should only use the `status` rule for its own stage's status. The point of an independent stage is to be independent from the rest of the build, so having a status rule that checks the overall build status doesn't make any sense. If that's a requirement, then the stage is no longer independent.

PR #1070, opened 2024-02-27
feat: add github app

jbrockopp:

PR #1217, opened 2024-10-25
feat: opt-in gh app integration

plyr4:
this PR adds opt-in [GitHub App](https://docs.github.com/en/apps/creating-github-apps/about-creating-github-apps/about-creating-github-apps) integrations to the server.

the main differences are:
- integrate with an OAuth GitHub App by providing an app id and private key
- use the GitHub App to generate the `.netrc` password for builds when the app is installed and it can access the repo
- adds webhook handlers for GitHub App installation and installation_repositories events, and syncs existing repos when changes are made
- ~adds Checks functionality, pulled from the original hackathon efforts #1070~ 
  - pulled out this code, it will be a follow-up PR

### New Flags

| Key                                          | Description | Default Value            |
| ----------------------------- | - | --------------------- |
| `VELA_SCM_APP_ID`                      | set to the App ID for the GitHub App | `N/A` |
| `VELA_SCM_APP_PRIVATE_KEY`         | the string value for the GitHub App private key generated through GitHub | `N/A` |

#### Permissions

the GitHub App requires the following permissions at the very minimum:

- `contents:read`
- `checks:write` 

builds would request write permissions through the `git` yaml block, see below.

#### Subscribed Events

- Installation target 

#### OAuth

the same configurations and oauth scopes should be assigned to the GitHub App, including:
- `oauth callback url` set to `/authenticate` (like usual)
- `Webhook URL` set to the base url (like usual) 

### New YAML block: `git`

integrating with a GitHub App allows the use of the `git` YAML block for customizing the permissions allocated to the netrc password embedded into Vela steps.

```yaml
git:
  token:
    repositories:
      - foo/bar
      - helloworld
    permissions:
      contents: write
      checks: write
```

this lets users customize the list of repositories that the `netrc` password has access to, but that list is restricted to ONLY the repos that the GitHub App org installation has been given access to. 

by default, the compiler will use the following configurations unless otherwise provided:

```yaml
git:
  token:
    repositories:
      - VELA_BUILD_REPO
    permissions:
      contents: read
      checks: write
```

## :warning: Netrc Considerations

this WILL impact builds, check out the following list of things to consider when migrating to GitHub App

### Cloning Private Repositories

Vela builds might lose the ability to read/write from certain private repos that the repo owner may have had access to due to the new restrictive policies set on the `netrc` token.

GitHub Apps do not support providing access to repos that are outside the installation org. meaning, for a Vela build to access private repos, Go modules, etc, that are _**outside of the repo's org**_ then the build author must provide override the `clone` step and use an alternative authentication method like a PAT

## :warning: OAuth Considerations

For this release, we recommend using a combination of both an OAuth app and a GitHub App with Authorization disabled.
The rest of the code base is not prepared to require user App installations, see the below examples for why.  

### Enabling Private Repositories (/source/repos)

Vela users will lose the ability to enable private repositories unless the GitHub App is installed to their personal account. this is due to changes to the default permissions when using a GitHub App as an OAuth provider.
see: https://docs.github.com/en/enterprise-server@3.13/apps/using-github-apps/authorizing-github-apps#difference-between-authorization-and-installation


JordanSussman:
Can you expand on why the `contents:write` is required instead of `contents:read`?

plyr4:
> Can you expand on why the `contents:write` is required instead of `contents:read`?

whoops thats a typo in the description. the code is accurate though, the minimum is `contents:read` 👍

scm/github/app_transport.go
```diff
@@ -0,0 +1,294 @@
...
+//
+// Client can also be overwritten, and is useful to change to one which
+// provides retry logic if you do experience retryable errors.
+//
+// See https://developer.github.com/apps/building-integrations/setting-up-and-registering-github-apps/about-authentication-options-for-github-apps/
+type Transport struct {
+	BaseURL                  string                           // BaseURL is the scheme and host for GitHub API, defaults to https://api.github.com
+	Client                   Client                           // Client to use to refresh tokens, defaults to http.Client with provided transport
+	tr                       http.RoundTripper                // tr is the underlying roundtripper being wrapped
+	appID                    int64                            // appID is the GitHub App's ID
```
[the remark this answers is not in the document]
plyr4: not true, its used as a claim in the round tripper. applying `nolint` sorry

scm/flags.go
```diff
@@ -67,4 +67,16 @@ var Flags = []cli.Flag{
...
+		EnvVars:  []string{"VELA_SCM_APP_ID", "SCM_APP_ID"},
+		FilePath: "/vela/scm/app_id",
+		Name:     "scm.app.id",
+		Usage:    "set ID for the SCM App integration (GitHub App)",
+	},
+	&cli.StringFlag{
+		EnvVars:  []string{"VELA_SCM_APP_PRIVATE_KEY", "SCM_APP_PRIVATE_KEY"},
+		FilePath: "/vela/scm/app_private_key",
+		Name:     "scm.app.private_key",
+		Usage:    "set value of base64 encoded SCM App integration (GitHub App) private key",
```
plyr4, line 80: should we make this "path to private key" instead of the value? would make it easier to drop files instead of env vars

JordanSussman:
> but that list is restricted to ONLY the repos that the GitHub App org installation has been given access to.

I think I know the answer, but just to confirm: if I install the GitHub app on both the foo and bar organizations and run a build within a repo in the foo org, does that mean I won’t be able to clone a repo from the bar org, even if I specify `bar/<repo>` in the token section?

plyr4:
> > but that list is restricted to ONLY the repos that the GitHub App org installation has been given access to.
> 
> I think I know the answer, but just to confirm: if I install the GitHub app on both the foo and bar organizations and run a build within a repo in the foo org, does that mean I won’t be able to clone a repo from the bar org, even if I specify `bar/<repo>` in the token section?

that's correct, org installations are scoped to repos in that org, unfortunately. its extremely similar to fine-grained access tokens (the beta PATs)

scm/github/repo.go
```diff
@@ -662,3 +674,132 @@ func (c *client) GetBranch(ctx context.Context, r *api.Repo, branch string) (str
...
+	}
+
+	for resource, perm := range permissions {
+		ghPermissions, err = ApplyInstallationPermissions(resource, perm, ghPermissions)
+		if err != nil {
+			return u.GetToken(), err
+		}
+	}
+
+	// the app might not be installedm therefore we retain backwords compatibility via the user oauth token
```
ecrupper, line 733:
```suggestion
	// the app might not be installed therefore we retain backwards compatibility via the user oauth token
```

scm/github/repo.go
```diff
@@ -662,3 +674,132 @@ func (c *client) GetBranch(ctx context.Context, r *api.Repo, branch string) (str
...
+			repos[i] = split[1]
+		}
+	}
+
+	// permissions that are applied to the token for every repo provided
+	// providing no permissions, nil, or empty map will default to the permissions
+	// of the GitHub App installation
+	//
+	// the Vela compiler follows a least-privileged-defaults model where
+	// the list contains only the triggering repo, unless provided in the git yaml block
```
ecrupper, line 713: I'm a little confused at this comment because the comment earlier in this function says
```
	// providing no repos, nil, or empty slice will default the token permissions to the list
	// of repos added to the installation
```

But then this one says that it will only contain the triggering repo. Am I getting confused?

api/repo/repair.go
```diff
@@ -62,6 +62,8 @@ import (
 
 // RepairRepo represents the API handler to remove
 // and then create a webhook for a repo.
+//
+//nolint:funlen // ignore statement count
 func RepairRepo(c *gin.Context) {
```
ecrupper, line 67: Do we want to add similar logic to the `api/scm/sync` API handlers?

database/build/last_repo_test.go
```diff
@@ -32,6 +32,7 @@ func TestBuild_Engine_LastBuildForRepo(t *testing.T) {
 	_repo.SetAllowEvents(api.NewEventsFromMask(1))
 	_repo.SetPipelineType(constants.PipelineTypeYAML)
 	_repo.SetTopics([]string{})
+	_repo.SetInstallID(0)
```
ecrupper, line 35: I'm somewhat sure that you can remove all these setters for install ID if you add the field to the testutils.APIRepo function

scm/flags.go
ecrupper: I think that is definitely more standard

scm/github/app_install.go
```diff
@@ -0,0 +1,156 @@
...
+	r.SetInstallID(installID)
+
+	h := new(types.Hook)
+	h.SetNumber(webhook.Hook.GetNumber())
+	h.SetSourceID(webhook.Hook.GetSourceID())
+	h.SetWebhookID(webhook.Hook.GetWebhookID())
+	h.SetCreated(webhook.Hook.GetCreated())
+	h.SetHost(webhook.Hook.GetHost())
+	h.SetEvent(constants.EventInstallation)
+	h.SetStatus(webhook.Hook.GetStatus())
```
ecrupper, line 83: the webhook.Hook should already have all this right? Is there a reason we create a new one here?

ecrupper (changes requested):

api/repo/repair.go
plyr4: yeah good call

scm/github/app_install.go
plyr4: yeah its a deep copy of the base hook for each installed repo since thats a one-to-many for install-to-repohooks.
we could re-use webhook.Hook but felt cleaner to do a deep copy and not mutate the input pointer.

database/build/last_repo_test.go
plyr4: you are a legend 🙏 like 50 changed files removed

scm/github/github_client.go
```diff
@@ -0,0 +1,158 @@
...
+
+	opts := &github.InstallationTokenOptions{
+		Repositories: repos,
+		Permissions:  permissions,
+	}
+
+	id := r.GetInstallID()
+
+	// if repo has an install ID, use it to create an installation token
+	if id == 0 {
```
wass3rw3rk, line 86: nit: comment is a little confusing if we're checking if value is `0` right after. i think we're trying to get the id from installations if we DON'T have a installation ID. but i could be wrong 🫣

wass3rw3rk (approved):
super minor stuff; tested functionally and works in both modes as advertised. non-breaking change until you enable app integration. nice work.

api/auth/get_token.go
```diff
@@ -69,6 +79,32 @@ func GetAuthToken(c *gin.Context) {
 	// capture the OAuth state if present
 	oAuthState := c.Request.FormValue("state")
 
+	// handle scm setup events
+	// setup_action==install represents the GitHub App installation callback redirect
+	if c.Request.FormValue("setup_action") == constants.AppInstallSetupActionInstall {
+		installID, err := strconv.ParseInt(c.Request.FormValue("installation_id"), 10, 0)
```
wass3r, line 85: we need to add installation_id to swagger annotation?

scm/flags.go
plyr4: done! heres a conversion one liner if you want to test using the key you might already have in `.env`
 
```
( source .env; echo -n "$VELA_SCM_APP_PRIVATE_KEY" |  base64 -d ) >> .privatekey && ( echo "VELA_SCM_APP_PRIVATE_KEY_PATH=.privatekey" >> .env && sed -i~ '/^VELA_SCM_APP_PRIVATE_KEY=/s/=.*/=""/' .env && rm .env~)
```

api/auth/get_token.go
plyr4: done

scm/github/github_client.go
plyr4: good callout, thats confusing. i updated it as follows
> // if the source scm repo has an install ID but the Vela db record does not
> // then use the source repo to create an installation token

wass3r (approved):

PR #1218, opened 2024-10-29
chore(lint): address existing linter issues

ecrupper:
A few things of note:

- I ditched `dupl`. This rule was really only pointing out false positives. I suppose you could argue there could be some convoluted abstraction to make step and service not copy so much code, but I don't think that would be worth it. It's really just noise right now.
- `int32` for DB types does not play well with `int`. `int` will expand based on system, so converting it to `int32` does not make much sense, especially when we are using `INTEGER` for the DB table declaration for both `int32` and `int64`. So I just made all `int` convert to `int64` for the DB.
- `contextcheck` was pretty particular about using `database.FromContext(c)` from inside helper functions, so I just grabbed the interface and passed that along instead.
- `uint64` for Starlark exec limit cannot be translated to a sql type, so I moved the conversion for that to right before we pass it along to Starlark.

golang-ci lint ✅ 🤞


wass3rw3rk (commented):
can we remove this too, then?

https://github.com/go-vela/server/blob/31e3eb432e3ee78fb9f44246af6d18b0644f00c5/.golangci.yml#L171

compiler/template/starlark/render.go
```diff
@@ -30,19 +31,20 @@ var (
...
+	if limit < 0 {
+		return nil, fmt.Errorf("starlark exec limit must be non-negative")
+	}
 
-	globals, err := starlark.ExecFile(thread, tName, tmpl, predeclared)
+	thread.SetMaxExecutionSteps(uint64(limit))
+
+	predeclared := starlark.StringDict{"struct": starlark.NewBuiltin("struct", starlarkstruct.Make)}
 
+	globals, err := starlark.ExecFileOptions(syntax.LegacyFileOptions(), thread, "templated-base", tmpl, predeclared)
```
plyr4, line 47: does this change impact functionality?
```go
globals, err := starlark.ExecFileOptions(syntax.LegacyFileOptions(), thread, "templated-base", tmpl, predeclared)
```

plyr4 (approved):
awesome

compiler/template/starlark/render.go
ecrupper: https://github.com/google/starlark-go/blob/1207426daebd7ea8eb86b8d917a344a72cc5167b/starlark/eval.go#L333

It's just using what the now deprecated function has. I think we can investigate turning some of those options on, but they're all set to false right now

wass3rw3rk (approved):
thanks!

PR #1220, opened 2024-11-05
enhance(yaml)!: allow for users to parse pipelines using old library

ecrupper:
## Background

It is imperative that the platform moves away from the buildkite YAML library, as it is archived.

This has been attempted and reverted a couple times:

(2024, go-yaml v3) https://github.com/go-vela/types/pull/386 —> https://github.com/go-vela/types/pull/392

(2020, goccy) https://github.com/go-vela/types/pull/118 —> https://github.com/go-vela/types/pull/144

This PR ensures that we will not have to revert this important change, as it will allow for users to immediately set 
```yaml
version: "legacy"
```
At the top level of their pipeline file (or template file) if they are encountering unexpected YAML issues.

This will also allow for troubleshooting / triaging for the Vela admins to cover all of our bases when shifting to this new library while having low impact.

The goal is to revert this change in a later release so that the entire platform uses only one YAML parser throughout the code base. 

However, one of the most common patterns that will now produce errors is the use of multiple `<<:` in a single YAML map. This pattern will compile correctly with this change but produce warnings. These warnings will be attached to the pipeline object and will be rendered in the UI. 

## What's in the PR

- new packages `compiler/types/yaml/yaml` and `compiler/types/yaml/buildkite` in place of the single `compiler/types/yaml`
- all the buildkite YAML types have a `ToYAML()` method that will convert to the go-yaml version. Once the pipelines have been unmarshaled, these two objects should be identical sans their type.
- an `internal/yaml.go` file that grabs the `version` from the pipeline file and verifies the library we should use.
- a new `warnings` field in the pipelines table to track warnings.


ecrupper:
Closing. Decided to break this PR up and https://github.com/go-vela/server/pull/1232 is the last of the effort.

PR #1224, opened 2024-12-02
enhance(compiler): cache templates per compilation

ecrupper:
This is a fairly uncommon issue, but whenever folks are looking to use a template multiple times in one build, Vela will fetch that template each time. With this in-memory cache on the compiler object, we can avoid unnecessary calls to the SCM.


KellyMerrick (approved):
lgtm

wass3rw3rk (approved):

PR #1227, opened 2024-12-16
feat(repo)!: add pending approval timeout

ecrupper:
Give repos a new setting called `approval_timeout` which will set builds that have been sitting in a `pending approval` state to a failure if they've been unapproved for the given time. 

Also applied some approval logic to restarts now that we store `fork` on builds.


KellyMerrick (approved):
lgtm

api/build/traffic.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	// publish the build to the queue
+	go Enqueue(
+		context.WithoutCancel(c.Request.Context()),
+		queue.FromGinContext(c),
+		database.FromContext(c),
+		item,
+		b.GetHost(),
+	)
```
plyr4, line 88: thoughts on separating the logic from the action taken? 
> 	// determine whether to send compiled build to queue 

do the determining, `return bool error` then gatekeep or enqueue afterwards?
seems cleaner that way, like how we separate logic and API response writer code in other spots

plyr4: returning bool based on some logic might also make the function more testable

ecrupper: I like the idea. There's a few spots within TrafficBuild and gatekeepBuild that leverage contextual services, which may make testing a little tricky. But I'll see what I can do.

KellyMerrick (approved):
lgtm

cmd/vela-server/main.go
```diff
@@ -133,6 +133,18 @@ func main() {
...
+			Name:    "default-approval-timeout",
+			Usage:   "override default approval timeout (days)",
+			Value:   constants.ApprovalTimeoutDefault,
+		},
+		&cli.DurationFlag{
+			EnvVars: []string{"VELA_BUILD_PENDING_APPROVAL_TIMEOUT"},
+			Name:    "pending-approval-timeout",
+			Usage:   "timeout for a build to wait for approval before being set to error",
+			Value:   7 * 24 * time.Hour,
+		},
```
wass3rw3rk, line 147: this does not seem to be used anywhere?

wass3rw3rk (approved):

plyr4 (approved):
sweet

PR #1228, opened 2024-12-18
refactor(db/build): drop source index, add event index

wass3rw3rk:
`source` column was used by legacy deployment implementation so should be dropped. it also had high cardinality, so is likely relatively large on bigger installations.

`event` has become more of a bottleneck for some queries for very busy repos when querying builds by event. we recommend adding it.

the changes in this PR only affect new installations. to accommodate these in existing installations, drop the `source` index (concurrently on Postgres) and add the new one with `CREATE INDEX CONCURRENTLY IF NOT EXISTS builds_event ON builds (event);` (for Postgres).

probably still wise to create the index prior to upgrading unless you have the [`VELA_DATABASE_SKIP_CREATION` or `DATABASE_SKIP_CREATION` flag](https://github.com/go-vela/server/blob/355107dec2300fde1ecd191be39e5a7620aa0078/database/flags.go#L90-L95) set to true, since the default statements don't use `CONCURRENTLY` on Postgres.


ecrupper (approved):

KellyMerrick (approved):
lgtm

PR #1232, opened 2024-12-31
feat(pipeline)!: add warnings to pipelines that use legacy YAML lib or have dupl anchors in map

ecrupper:
Added `warnings` field to pipelines that will aggregate non-breaking but call-out worthy YAML transgressions.

### DIFF
--- a/database/types/pipeline.go
+++ b/database/types/pipeline.go
@@ -202,13 +202,13 @@ func (p *Pipeline) Validate() error {
 		return ErrEmptyPipelineVersion
 	}
 
-	// calculate total size of favorites
+	// calculate total size of warnings
 	total := 0
 	for _, w := range p.Warnings {
 		total += len(w)
 	}
 
-	// verify the Favorites field is within the database constraints
+	// verify the Warnings field is within the database constraints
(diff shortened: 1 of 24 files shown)


database/types/pipeline.go
```diff
@@ -193,6 +202,19 @@ func (p *Pipeline) Validate() error {
...
+	for _, w := range p.Warnings {
+		total += len(w)
+	}
+
+	// verify the Favorites field is within the database constraints
+	// len is to factor in number of comma separators included in the database field,
+	// removing 1 due to the last item not having an appended comma
+	if (total + len(p.Warnings) - 1) > constants.PipelineWarningsMaxSize {
+		return ErrExceededWarningsLimit
+	}
```
wass3rw3rk, line 216: strange way to deal with "favorites" ;) — see code comments

wass3rw3rk (changes requested):

database/types/pipeline.go
ecrupper: :pain_disintegrate:

wass3rw3rk (approved):

plyr4 (approved):
awesome!

PR #1235, opened 2025-01-02
feat: status endpoints

plyr4:
implements https://github.com/go-vela/community/issues/749 in a way that doesnt impact current status svg responses.

instead, two new endpoints to get unauthenticated access to "lite" versions of repo and builds.

### Concerns
- potential impact to the database from unauthenticated and untrackable spam. but... this is not a new problem. could easily do it elsewhere (dashboards, pat auth, etc) if you were determined, but still.
- the `/badge` endpoint for "private repos" have never been protected, but this is obviously opening the door much wider.

### DIFF
--- a/api/build/status.go
+++ b/api/build/status.go
@@ -11,7 +11,7 @@ import (
 	"github.com/go-vela/server/router/middleware/build"
 )
 
-// swagger:operation GET /status/{org}/{repo}/{build} builds GetStatus
+// swagger:operation GET /status/{org}/{repo}/{build} builds GetBuildStatus
 //
 // Get a build status
 //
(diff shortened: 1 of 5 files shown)


timhuynh94 (dismissed):

wass3rw3rk (commented):
test fail. swagger annotation is defining `GetStatus` twice. otherwise 👍🏼

timhuynh94 (approved):

wass3rw3rk (commented):
what do you think about dropping `email` from the build status endpoint at least? maybe `deploy_payload`, as well.

wass3rw3rk (approved):

timhuynh94 (approved):

PR #1239, opened 2025-01-10
fix(deployment): use Config instead of ConfigBackoff when fetching deploy config

ecrupper:
users supplying a bad ref should not result in a several second response time. the backoff is only helpful when dealing with concurrent webhooks with legit refs


wass3rw3rk (approved):

KellyMerrick (approved):
lgtm

PR #1242, opened 2025-01-14
fix(build): ensure count/pagination links account for filter/query params

wass3rw3rk:
for listing builds, the previous implementation did a count, but always returned the total count while excluding after/before query params (but honoring other filters, if present). this resulted in incorrect header links. the header links pointing at subsequent pages also didn't honor existing query params, so if you followed the links you could end up with different results.


plyr4 (approved):
lgtm

plyr4 (approved):

timhuynh94 (approved):

PR #1244, opened 2025-01-17
Feat test report

timhuynh94:


wass3rw3rk (commented):
i'm hoping we can break this PR up. maybe one that just deals with storage addition, and one that deals with the test reporting additions (at the very least).

PR #1249, opened 2025-01-30
fix(compiler): bust template cache on Duplicate

ecrupper:
When we don't set this to an empty map, it'll use old sources of templates


plyr4 (approved):

KellyMerrick (approved):
lgtm

PR #1255, opened 2025-02-17
refactor(ruleset)!: improve match code efficiency and properly handle status

ecrupper:
```yaml
steps:
  - name: test
    image: alpine:latest
    ruleset:
      event: pull_request
      branch: [ hotfix-* ]
      operator: or
    commands:
      - echo blah
```

The above does not purge. It only gets skipped when the worker handles skipping. The compiler should be corrected here and return `FALSE` when the above is calculated with `rule data == push on main` for example

Further, I refactored the match code to be more efficient. I removed the unnecessary `Parallel` key, which will be addressed in a follow-up worker PR, which introduces stage status.

Added some bench mark testing which is just there for fun right now, but I used it to confirm the improvements by comparing these changes to main. Improvement of about half a second, which definitely adds up over parsing thousands of rulesets.

Another thing I added was separation of `if` and `unless` operators/matchers if defined. Otherwise, they'll inherit from the `Ruleset` like usual (non-breaking).


plyr4 (dismissed):

KellyMerrick (dismissed):

timhuynh94 (approved):

KellyMerrick (dismissed):

KellyMerrick (approved):

KellyMerrick (approved):

compiler/types/pipeline/ruleset_test.go
```diff
@@ -269,7 +298,139 @@ func TestPipeline_Ruleset_Match(t *testing.T) {
...
+			Target:   []string{"production"},
+			Label:    []string{"label1", "label2"},
+			Instance: []string{"instance1"},
+			Operator: "or",
+			Matcher:  "filepath",
+		},
+	}
+
+	// run the benchmark
+	for i := 0; i < b.N; i++ {
```
wass3rw3rk, line 430: we should use b.Loop instead - see https://go.dev/blog/testing-b-loop

wass3rw3rk (changes requested):
love seeing the `testing.B` goodness. just a quick change on that part. planning to review a bit more soon!

KellyMerrick (dismissed):

wass3rw3rk:
nice work. we might need some doc updates also for ruleset changes

PR #1258, opened 2025-02-19
fix(deployment): validate inputs in a case-insensitive manner

ecrupper:
Given the following deployment config:
```yaml
deployment:
  parameters:
    region:
      description: cluster region to deploy
      required: true
      type: string
      options:
        - us-west-1
        - us-south-2
    
    cluster_count:
      description: number of clusters to deploy to
      required: false
      type: integer
      max: 10
```

A deployment triggered via CLI like this:
```sh
vela add deployment  --target here --parameter "region=us-west-1" --parameter "CLUSTER_COUNT=50"
```

Would pass validation, even though the parameter is injected as `DEPLOYMENT_PARAMETER_CLUSTER_COUNT` regardless of the casing when it's passed in.

### DIFF
--- a/compiler/types/pipeline/deployment.go
+++ b/compiler/types/pipeline/deployment.go
@@ -63,9 +63,9 @@ func (d *Deployment) Validate(target string, inputParams map[string]string) erro
 	}
 
 	// make lowercase key map for validation (it is all uppercase in the end but users can write YAML how they want)
-	standardizedInput := make(map[string]string)
-	for k := range inputParams {
-		standardizedInput[strings.ToLower(k)] = inputParams[k]
+	standardizedInput := make(map[string]string, len(inputParams))
+	for k, v := range inputParams {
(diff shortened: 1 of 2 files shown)


KellyMerrick (approved):

compiler/types/pipeline/deployment.go
```diff
@@ -61,21 +62,31 @@ func (d *Deployment) Validate(target string, inputParams map[string]string) erro
 		return fmt.Errorf("deployment target `%s` not found in deployment config targets", target)
 	}
 
+	// make lowercase key map for validation (it is all uppercase in the end but users can write YAML how they want)
+	standardizedInput := make(map[string]string)
```
wass3rw3rk, line 66: minor: since we know the size, we can probably make a right-sized map

wass3rw3rk (approved):

KellyMerrick (approved):

PR #1260, opened 2025-02-21
fix(database)!: adjust table types to match database/sql types

ecrupper:
`github.com` has deployment numbers exceeding `int4` size. Local stacks hooked up to public github are unable to test deployments as a result. 

This also obviously spells problems for actual installations. There will be an addition to the migration script to call this out and suggest people use the `BIGINT` column type when dealing with `sql.NullInt64` values

Also forcing the use of `int32` for types that have `INTEGER` columns so there is no system mismatch which could cause problems.


KellyMerrick (approved):

timhuynh94 (dismissed):

KellyMerrick (approved):

wass3r:
We should mark this as breaking probably

KellyMerrick (approved):

timhuynh94 (approved):

PR #1262, opened 2025-02-25
fix: support list of cors origins

plyr4:
this "fix" adds support for alternative browser client implementations

## Environment Variable

use `VELA_CORS_ALLOW_ORIGINS="https://vela-frontend-module.com,https://other-platform-frontend-module.com"` to set an optional list of CORS allowed origins 

the code still defaults to `m.Vela.WebAddress` so current configurations of `VELA_WEB_ADDRESS` are not impacted

### DIFF
--- a/cmd/vela-server/main.go
+++ b/cmd/vela-server/main.go
@@ -76,6 +76,11 @@ func main() {
 			Value:   "/account/authenticate",
 		},
 		&cli.StringFlag{
+			EnvVars: []string{"VELA_CORS_ALLOW_ORIGINS", "VELA_CORS_ALLOWED_ORIGINS"},
+			Name:    "cors-allow-origins",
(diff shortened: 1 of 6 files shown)


cmd/vela-server/main.go
```diff
@@ -75,6 +75,11 @@ func main() {
 			Usage:   "web ui oauth callback path",
 			Value:   "/account/authenticate",
 		},
+		&cli.StringFlag{
```
ecrupper:
```suggestion
		&cli.StringSliceFlag{
```

Right? Now I'm curious if the CLI library just parsed it out anyway when you referenced it as `c.StringSlice(` later?

ecrupper (approved):

wass3rw3rk (approved):

PR #1268, opened 2025-03-11
fix(compiler): split validate into YAML and Pipeline

ecrupper:
There are a couple things that are checked during `Validate` that may not accurately reflect the build that ends up being run. For instance, you could name steps the same if it's guaranteed they will never run at the same time due to rulesets. By splitting validate into a YAML validation and a pipeline validation, we have a more accurate validator.

### DIFF
--- a/compiler/native/validate.go
+++ b/compiler/native/validate.go
@@ -7,12 +7,13 @@ import (
 
 	"github.com/hashicorp/go-multierror"
 
+	"github.com/go-vela/server/compiler/types/pipeline"
 	"github.com/go-vela/server/compiler/types/yaml/yaml"
 	"github.com/go-vela/server/constants"
 )
 
 // Validate verifies the yaml configuration is valid.
-func (c *client) Validate(p *yaml.Build) error {
+func (c *client) ValidateYAML(p *yaml.Build) error {
 	var result error
 	// check a version is provided
 	if len(p.Version) == 0 {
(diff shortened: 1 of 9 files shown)


timhuynh94 (dismissed):

KellyMerrick (approved):

KellyMerrick (approved):

timhuynh94 (approved):

timhuynh94 (dismissed):

compiler/native/validate.go
```diff
@@ -46,118 +48,159 @@ func (c *client) Validate(p *yaml.Build) error {
...
+	// validate the steps block provided
+	err = validatePipelineContainers(p.Steps, &reportCount, make(map[string]string), make(map[string]bool), "")
+	if err != nil {
+		result = multierror.Append(result, err)
+	}
+
+	return result
+}
+
+// validateStages is a helper function that verifies the
```
wass3rw3rk, line 160: function name is not correct

compiler/native/validate.go
```diff
@@ -46,118 +48,159 @@ func (c *client) Validate(p *yaml.Build) error {
...
 	return result
 }
 
-// validateServices is a helper function that verifies the
-// services block in the yaml configuration is valid.
-func validateServices(s yaml.ServiceSlice) error {
-	for _, service := range s {
-		if len(service.Name) == 0 {
-			return fmt.Errorf("no name provided for service")
+// validateStages is a helper function that verifies the
```
wass3rw3rk, line 71: function name not correct

wass3rw3rk (changes requested):
super minor changes

KellyMerrick (approved):

PR #1273, opened 2025-03-14
enhance(scm)!: support truly custom webhook address

ecrupper:
If an installation wants to support a different webhook address for a pass through to the Vela server, the suffix of `/webhook` should not be forced onto those custom addresses

Marking as breaking because users could have already accounted for this by supplying a custom address knowing it ends in `/webhook`


plyr4 (approved):
sweet

KellyMerrick (approved):

timhuynh94 (approved):

PR #1277, opened 2025-03-26
fix(linter): migrate to v2 due to breaking changes

timhuynh94:


.golangci.yml
```diff
@@ -9,166 +10,90 @@
...
   # colored-line-number|line-number|json|tab|checkstyle|code-climate, default is "colored-line-number"
   formats:
-    - format: json
+    text:
+      path: stdout
+      # print linter name in the end of issue text, default is true
+      print-linter-name: true
+      # print lines of code with issue, default is true
+      print-issued-lines: true
+    json:
```
wass3rw3rk, line 19: let's get rid of the json output, it was added in an attempt to fix an issue elsewhere and didn't actually address the issue.

.golangci.yml
```diff
@@ -9,166 +10,90 @@
...
-# golangci will execute. Several of them were disabled by
-# default but we've opted to enable them.
 linters:
   # disable all linters as new linters might be added to golangci
-  disable-all: true
-
+  # in v2, this is for disable-all: true
+  default: none
   # enable a specific set of linters to run
   enable:
```
wass3rw3rk, line 26: curious why the `enable:` list is empty in the new version?

timhuynh94: not sure why it was empty but should be fixed now

wass3rw3rk: thanks. i see we re-added the list of inactive rules also. i wonder if we should even keep that in? i think the only value it provides is a record of decision of _why_ certain linters are not enabled via the comments. but there's some maintenance overhead with keeping it accurate. some entries are already not supported anymore. git history is probably a good enough record keeper?

timhuynh94: git history is good but would be a pain to search through the history if needed? I'm fine either way.

wass3rw3rk (approved):

ecrupper (approved):

PR #1281, opened 2025-04-01
feat: custom properties tracking for repository resource and OIDC claim field

ecrupper:
Adding the `custom_props` field that is part of standard webhook payloads. This extra data can be helpful for enterprise cloud instances, especially when combined with OIDC claims.


internal/token/mint.go
```diff
@@ -137,6 +137,8 @@ func (tm *Manager) MintIDToken(ctx context.Context, mto *MintTokenOpts, db datab
 	// initialize claims struct
 	var claims = new(api.OpenIDClaims)
 
+	logrus.Infof("MTO MTO MTO: %v", mto.Build.GetRepo().GetCustomProps())
```
wass3rw3rk, line 140: is this intended?

api/webhook/post.go
```diff
@@ -237,6 +237,18 @@ func PostWebhook(c *gin.Context) {
 		}
 	}
 
+	// update custom props on all events (some repos may have these fields set prior to installing Vela)
+	repo.SetCustomProps(r.GetCustomProps())
+	repo.SetTopics(r.GetTopics())
+
+	repo, err = db.UpdateRepo(ctx, repo)
```
wass3rw3rk, line 244: we're updating repo twice then in this flow? thrice for repo events, i think? does this need to happen here?

internal/token/mint.go
ecrupper: Nope! thanks lol

api/webhook/post.go
ecrupper: let me move it under this repo event block... but yeah we need this here because the `CompileAndPublish` does not take in the webhook repo that is parsed

wass3rw3rk (approved):

PR #1290, opened 2025-04-07
chore(deps): migrate urfave/cli from v2 to v3

ecrupper:
Following migration guide from https://cli.urfave.org/migrate-v2-to-v3/

The main takeaway is that `cli.Context` is not a thing anymore; everything is baked into the `cli.Command`.

I added a bunch of flag validation... some of it is doubly applied (see SCM, Database, etc). I couldn't decide if we should validate flags or structs for those packages since they _could_ be exportable as services in other codebases, though not realistically. I kept both validation. Would be open to removing them.


timhuynh94 (approved):

util/util.go
```diff
@@ -145,3 +145,17 @@ func Sanitize(field string) string {
...
+	if v > 2147483647 {
+		return 2147483647
+	}
+
+	if v < -2147483648 {
+		return -2147483648
+	}
+
+	return int32(v)
+}
```
wass3rw3rk, line 161: suggestion: utilize constants math.MinInt32 and math.MaxInt32

wass3rw3rk (approved):

KellyMerrick (approved):

timhuynh94 (approved):

wass3rw3rk (approved):

PR #1291, opened 2025-04-08
fix(lint): export returned structs from instantiation funcs and pass context to queue

ecrupper:
The `revive` linter keeps complaining about unexported returns from our `New` funcs. The linter also gets confused as to _when_ that rule is actually being applied or not when `//nolint:revive` is declared. It's a bit of a mess, but the solution seems to be rather simple: just fix it? 

The unexported fields that belong to these structs are what is set during these instantiations, so I see no issue in just exporting the structs themselves. Let me know if you disagree!

There were also some linter errors about the lack of context passing to various functions in `cmd/vela-server/server.go`. I addressed those as well by making the following updates:

**START**: Parent Context (`context.Background()` from `cmd.Run(context.Background(), os.Args)`) 

- Passed to all setup funcs (`native.FromCLICommand(ctx, cmd)`, `database.GetSettings(ctx)`, etc)
- Passed to subprocesses: settings refresh, cleanup pending approval builds, process schedules

**NEXT**: Parent Context -> NEW context with cancel (`sCtx`)

**THEN**: `g` error group formed from `sCtx`, which runs all the subprocesses. Only the server failing can cancel `sCtx` which will cancel the error group. Failures from other subprocesses are not actually returning errors in the code, so the context should not be cancelable (hence why we pass in parent context).


wass3rw3rk (commented):
should we call out the context updates also in the description?

KellyMerrick (approved):

timhuynh94 (approved):

PR #1294, opened 2025-04-09
Test feat test report

timhuynh94:


cmd/vela-server/server.go
```diff
@@ -193,10 +199,12 @@ func server(ctx context.Context, cmd *cli.Command) error {
 		middleware.Metadata(metadata),
 		middleware.TokenManager(tm),
 		middleware.Queue(queue),
+		middleware.Storage(st),
 		middleware.RequestVersion,
 		middleware.Secret(cmd.String("vela-secret")),
 		middleware.Secrets(secrets),
 		middleware.Scm(scm),
+		middleware.Storage(st),
```
wass3rw3rk: added twice?

api/admin/storage.go
```diff
@@ -0,0 +1,162 @@
...
+
+	"github.com/gin-gonic/gin"
+	"github.com/sirupsen/logrus"
+
+	"github.com/go-vela/server/api/types"
+	"github.com/go-vela/server/storage"
+	"github.com/go-vela/server/util"
+)
+
+// swagger:operation PUT /api/v1/admin/storage/bucket admin CreateBucket
```
wass3rw3rk: why PUT instead of POST?

api/storage/storage.go
```diff
@@ -0,0 +1,238 @@
...
+
+	"github.com/gin-gonic/gin"
+	"github.com/sirupsen/logrus"
+
+	"github.com/go-vela/server/api/types"
+	"github.com/go-vela/server/storage"
+	"github.com/go-vela/server/util"
+)
+
+// swagger:operation POST /api/v1/storage/info storage StorageInfo
```
wass3rw3rk: This should be GET?

api/storage/storage.go
```diff
@@ -0,0 +1,238 @@
...
+//   '500':
+//     description: Unexpected server error
+//     schema:
+//       "$ref": "#/definitions/Error"
+
+// ListObjects represents the API handler to list objects in a bucket.
+func ListObjects(c *gin.Context) {
+	enable := c.MustGet("storage-enable").(bool)
+	if !enable {
+		l := c.MustGet("logger").(*logrus.Entry)
```
wass3rw3rk: we have to get logger either way, so just get it at the top level?

KellyMerrick:
I've been using `attachments` or `test attachments`. `Artifacts` also works. The database is currently coded as `testattachments` and `testreports`. Easily can be simplified to just `attachments` or `artifacts`, and `reports`.

Agree, the naming has been difficult and am completely open to a rename or at least alignment for consistency across the codebase.

`Test reports` is not quite right to me, as `Reports`, can be more than just test "attachments", such as a test results summary, code coverage, or anything pertinent to a single build.

Rough idea, with just `attachments` for now, but could have other options in the future:
image

PR #1304, opened 2025-04-24
feat!: role mapping for vela permissions

ecrupper:
At least for now, I don't really see why Vela permissions need to be as granular as they have become for GitHub recently. The standard permission levels are now considered legacy, with `roles` taking their place.

Further, there are now abilities for organizations and enterprises to define custom roles. It will be super useful for Vela to be able to interpret all these role types and translate them to the core Vela permissions: `read`, `write`, and `admin` (or `none` I suppose).


KellyMerrick (approved):

router/middleware/perm/perm.go
```diff
@@ -270,7 +271,7 @@ func MustSecretAdmin() gin.HandlerFunc {
 				logger.Errorf("unable to get user %s access level for org %s: %v", u.GetName(), o, err)
 			}
 
-			if !strings.EqualFold(perm, "admin") {
+			if !strings.EqualFold(perm, constants.PermissionAdmin) {
```
wass3rw3rk, line 274: same here and below in this file

scm/github/access_test.go
```diff
@@ -74,7 +74,7 @@ func TestGithub_OrgAccess_Member(t *testing.T) {
 	defer s.Close()
 
 	// setup types
-	want := "member"
+	want := "read"
```
wass3rw3rk, line 77: should we use constants here?

wass3rw3rk (commented):
couple of minor things

wass3rw3rk (commented):
can we address the one linter issue
image

KellyMerrick (approved):

wass3rw3rk (approved):

PR #1308, opened 2025-05-05
enhance: optimize source repos

ecrupper:
The current implementation fetches all repo records in an org. For any mono-org installation, this could potentially be a cumbersome call (fetch all repos). 

This change introduces a `GetReposInList` DB function to fetch fewer records. To make this more seamless / have less processing, I made the SCM function `ListUserRepos` just return a list of full names rather than a bunch of partial structs.

That's the main helper here, but I did add some additional optimizations in the handler to avoid retreading the map values when filling in active status.


wass3r (approved):
🥳

plyr4 (approved):
awesome

PR #1311, opened 2025-05-08
enhance(log)!: add timestamp field

wass3rw3rk:
adding timestamp field to log table.

to note, `BIGINT` used for consistency with other tables, but might not be the most straight-forward for partitioning by time spans. open to change that or add on to it.

### DIFF
--- a/api/log/create_service.go
+++ b/api/log/create_service.go
@@ -5,6 +5,7 @@ package log
 import (
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/gin-gonic/gin"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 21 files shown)


ecrupper (changes requested):
Nice — I believe we'll want to also add a line to set the timestamp in the `Plan` functions for steps and services. The worker never actually calls the POST endpoint for these logs, just PUTs. 

I tried this branch locally and the timestamp was populated automatically because of the `created_at` column name, but it attempted to use a non-Unix date format and bombed out.

wass3r:
@ecrupper good catch, added in https://github.com/go-vela/server/pull/1311/commits/eacadd4cf68367710d05b5398223c4572c4a0451

plyr4 (approved):

ecrupper (approved):

PR #1316, opened 2025-05-16
fix(secret): use different query for empty allowlist

ecrupper:
Tried to be too cute with the `NOT IN` clause. Empty inputs are invalid.

Also making use of the unique util func that we use on images.


wass3rw3rk (commented):
not directly related to your changes here, but since you are dabbling in this area, we should probably sanitize like we do for images: https://github.com/go-vela/server/blob/51dbf45ee6dd0504872310a787e87d9672d208cd/database/types/secret.go#L257-L261

wass3rw3rk (approved):
nice - thanks!

KellyMerrick (approved):

PR #1318, opened 2025-05-28
enhance(oidc)!: stringify int and bool claims

ecrupper:
Some integrations (namely Artifactory) expect all OIDC claims to be strings. Updating these will allow for better mapping options.


wass3rw3rk:
closes https://github.com/go-vela/community/issues/1036

PR #1320, opened 2025-05-29
feat!: support restart limits for pending builds based on queue length

ecrupper:
### Breaking Changes

Adding a `Route` field to `Build` and a `QueueRestartLimit` field to `Settings`, which have new columns in the DB.

Before, to maintain a record of the route, we would set the `Host` field to the route until the job was picked up by a worker. This change allows us to key off of a more stateful route value and check if a build should be allowed to restart.

### Core Change

Admins can set a max limit for queue length before the server will start rejecting restart requests from pending builds. It's less of protection against natural build increases but rather protection against unnecessary restart requests.


wass3rw3rk (commented):
weary about adding a column. should the new column be indexed? i imagine on a large table like this and folks spamming restart, even with the limitation, this would cause a noticeable performance impact that might negate the good that this PR is trying to introduce.

api/build/restart.go
```diff
@@ -105,6 +107,25 @@ func RestartBuild(c *gin.Context) {
...
+		// check length of specified route for the build
+		queueLength, err := queue.FromContext(c).RouteLength(ctx, b.GetRoute())
+		if err != nil {
+			util.HandleError(c, http.StatusInternalServerError, fmt.Errorf("unable to get queue length for %s: %w", b.GetRoute(), err))
+
+			return
+		}
+
+		if queueLength >= int64(ps.GetQueueRestartLimit()) {
+			retErr := fmt.Errorf("unable to restart build %s: queue length %d exceeds configured limit %d", entry, queueLength, ps.GetQueueRestartLimit())
```
wass3rw3rk, line 121: should we make this more actionable? i know we are not good with this elsewhere (hello "context exceeded"!), but would be nice to talk about what the user can do in this situation (other than message admins). i realize queue might exist for various reasons though - just "thinking" out loud here. i don't have great suggestions at the moment.

(same in redeliver, etc)

api/types/build.go
```diff
@@ -102,6 +103,7 @@ func (b *Build) Environment(workspace, channel string) map[string]string {
 		"VELA_BUILD_EVENT":         ToString(b.GetEvent()),
 		"VELA_BUILD_EVENT_ACTION":  ToString(b.GetEventAction()),
 		"VELA_BUILD_HOST":          ToString(b.GetHost()),
+		"VELA_BUILD_ROUTE":         ToString(b.GetRoute()),
```
wass3rw3rk, line 106: is this the same as `VELA_BUILD_CHANNEL`? should there be changes made to that at all? should we differentiate between the two better if both need to exist?

api/build/restart.go
```diff
@@ -105,6 +107,25 @@ func RestartBuild(c *gin.Context) {
...
+		// check length of specified route for the build
+		queueLength, err := queue.FromContext(c).RouteLength(ctx, b.GetRoute())
+		if err != nil {
+			util.HandleError(c, http.StatusInternalServerError, fmt.Errorf("unable to get queue length for %s: %w", b.GetRoute(), err))
+
+			return
+		}
+
+		if queueLength >= int64(ps.GetQueueRestartLimit()) {
+			retErr := fmt.Errorf("unable to restart build %s: queue length %d exceeds configured limit %d. Please wait for the queue to decrease in size before retrying", entry, queueLength, ps.GetQueueRestartLimit())
```
wass3rw3rk, line 121: i like it, but always think it's strange to start capitalizing in the middle of error message. perhaps `;` or `,` so we can cheat and continue w/ lower case?

```suggestion
			retErr := fmt.Errorf("unable to restart build %s: queue length %d exceeds configured limit %d, please wait for the queue to decrease in size before retrying", entry, queueLength, ps.GetQueueRestartLimit())
```

wass3rw3rk (commented):
like it - thanks for the edits. i didn't check thoroughly, but we don't consume `VELA_BUILD_CHANNEL` anywhere? i see a reference in go-vela/vela-slack and, of course, in docs. so we'll need to update those references. outside of that, it looks like we're in the clear based on a light scan.

wass3rw3rk (approved):

plyr4 (approved):
nice

PR #1329, opened 2025-07-08
fix(ci): disable CGO for static CI builds in workflows

wass3r:
companion to https://github.com/go-vela/worker/pull/644 (this was the origin for it being enabled in go-vela/worker in the first place - copy pasta issue as i was updating workflows). as far as i can tell, this has always been enabled 🤔

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -38,7 +38,7 @@ jobs:
       - name: build
         env:
           GOOS: linux
-          CGO_ENABLED: "1"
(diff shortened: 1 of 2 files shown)


KellyMerrick (approved):

wass3r:
ah, ok. i vaguely remembered something about sqlite, and sure enough: https://github.com/mattn/go-sqlite3?tab=readme-ov-file#installation

so, it has to be enabled. we probably need other checks to catch dynamic linking issues since that won't work in SCRATCH image.

PR #1339, opened 2025-08-18
fix(compiler)!: rectify service and secret rulesets for Compile and CompileLite

ecrupper:
### Breaking change

Users who have specified rulesets for services will now see them honored 😅 

### Other change

CompileLite will honor service and secret origin rulesets


plyr4 (approved):
nice ty for adding those

PR #1343, opened 2025-08-26
feat(events)!: support merge queue for GitHub SCM

ecrupper:
ref: https://github.com/go-vela/community/issues/1061

### Breaking Changes
- New database field for repos `merge_queue_events`, which will be a list of events that the `merge_group` event build will report to. This is necessary to preserve normal required status checks while also allowing for a merge queue. Vela is event-based in its status checks, not job-name-based. So required checks from a `pull_request` build will need to be targeted for `merge_group` events.

### Opt In

Ultimately, this is opt-in functionality. If users do not supply any override events, then Vela will behave the exact same.

### Pipeline Changes

```yaml
steps:
  - name: tests
     image: golang:latest
     ruleset:
        event: [ pull_request, merge_group ] 
     commands:
       - go test ./...
```

### Auto Cancel

Determining stale builds for a merge group is different than normal builds. Stale builds are defined as builds that are `pending` or `running` that are part of the merge group and also include a build that failed. There's a more detailed description of this in the reference link.

### DIFF
--- a/api/build/update.go
+++ b/api/build/update.go
@@ -175,7 +175,7 @@ func UpdateBuild(c *gin.Context) {
 		b.GetStatus() == constants.StatusKilled ||
 		b.GetStatus() == constants.StatusError) && b.GetEvent() != constants.EventSchedule {
 		// no need to update status for auto canceled merge group builds
-		if b.GetEvent() == constants.EventMergeGroup && strings.Contains(b.GetError(), "merge group build was auto canceled") {
+		if b.GetEvent() == constants.EventMergeGroup && strings.Contains(b.GetError(), constants.ErrorMergeGroupBuildCanceled) {
 			return
 		}
 		// send API call to set the status on the commit
--- a/api/repo/update.go
+++ b/api/repo/update.go
@@ -181,7 +181,7 @@ func UpdateRepo(c *gin.Context) {
 		for _, event := range input.GetMergeQueueEvents() {
 			// only allow events possibly related to a PR merge queue
 			if !slices.Contains([]string{constants.EventPush, constants.EventPull, constants.EventComment}, event) {
-				retErr := fmt.Errorf("merge_queue_event of %s is invalid", event)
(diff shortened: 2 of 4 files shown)


api/build/restart.go
```diff
@@ -108,6 +108,15 @@ func RestartBuild(c *gin.Context) {
 		return
 	}
 
+	// deny restarts for merge queue builds. Users should enqueue a new build instead.
+	if b.GetEvent() == constants.EventMergeGroup {
+		retErr := fmt.Errorf("unable to restart build %s/%d: cannot restart a build in a merge group", r.GetFullName(), b.GetNumber())
```
wass3r: i like how you said what should be done instead in the comment, would that be useful for the user to see too?

scm/github/repo_test.go
```diff
@@ -1001,6 +1001,81 @@ func TestGithub_Status_Success(t *testing.T) {
...
+	step.SetID(1)
+	step.SetNumber(1)
+	step.SetName("test")
+	step.SetReportAs("test")
+	step.SetStatus(constants.StatusSuccess)
+
+	client, _ := NewTest(s.URL)
+
+	// run test
+	err := client.Status(context.TODO(), b, "foo", "bar", "bar")
```
wass3r:
```suggestion
	err := client.Status(t.Context(), b, "foo", "bar", "bar")
```

.. in a few places

api/build/cancel.go
```diff
@@ -232,3 +162,123 @@ func CancelBuild(c *gin.Context) {
 
 	c.JSON(http.StatusOK, b)
 }
+
+// CancelRunning is a helper function that determines the executor currently running a build and sends an API call
+// to that executor's worker to cancel the build.
+func CancelRunning(c *gin.Context, b *types.Build) (*types.Build, error) {
```
wass3r: should we use a helper for the http clients, etc in here?

wass3r (changes requested):
first pass - see below

PR #1355, opened 2025-09-17
fix(oidc): no trailing colon on non-action events in claim

ecrupper:
Should this be marked breaking and put in `v0.28` or slotted into `v0.27` as part of a patch?


wass3rw3rk (approved):
patch on 27

PR #1356, opened 2025-09-17
enhance(oidc): add jti claim

ecrupper:
Adding an additional layer of security as an ID provider with the JTI claim


wass3rw3rk (approved):
lgtm

PR #1436, opened 2026-04-10
feat(storage): add secured=false option for public artifact access

dan-kirberger:
slopped together idea for https://github.com/go-vela/community/issues/1086

> ⚠️ This PR summary was generated by an AI language model.

## Why

Adds first-class support for publicly-accessible build artifacts. Teams often need to share artifact URLs (e.g. test reports, binaries) without requiring presigned URL expiry or authentication. Previously there was no way to opt individual artifacts out of presigned-URL access.

## What changed

- New `secured: false` field on the `artifacts` block in `.vela.yml`. When set, uploaded objects are stored under a `public/{org}/{repo}/{build}/` prefix in MinIO and returned with a direct (non-presigned) GET URL. Defaults to `true` (presigned, existing behaviour).
- `DirectObjectURL` method added to the `Storage` interface and MinIO implementation to construct direct URLs from `{endpoint}/{bucket}/{key}`.
- `ListObjects` now iterates both the standard and `public/` prefixes and returns direct URLs for public objects.
- New `--storage-public-policy` / `VELA_STORAGE_PUBLIC_POLICY` server flag that, when enabled, automatically applies an anonymous-GET bucket policy scoped to the `public/*` prefix on startup.
- `?secured=false` query parameter support added to the storage API upload handler.
- Full test coverage added/updated for all changed paths; `go test ./...` passes clean.


storage/minio/minio.go
```diff
@@ -72,15 +75,39 @@ func New(endpoint string, opts ...ClientOpt) (*Client, error) {
...
+			logrus.Warnf("storage: failed to apply public bucket policy: %v", err)
+		}
+	}
+
 	return c, nil
 }
 
+// applyPublicPolicy sets an anonymous read-only policy on the public/* prefix of the bucket,
+// allowing unauthenticated GET requests for objects stored under that prefix.
+func (c *Client) applyPublicPolicy(ctx context.Context) error {
```
dan-kirberger, line 89: maybe this doesn't make sense here? better applied separately as some administrative setup?

dan-kirberger:
there'd be similar PRs in worker + sdk i assume

compiler/types/yaml/artifacts.go
```diff
@@ -9,13 +9,21 @@ import (
...
-	Paths raw.StringSlice `yaml:"paths,omitempty" json:"paths,omitempty"`
+	Paths   raw.StringSlice `yaml:"paths,omitempty"   json:"paths,omitempty"`
+	Secured *bool           `yaml:"secured,omitempty" json:"secured,omitempty"`
 }
 
 // ToPipeline converts the Artifact type
 // to a pipeline Artifact type.
 func (a *Artifacts) ToPipeline() *pipeline.Artifacts {
+	// default secured to true when not explicitly set
+	secured := true
```
ecrupper, line 20: It may be easier for implementation/styling if the YAML is inverted here. Something like `insecure` or `public`, where the default of `false` is the more secure option.

dan-kirberger: er, yeah that makes way more sense - thanks! does the general "ask" here sound reasonable? i kinda flailed at a random robo-generated solution here. definitely not tied to anything.

PR #1438, opened 2026-04-13
fix: tighten up UpdateRepo calls and add sync-up to repair

ecrupper:
Removing stray `UpdateRepo` call in the hot path, adding an omission of `hook_counter` to the UpdateRepo function (for user CRUD calls that may occur concurrently with webhook delivery), and adding a small section for fixing out-of-sync hook counters in the repair handler. 

We've only seen an out-of-sync scenario once in our installation of Vela, but hopefully these changes button that all up so we never see it 😅

### DIFF
--- a/api/repo/repair.go
+++ b/api/repo/repair.go
@@ -75,83 +75,25 @@ func RepairRepo(c *gin.Context) {
 
 	l.Debugf("repairing repo %s", r.GetFullName())
 
-	var hook *types.Hook
-
-	// repair out-of-sync hook counter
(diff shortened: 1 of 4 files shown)


api/repo/repair.go
wass3rw3rk, line 66: should we add `nolint:funlen`, or break stuff up to avoid this `funlen` linter warning, if it makes sense to? basically, anything to squash this warning 😆

wass3rw3rk (approved):

KellyMerrick (approved):

PR #1449, opened 2026-05-04
chore(deps): upgrade lestrrat-go/jwx lib to v4

ecrupper:
A couple breaking changes: `GOEXPERIMENT=jsonv2` is required, and the `Import` functions are now generic.


wass3rw3rk:
i'd prefer to wait until json v2 is not hidden behind experimental flag (the proposal is still marked as "under review" here https://github.com/orgs/golang/projects/50). it does seem like it is close and i saw chatter for this to be GA in Go 1.27. for now, it seems this package supports both v3 and v4 (saw them cut a release for both on the same day), so i would propose to slot this in for v29 or beyond even (whenever it drops as GA). thoughts?

PR #1455, opened 2026-05-14
fix: Consider all commits' touched files for push events

colindean:
This includes in the paths changed for a push event all files changed in all commits in that push.

Previously, only the files changed in the only _head_ commit were considered changed in the event. This meant that when e.g. three commits were pushed simultaneously, the files changed in the first two were ignored when considering them as paths for ruleset processing. Only the head, most recent commit, files were considered changed.

Conversations indicated that this is likely a bug, but this fix may be a breaking behavior change warranting further scrutiny. I did not include a way to opt-out of this new behavior.

### DIFF
--- a/scm/github/webhook.go
+++ b/scm/github/webhook.go
@@ -7,8 +7,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"maps"
 	"mime"
 	"net/http"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


scm/github/webhook.go
```diff
@@ -254,6 +254,25 @@ func (c *Client) processPushEvent(_ context.Context, h *api.Hook, payload *githu
...
+	for _, commit := range commits {
+		for _, file := range commit.Added {
+			set[file] = struct{}{}
+		}
+		for _, file := range commit.Removed {
+			set[file] = struct{}{}
+		}
+		for _, file := range commit.Modified {
+			set[file] = struct{}{}
+		}
```
colindean, line 270: **note:** I couldn't figure out _quickly_ how to make a simple inline list in Go so I just did this stupidity for now. Suggestions welcome.

ecrupper: honestly this may be the best route considering uniqueness

ecrupper (approved):

plyr4 (approved):
awesome

PR #1456, opened 2026-05-14
fix(auto_cancel): correct build on nil + number order enforcement

ecrupper:
Fixing a bug introduced in https://github.com/go-vela/server/pull/1441 where `b` is not the build that should be getting updated — rather, `rB` should be. This basically creates a build-blocking bug if any build is found in a hanging running state without a worker

Also, the improvements in https://github.com/go-vela/server/pull/1353 make it far more likely that two builds reach the auto cancel code path simultaneously and result in mutual cancellation.

The first is a more obvious and frequent bug. Second, less so.


wass3r (approved):
🙏

plyr4 (approved):

PR #1464, opened 2026-05-27
fix: use app tokens for all hot path SCM calls if app installed

ecrupper:
Moves the `compileToken` variable higher up in the CompileAndPublish function and adds a token parameter to a few more SCM functions. This essentially removes the owner token from the entire hot path if the repo has the Vela GH App installed.

Updated SCM Functions:

- `GetPullRequest`: for comment events
- `GetBranch`: for schedule events
- `GetUserID`: for webhooks missing the sender SCM ID
- `ConfigBackoff`: for actually fetching the pipeline file for all webhook events

### DIFF
--- a/api/build/compile_publish.go
+++ b/api/build/compile_publish.go
@@ -183,8 +183,15 @@ func CompileAndPublish(
 	// the code in ProcessWebhook implies that the sender may not always be present
 	// fallbacks like pusher/commit_author do not have an id
 	if len(b.GetSenderSCMID()) == 0 || b.GetSenderSCMID() == "0" {
+		senderFetchToken := compileToken
+
+		// use owner token if token is not provided
+		if compileToken == "" {
+			senderFetchToken = r.GetOwner().GetToken()
+		}
+
 		// fetch scm user id for pusher
-		senderID, err := scm.GetUserID(ctx, b.GetSender(), compileToken)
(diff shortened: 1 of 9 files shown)


wass3r (dismissed):

api/build/compile_publish.go
```diff
@@ -92,10 +92,32 @@ func CompileAndPublish(
...
+				return nil, nil, http.StatusInternalServerError, fmt.Errorf("unable to store permission token in cache for installation %d: %w", b.GetRepo().GetInstallID(), err)
+			}
+		}
+	}
+
 	// if the event is issue_comment and the issue is a pull request,
 	// call SCM for more data not provided in webhook payload
 	if strings.EqualFold(cfg.Source, "webhook") && strings.EqualFold(b.GetEvent(), constants.EventComment) {
-		commit, branch, baseref, headref, err := scm.GetPullRequest(ctx, r, prNum)
+		commit, branch, baseref, headref, err := scm.GetPullRequest(ctx, r, prNum, compileToken)
```
wass3r, line 120: so, this was previously using the owner token and now it will use app token, but what happens if it's not an app setup? seems like it now depends on app setup or am i reading it wrong?

ecrupper: good call. will need to address this

wass3rw3rk (approved):

timhuynh94 (approved):

